PR #21, opened 2016-01-06
Mathsat5: Support of interrupting AllSAT calls.

stahlbauer:
MathSat5 has the possibility to interrupt AllSAT queries. This set of changes adds support for this feature in JavaSMT.


PhilippWendler:
Why is this necessary? MathSAT already checks the ShutdownNotifier internally during solving. If this does not work, we should report specific instances of this problem to the MathSAT developers instead, because they do want to fix these.

cheshire:
Actually a good point.
Also the `shutdownIfNecessary` calls are quite far from being free.

PR #78, opened 2016-08-17
Split different linters into different jobs.

cheshire:


PhilippWendler:
I think I have seen a trick somewhere to add a environment variable like `CHECK` and set it to a distinct value in each config. The variable does not need to be used somewhere, but it is shown on the Travis website and be used to distinguish the individual builders there.

PhilippWendler:
I am not sure whether deployment is done for all matrix configs? If yes, we should probably add something to execute them only where desired, if not, the javadoc builder needs to run in the main config where the deployment is done. `build-project-ecj` can be in a separate config, though.

cheshire:
> I think I have seen a trick somewhere to add a environment variable like CHECK and set it to a distinct value in each config. 

Yep, should do this.

> I am not sure whether deployment is done for all matrix configs?

Good point, don't know now, will check.

cheshire:
I'm afraid I will have to just merge it in order to see how deployment is done for the matrix.

PR #96, opened 2017-01-10
Enable model and unsat-core generation in Z3 only for solver instances where we need it.

PhilippWendler:
Previously model generation was always turned on, and unsat-core generation never.

I wonder why unsat-core generation seems to succeed anyway (the [default is false](https://github.com/Z3Prover/z3/blob/e8f4dd76c21b52460b6bd95487183aea88749ce0/src/cmd_context/context_params.cpp#L169)), so I would like to have your opinion on this.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3SolverContext.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3SolverContext.java
@@ -165,8 +165,6 @@ final class Z3SolverContext extends AbstractSolverContext {
     }
 
     long cfg = Native.mkConfig();
-    Native.setParamValue(cfg, "MODEL", "true");
-
     if (extraOptions.requireProofs) {
       Native.setParamValue(cfg, "PROOF", "true");
     }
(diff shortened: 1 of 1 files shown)


cheshire:
@PhilippWendler seems reasonable, I trust that you've checked that the created string symbol does not need to be explicitly freed?

cheshire:
also, nice catch!

PhilippWendler:
> @PhilippWendler seems reasonable, I trust that you've checked that the created string symbol does not need to be explicitly freed?

No, I just copied existing code using `mkStringSymbol`. However, I cannot find a method for freeing symbols. I am not sure what this means, should we avoid creating a new symbol every time?

cheshire:
Pending on https://github.com/Z3Prover/z3/issues/886 resolution. If the memory is indeed leaking, the new code is suboptimal (leaking allocated string on every invocation).

cheshire:
Should be good to merge once https://github.com/Z3Prover/z3/pull/924 goes through :P

cheshire:
Seems good to go. Internally string symbols are shared via hash-consing.

Issue #76, opened 2016-08-16
Variable names are not quoted in dumps of Princess formulas


PhilippWendler:
When a variable like `main::a` (which needs quoting with `|` in SMTLIB) is created and dumped with Princess, the result misses the quoting.

To reproduce, simply change variable name `a` in `SolverFormulaIOTest.varDumpTest` to `main::a` and the `(declare-fun a () Bool)` in the assertion to `(declare-fun |main::a| () Bool)`.

PR #101, opened 2017-02-17
quote defined/declared symbols correctly

pruemmer:
Fix for #76


cheshire:
Looks good, I'll check whether we have tests for this and then merge.

PhilippWendler:
> Looks good, I'll check whether we have tests for this and then merge.

cf #76, the tests from `SolverFormulaIOTest` can be extended to use all kinds of "special" variable names (probably best to use all names from `VariableNamesTest`).

cheshire:
Right, seems to work.

Issue #113, opened 2017-07-04
Unsat-core generation fails with MathSAT5 fails for BV-theory


mdangl:
If I use a Mathsat5TheoremProver with the option ProverOptions.GENERATE_UNSAT_CORE
and using the theory of bitvectors, the following exception occurs during the SAT-check:

```
Exception in thread "main" java.lang.IllegalStateException: msat_solve returned "unknown": EagerBVSolverInterface does not support proof generation
	at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.processSolveResult(Mathsat5NativeApi.java:195)
	at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_check_sat(Mathsat5NativeApi.java:162)
	at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5AbstractProver.isUnsat(Mathsat5AbstractProver.java:69)
```

This suggests to me that it might work if instead of the "Eager"BVSolverInterface a "lazy" alternative is used.
Interestingly, an option for this that looks promising is used automatically in Mathsat5InterpolatingProver:
```.put("theory.bv.eager", "false")

It seems to me that this option should be set (to false) in Mathsat5TheoremProver.createConfig, too, if ProverOptions.GENERATE_UNSAT_CORE is given.

PR #114, opened 2017-07-04
Fix issue #113

mdangl:
If instantiating a Mathsat5TheoremProver to generate unsat cores, automatically set the option to switch off the eager BV interface to fix issue #113.


PhilippWendler:
We might want to set this option only explicitly to `false` for unsat cores, but continue to use MathSAT's default (even if it changes in the future) if no unsat cores are requested.

mdangl:
I applied the requested changes.

PR #120, opened 2017-11-25
Sonarcube integration

kfriedberger:
Work in progress. Do not merge!


kfriedberger:
Sonarcube will not be integrated for now.
There are too many false alarms and no benefits for the current situation.

Issue #73, opened 2016-08-03
Ensure trivial automatic formula simplifications for all solvers


PhilippWendler:
Most solvers immediately simplify formulas when they are created, for example they remove trivial and duplicate operands of `and` and `or`, and flatten `and` and `or` if possible. These optimizations are cheap because they only affect the current operator and its operand, not the whole formula (in contrast to the `simplify` operation we provide). Altogether, these optimizations can still reduce the formula size significantly. Maybe this does not increase solving performance, but at the very least it makes formulas easier to read when they are dumped, e.g., for debugging. Thus I suggest to add implement such simplifications for solvers that do not do them automatically, i.e., Z3.

cheshire:
@PhilippWendler E.g. check both operands and shortcut on `true`/`false` for `and` and `or` functions? Wouldn't it make much more sense to do that in Z3?

PhilippWendler:
Yes, and the same for n-ary operations, too. If operands contain duplicate entries, these can be removed, too.

Z3 could do so itself, like our other solvers do, but I am not sure whether they want to (one could assume they would have done this a long time ago if they wanted).

cheshire:
I have opened Z3 ticket, if they'll close it, I'll do it in JavaSMT.

cheshire:
@PhilippWendler OK I'll do it, do we want any other trivial simplifications (multiplication by zero, multiplication by one, adding zero?...)

PhilippWendler:
Maybe, yes. As long as checking is local (i.e., `O(n)` with `n` being the number of parameters of the current term) I guess it is acceptable performance-wise.

What other solvers also do is elimination of duplicate arguments to operators like `and` and `or`. This can also be done in `O(n)` by traversing over the arguments and copying them into a new list or array, keeping a HashSet of "already seen" arguments.

cheshire:
Interesting, I suspect that with hash consing we do not even need to do
equality testing on AST objects, as the returned native pointers would be
identical.

cheshire:
@PhilippWendler Do you think hould we have an optional checking (controlled by a flag), or just always-enabled? I think I would be in favor of a flag.

PhilippWendler:
I would do this always. Why would anyone want to turn this off? For other solvers it is also always on, and we can't change it.

kfriedberger:
[...]

PR #161, opened 2019-08-14
Simplify n-ary and/or with duplicate operands like and(x, y, x)

PhilippWendler:
SMTInterpol already does this, now we do it for all other solvers as well.
This basically re-adds 36e47b5b (which was only for Z3), but in a more memory-efficient way (`Stream.distinct()` typically uses a regular `HashSet`, the old commit used a `LinkedHashSet`).

The performance and memory consumption are O(n) (amortized), just like the existing code. The overhead is basically one additional `HashSet`.

In the discussion of 36e47b5b it was decided to not do this, and I am also not really convinced this is worth it (we now consistently do simplifications of `true` and `false` operands including short-circuiting, this PR would only filter out duplicate operands, which are probably rare).
But given the new implementation and that this is the first time we have an implementation that would make the behavior across all solvers consistent, I would like to put this up for discussion. At the least we will have this PR as permanent record of this potential solution.

Cf. #73

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractBooleanFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractBooleanFormulaManager.java
@@ -139,15 +139,19 @@ public abstract class AbstractBooleanFormulaManager<TFormulaInfo, TType, TEnv, T
 
   /**
    * Create an n-ary conjunction. The default implementation delegates to {@link #and(Object,
-   * Object)} and assumes that all simplifications are done by that method. This method can be
-   * overridden, in which case it should filter out irrelevant operands.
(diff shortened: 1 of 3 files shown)


kfriedberger:
As we never communicate towards the user, that formulae are structured as how they are build, I do not see a problem with merging this PR. There are more solvers (coming sooner or later) that do not support automatic simplification on that level.
Additionally, this might even replace and simplify user code that does the same filtering on its own.

PhilippWendler:
Yes, from a functional point of view there is no problem, there is just the overhead that was the cause for rejection in the past.

Not sure how many users actually do filter duplicates themselves, and everything except duplicates is also filtered out without this commit.

PhilippWendler:
Ping? How about deciding on this one?

PR #166, opened 2019-10-07
CPAchecker LassoRanker Update

kfriedberger:
This PR contains changes required by CPAchecker for updating the LassoRanker library that strongly depends on SMTInterpol and requires direct access to its components.


PhilippWendler:
I am wondering whether this sets a bad precedent? If we do changes like this for CPAchecker, other people could come as well and ask for breaking the encapsulation further. Where would we stop?

Is this really necessary? CPAchecker could for example use SmtInterpol directly without wrapping it in JavaSMT if it needs full access. Where is the code that requires this change?

kfriedberger:
The corresponding code is local work in progress at @TBunk . CPAchecker generates formulae via its default PathFormulaManager (internally using SMTInterpol) and gives them towards LassoRanker (directly using SMTInterpol).

PR #180, opened 2020-02-08
Boolector v3.2.0 update

baierd:
You may discuss this issue here

New Boolector version released (v3.2.0)
-new API calls
-bugfixes
-it may be possible to use parsing now (for our purposes)


kfriedberger:
What do you mean by "it may be possible to use parsing".
Can you try out whether this works or not.
If it works, is there a JUnit test for this?
You might want to enable Boolector in some more JUnit tests.

kfriedberger:
When compilign Boolector with our `ant package-boolector` with version `3.2.0` (`365eaab6`), I get an error:
<pre>
workspace/boolector/src/interface_wrap.c: In function ‘Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_boolector_1help_1parse’:
/home/fri/workspace/boolector/src/interface_wrap.c:3434:14: error: too few arguments to function ‘boolector_parse’
     result = boolector_parse(arg1, f, "temp", fout, errormsg, status);
</pre>

@baierd Could you take a look?

baierd:
I fixed this, will be pushed shortly. They did some undocumented changes in their code, as well as a new API call that wasnt documented.

baierd:
> What do you mean by "it may be possible to use parsing".
> Can you try out whether this works or not.
> If it works, is there a JUnit test for this?
> You might want to enable Boolector in some more JUnit tests.

They added the possibility to parse without checking SAT immediately. 
Therefore it may be possible to parse and manipulate/use the parsed formulas and then check SAT now.
I will test this and report back/implement it in the course of this week, as well as all other Boolector 3.1/3.2 related issues. (ShutdownManager)

kfriedberger:
@baierd 
I force-pushed some commits to have a cleaner history without the binary library files.
I updated Boolector in the Ivy repository.

The library of CryptoMinisat is missing or not included or not build or not configured. Thus I removed it from the list of sat solvers. The BtorOptions are updated now.

Please take a look at the remaining changes and test again on your system.
Some feedback about the parsing process of Boolector would also be nice.

baierd:
@kfriedberger 
Thank you for your assistance i will take a look at the changes.

The usage of the general parsing method was changed without documentation and i am currently working on fixing it and will test it afterwards.

Should i send you the newly compiled librarys after i am done or do you want the used compile arguments and options to do it yourself?

kfriedberger:
All compile options should be included in the files for building the solver with ant, see `build/publish-solvers.xml`. All source files shouldbe checked in at `lib/native/source/`. No need to send anything by mail.

kfriedberger:
@baierd 
Is there a reason for changing the BtorOptions? The CI complains about it. 
Either the CI or the BtorOption is invalid.

baierd:
@kfriedberger 
Maybe because i used a comment inside an enum. 
I will take a look and fix the CI this evening.
I changed some options to reflect the changes in btortypes.h (the "mother" of our BtorOptions).

kfriedberger:
Maybe they renamed their options again after the release.
I took their git revision tagged with `3.2.0`.
Maybe you used a newer version, then it would be nice to give me the used revision.

baierd:
@kfriedberger 
I am using release 3.2.0, commit 365eaab in their github repo.
But i see now that they fixed a bug 1 commit later, so maybe we should just use 9f4044c

kfriedberger:
They fixed a bug in a Python file. I assume this has nothing to do with JavaSMT. :-)
Updating to the latest revision is possible and I will do it before merging this PR.
Until then, you need to fix your latest change of the BtorOptions (either the code or the unit test) and provide some info about the parsing process (and possibly enable it for most of our unit tests, see `SolverBasedTest0#requireParser()`).
Could you execute the `BoolectorNativeApiTest` on your machine for your version of Boolector?

baierd:
True, i was just taking a peek at my phone, my bad. 
We can use 448b3c6 which is the current commit.
I will change the tests and run `BoolectorNativeApiTest `.
I doubt that parsing will be useable for us because you still can't parse after you added anything to your boolector instance or after you parsed. Conclusive answers and if possible implementations (tests/methods) in regards to parsing/shutdownManager should be done by monday.

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -3452,13 +3475,13 @@ SWIGEXPORT jstring JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJN
       if (!arg2) return 0;
     }
 	
-    FILE *f = 0;
-    f = fopen("temp", "w+");
-	if(f==NULL) {
+    FILE *fparse = 0;
+    fparse = fopen("tempparse", "w+");
```
kfriedberger: @baierd Can we use a temporary file (maybe via "tmpfile()") here instead of creating a real file? The current file remains on disk and might look strange for users.

baierd: @kfriedberger 
Pretty sure we can, i will take a look.

baierd:
@kfriedberger 
Temporary files were not an option because we need the name of the file and temp files dont have names. 
However the files are now deleted after they are not needed anymore.
The only exception to that is if the method crashes in between file creation and deletion, which is happening in the parse method at the moment, i am still investigating as i think the bug is on our side, not Boolectors.
We need a new test in TimeoutTest for solvers that don't use integer theory and we are good to go.
I should be able to provide one tomorrow.

PhilippWendler:
Using hard-coded filenames for temporary files is a [security risk](https://cwe.mitre.org/data/definitions/377.html) and must be avoided. One must use appropriate library functions to create temp files, and there do exist variants where the temp file has a name.

build/build-publish-solvers.xml
```diff
@@ -206,7 +206,7 @@
             <arg value="libpicosat.so" />
         </exec>
         <exec executable="strip" dir="${boolector.path}/deps/install/lib/">
-            <arg value="libminisat.so" />
+            <arg value="libminisat.so.2" />
```
kfriedberger: This change should be part of a different commit. Maybe split commits next time.

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -3519,56 +3556,57 @@ SWIGEXPORT jstring JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJN
...
     
     //read
-    if (f)
-    {
+    if (f) {
      fseek (f, 0, SEEK_END);
       length = ftell (f);
       fseek (f, 0, SEEK_SET);
-     buffer = malloc (length);
+     buffer = malloc (length);		//We leak this memory at the moment!
```
kfriedberger: Why do we need to leak memory? Can we free it at the end of this method?

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -235,6 +238,30 @@ static int32_t java_termination_callback(void *user_data) {
 	return (int32_t)(result != JNI_FALSE);
 }
 
+//Takes String and appends it to valid path for temporary files
+//Make sure that filename is compliant with the used temp file method
+//Returns NULL in case of NULL filename (so make sure you dont enter NULL!)
+char *addTemppathToFilename(char *filename) {
+	
+	if(!filename) {
```
kfriedberger: Spaces ot Tabs?

lib/native/source/libboolector/interface_wrap.c
baierd: Fixed in next commit.

build/build-publish-solvers.xml
baierd: Sorry. 
Will be seperate next time.

lib/native/source/libboolector/interface_wrap.c
baierd: Spaces.
Fixed in next commit.

baierd:
@PhilippWendler 
> Using `strncpy` with `strlen` as last parameter does not make sense. The one advantage that `strncpy` has is to be safe against buffer overflows if the string given in `dir` is larger than expected. But if the string is larger than expected, of course `strlen` will just return that large size. So `strncpy(s1, s2, strlen(s2))` is identical to `strcpy(s1, s2)` except that the former skips the terminating null byte of `s2`.
> 
> The length parameter given to `strncpy` here should be calculated from `n`, to detect bugs if that logic is incorrect.

My intention was exactly to skip the null byte as i am appending the filename in the next step that has a null byte at its end.
What do you mean with n exactly?

Would you say "TMPDIR" as environmental variable for temp files is enough or should i check "TMP" etc.?

PhilippWendler:
> > Using `strncpy` with `strlen` as last parameter does not make sense. The one advantage that `strncpy` has is to be safe against buffer overflows if the string given in `dir` is larger than expected. But if the string is larger than expected, of course `strlen` will just return that large size. So `strncpy(s1, s2, strlen(s2))` is identical to `strcpy(s1, s2)` except that the former skips the terminating null byte of `s2`.
> > The length parameter given to `strncpy` here should be calculated from `n`, to detect bugs if that logic is incorrect.
> 
> My intention was exactly to skip the null byte as i am appending the filename in the next step that has a null byte at its end.

That is of course ok.

> What do you mean with n exactly?

You have a variable called `n`, I was referring to that. The length argument of `strncpy` should be calculated in a way that it is guaranteed to be shorter than the target length. Of course, in the current code that is the case, but it is not obvious and if someone changes the code, it would be easy to mistakenly add an error. If the length argument is derived directly from the target length `n` instead of the length of the source string, such errors are less likely.

> Would you say "TMPDIR" as environmental variable for temp files is enough or should i check "TMP" etc.?

`TMPDIR` is fine: https://unix.stackexchange.com/a/352108/15398

baierd:
@kfriedberger 

The code should stop in all error cases now.
Unfortunately `boolector_parse()` in its current form isn't really useable for us, but it works.
(As long as we dont have the API to change existing nodes/formulas its not helpful anyway)
Are there any other issues that should be tackled in this branch?

PR #181, opened 2020-02-16
Integration of Yices2

kfriedberger:
This branch tracks the progress of the integration of Yices2.

As the SMT solver Yices2 is available under GPLv3, we can not provide the binary directly with JavaSMT. We can only provide an optional dependency for users.

Steps to do:
- Mark Yices2-related code as GPLv3 and Apache2.
- Divide the Ant build process of Java code for JavaSMT-related and the Yices2-related code and provide 2 separate Jar files via Ivy.
- Add a dependency towards the binary files of Yices2.
- Document and Provide the necessary Ivy-dependencies.


kfriedberger:
To get around the kicense compatibility issue (Apache2 vs GPLv3) we need to split the binary archive (jar file) for publication.
There are (at least) two ways to do this:
- modify the build scripts for ant/ivy and maven to automatically publish two seperate jars simultanously. This would allow internal modifiactions and avoids any incompatibilities between the jar files (they would have the same version number in the release).
- extract the components for Yices2 into a completely independent project. This would allow easier maintenance for each individual project, e.g., using common ant/ivy files from the group's template repository without modifying them.

@PhilippWendler : do you have any preference?

PhilippWendler:
The first seems less effort, so I suggest to do this until there is a reason for the split.

baierd:
@kfriedberger @PhilippWendler 

We have the attribute `packageset` and `excludepackagenames` both set in `build-documentation.xml`. This results in neither of them working (tested and confirmed) and according to documentation:

> If you use multiple ways to specify where javadoc should be looking for sources, your result will be the union of all specified documentations. If you, e.g., specify a sourcepath attribute and also a nested packageset both pointing at the same directory your excludepackagenames attribute won't have any effect unless it agrees with the exclude patterns of the packageset (and vice versa).

Do we want that behaviour? (At the moment nothing is excluded from JavaDoc)

If not i'd suggest we choose one method of excluding packages in the javadoc.

(My commit [35fd485](https://github.com/sosy-lab/java-smt/pull/181/commits/35fd485671db942e00bb9403ff2dc120e76854aa) isn't working so i investigated)

Why are the solvers z3, mathsat5, princess and smtinterpol excluded in JavaDoc (according to build.xml)?

PhilippWendler:
Please try to avoid changing `build/build-jar.xml`, because we have several Java projects and want to keep our build files synchronized over these projects to reduce the effort. If it is not possible to go without changing the file, then introduce a new property in this file, and define its value in `build.xml` (cf. the comments and existing examples at the top of the file).



> We have the attribute `packageset` and `excludepackagenames` both set in `build-documentation.xml`. This results in neither of them working (tested and confirmed) and according to documentation:

I did not find `packageset` anywhere? And our [API documentation](https://sosy-lab.github.io/java-smt/api/index.html) seems to work fine?

(Although the documentation mentions the CVC4 classes, which it should not.)

PhilippWendler:
Ok, now I understand, you meant `fileset` instead of `packageset`. You were right, I fixed that in 093ac25 and it now works even without needing to list all solvers.

src/org/sosy_lab/java_smt/solvers/yices2/package-info.java
```diff
@@ -0,0 +1,15 @@
+/*
+ *  JavaSMT is an API wrapper for a collection of SMT solvers.
+ *  This file is part of JavaSMT.
+ *
+ *  Copyright (C) 2007-2020  Dirk Beyer
+ *  All rights reserved.
+ *
+ *  SPDX-License-Identifier: Apache-2.0 AND GPL-3.0
```
PhilippWendler, line 8: Is there a reason why this is a conjunction? This imposes the most restrictions on us, for example we are now always forced to honor both licenses in the future. If it were a disjunction, we could for example later drop the GPL if Yices moves to a different license.

Note that this would not change the licenses that users of the combination of JavaSMT would need to honor, because `Apache-2.0 AND GPL-3.0 AND (Apache-2.0 AND GPL-3.0)` (for JavaSMT + Yices + this package) is of course the same as `Apache-2.0 AND GPL-3.0 AND (Apache-2.0 OR GPL-3.0)`.

baierd:
@kfriedberger 
With commit [3201699 ](https://github.com/sosy-lab/java-smt/pull/181/commits/3201699bd77e96d9b21ef3fb9f5aa6c079424cdd) the command publish-yices2 enables building and publishing of `javasmt-yices2*.jar` and the corresponding sources jar.
There is a little problem as to why its 2 commands and why you can't call both of them at the same time though: ivy uses the `ivy-[version].xml` that gets generated to publish the artifacts. But because they are all named the same we have to overwrite/delete the old xml file before we can publish something new. I added a notice at the end of the `publish` command to reflect this.
If we want 1 command or both of them at the same time we need to either separate the files in different folders or simply change the names and add the module name, for example: `ivy-[module]-[version].xml`


Additionally i tested the JavaSMT-Yices2 Wrapper as JNI-Wrapper for Yices2 without JavaSMT (Hence the change of visibility of the Native-API) and it works.
I made some naive tests for the JavaSMT-Yices2 as well and all went well but i'll test it once more in the CPAchecker and report back.

PhilippWendler:
A separate directory for publishing the Yices stuff would be the obvious solution I guess.

baierd:
Does the generated `ivy-*.xml ` have any use in the JavaSMT folder after publishing?

PhilippWendler:
> Does the generated `ivy-*.xml ` have any use in the JavaSMT folder after publishing?

No, and the respective JAR file with the version number in the name also not.

baierd:
@kfriedberger 
Do you want Yices2 binaries to be build in re-entrant mode?
(Would enable multi-thread tests/support)

kfriedberger:
If that is easily possible and does not affect the runtime, then it would be nice to directly have a reentrant binary.

kfriedberger:
Returning an invalid result might be a bug in Yices2.
Ok, then lets shift the reentrant version for later.
Thanks for investigation.

kfriedberger:
This branch is developed as far as possible to have a working state.  
We can integrate it into `master` and fix the remaining things later.  
This makes Yices2 available with the next release of JavaSMT.

doc/Developers.md
```diff
@@ -201,6 +201,49 @@ The scripts for publishing Princess and SMTInterpol are available
...
+### Publishing Yices2
+
+Yices2 consists of two components: the solver binary and the Java components in JavaSMT.
+The Java components were splitt from the rest of JavaSMT because of the GPL.
+
+#### Publishing the solver binary for Yices2
+
+Prepare gperf and gmp (required for our own static binary):
+```
+wget http://ftp.gnu.org/pub/gnu/gperf/gperf-3.1.tar.gz && tar -zxvf gperf-3.1.tar.gz && cd gperf-3.1 && ./configure && make
```
baierd, line 213: Is this simplyfied version of how to build (compared to the original build information in compile.sh) working flawlessly?

kfriedberger: I will test it again. Maybe I have overlooked something.

baierd: I was just asking because in the compile.sh of Yices2 provided by the author has some extra flags. I didn't test (but can if you want me to).

kfriedberger: The script compile.sh was written by a BA student and possibly copied together from other scripts.
I know that `-fpic` is important, but I could not yet crash the application.
Performance could also be different on larger queries.
I will check again.

baierd: I tested the concurrency with those flags. Maybe without them its working? I'll test tomorrow. Want me to run it through CPAchecker to check the performance?

Issue #168, opened 2019-10-14
Update Z3 to upcoming release


PhilippWendler:
Z3 has added two new features that are useful for JavaSMT and which should be available in the next release of Z3:
- https://github.com/Z3Prover/z3/issues/867 (SMT2 log file for solver interaction via the API)
- https://github.com/Z3Prover/z3/issues/1006 (Allow interrupts on individual solvers)

As soon as a new release is made, we should upgrade to it. Also related to the upgrade: #116

baierd:
Unfortunately both Issues are not available in Java at the moment. 
I will request Java implementation.
(The Z3 JNI wrapper is pretty good. We could fix it ourselfs and request a PR if need be)

kfriedberger:
This issue is solved for now.

Issue #154, opened 2019-04-16
Support check-sat-using tactic for Z3


PhilippWendler:
Z3 provides not only `check-sat` but also `check-sat-using <tactic>` in SMTLib, so probably also via the API. This can be used to let Z3 use different solving strategies, which might be more effective in some scenarios.

Users would like to experiment with this ([example](https://groups.google.com/d/msg/cpachecker-users/LK4DzRhR7Xc/Z6nWzU4ACwAJ)), so I suggest we add a configuration option for Z3 that lets users specify such a tactic which then gets used for all solving calls.

Issue #148, opened 2019-02-18
Update Z3 to 4.8.* or later


kfriedberger:
This issue tracks the progress of updating Z3 to a version later than 4.7.1., which is the current version of Z3 in JavaSMT.
Version 4.7.1 is the last release with interpolation support. With a later version we will have to drop interpolation for Z3 in JavaSMT, too. Additionally there seem to be some further changes in the API.

kfriedberger:
Fixed with #182 .

Issue #116, opened 2017-08-21
Adjust to new interrupt handling of Z3


PhilippWendler:
https://github.com/Z3Prover/z3/commit/7a977f0106fe0ae07c90d1566a8a75e2d430f2f9 seems to have changed the handling of interrupts of Z3. It is not clear to me whether Z3 now still returns `canceled` or `interrupted` as error message when we interrupt it, so we need to test this and adjust our error handling accordingly.

PhilippWendler:
Was this done now that Z3 was updated?

kfriedberger:
For several cases, the interrupt behaviour of Z3 did not change.
For optimizazion (min/max) there seems to be a difference, which will be adapted soon.

kfriedberger:
Should be fixed with 764addf53939711fb1e98aadc15463083c726821.

PhilippWendler:
Is there still something left to do?

kfriedberger:
I did not notice any problems since then, so we can close this issue.

PR #182, opened 2020-03-11
Z3 v4.8.7 update

baierd:
Discuss the update of Z3 to v4.8.7 as well as the issues #168  #154 #116  and #148 here.
Z3 does only provide binaries for Ubuntu 16.04 and Windows at this point, so we will have to compile it ourselfs.
The JavaSMT code for interpolation was already removed and the tests adapted.
@kfriedberger


PhilippWendler:
Why can't we use the binary for Ubuntu 16.04? Is this too new for what JavaSMT wants to support?

kfriedberger:
The binary for Ubuntu16.04 should also work on newer Ubuntu versions. It should be fine.
@baierd please test and update the scripts, afterwards I will update the Ivy-part and merge this PR.

kfriedberger:
I fixed the ant script for building Z3 from its release and updated the Z3 version in Ivy up to 4.8.7.
There are several SegFaults in the JUnit tests (at least on my laptop).

@baierd 
Can you take a look? Is this a problem in Z3 or on our site?

baierd:
I accidentally reverted my Z3 lib, hence why i did not catch that.

There is a change in the return value of the parser that needs fixing (It returns a vector of
formulas as opposed to a conjunction of formulas now).
I will fix this tomorrow.
Additionally i will add solver interrupt and smt2 logging (because the Native module is able to use these after all)

kfriedberger:
Currently, we need to fix the model evaluation for Z3 (high priority).
Additionally, we can extend the current functionality in JavaSMT with new features like SMT-logging in Z3 (low priority).

baierd:
I fixed the parser and i am working on the model at the moment.

src/org/sosy_lab/java_smt/solvers/z3/Z3SolverContext.java
```diff
@@ -162,6 +162,7 @@ public static synchronized Z3SolverContext create(
       Native.setParamValue(cfg, "PROOF", "true");
     }
     Native.globalParamSet("smt.random_seed", String.valueOf(randomSeed));
+    Native.globalParamSet("model.compact", "false");
```
kfriedberger: @baierd can you give more details on this option? is this option disabling a `partial model`?

baierd: It deactivates a new default option that compresses function definitions. Deactivating restores the old standard behavior.

kfriedberger:
The logfile is currently empty for my own tests.

kfriedberger:
@baierd Could you update the documentation for Z3?
At least the following places seem to be outdated:
- https://github.com/sosy-lab/java-smt/blob/master/doc/Developers.md#publishing-z3
- https://github.com/sosy-lab/java-smt/blob/master/doc/KnownIssues.md#z3

baierd:
Sure, ill take a look.

baierd:
@kfriedberger 
The SMT2-Log had a bug that is now fixed in Z3.
Should i remove the SMT2-Log until the next release of Z3?

kfriedberger:
Yes, remove the log from JavaSMT as long as it does not work properly.
If possible, add a reference to the bug report in Z3 into this discussion.
Then we can merge this PR or is anything remaining?

kfriedberger:
@baierd any news?

baierd:
The Z3 Log is removed.
I think we can merge.

Z3 Issue for the SMT2 Log: [https://github.com/Z3Prover/z3/issues/3513](url)

Issue #183, opened 2020-03-21
Boolector: SegFault when dumping a formula. 


kfriedberger:
Sometimes it happens, that Boolector crashes with a memory error.
This seems to be randomly, but I got at least the following messages (from different runs of the `VariableNamesTest`):
- `free(): invalid next size (fast)`
- `java: malloc.c:2401: sysmalloc: Assertion "(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)" failed.`
- `A fatal error has been detected by the Java Runtime Environment:
  SIGSEGV (0xb) at pc=0x00007fe844fcb0dd, pid=23305, tid=23311
JRE version: OpenJDK Runtime Environment (11.0.6+10) (build 11.0.6+10-post-Ubuntu-1ubuntu118.04.1)
Java VM: OpenJDK 64-Bit Server VM (11.0.6+10-post-Ubuntu-1ubuntu118.04.1, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
Problematic frame: C  [libc.so.6+0x950dd]`

I am not sure, but the problem might be located in `BtorJNI.boolector_help_dump_node_smt2` called from `BoolectorFormulaManager.dumpFormula`, as this method seems to be called directly before such crashes. I could not extract any more precise stack trace.

kfriedberger:
This seems to be fixed with #185 .

PR #185, opened 2020-03-31
Fixed SegFault when dumping and cleaned up the corresponding code

baierd:
@kfriedberger 
Fix for #183 is implemented and working.
I cleaned up some code.

If its ok i will clean up the wrapper some more because its a little confusing and the SegFault is still existing in the unused methods.

### DIFF
--- a/lib/native/source/libboolector/interface_wrap.c
+++ b/lib/native/source/libboolector/interface_wrap.c
@@ -251,16 +251,16 @@ char *addTemppathToFilename(char *filename) {
   if(dir == NULL || strlen(dir) == 0) {
     dir = "/tmp/";
   }
-	
+  
   int dirLength = (int)strlen(dir);
   int filenameLength = (int)strlen(filename);
   int completeNameLength = dirLength + filenameLength + 1;
 
-  char *tfnwd = (char *)malloc(completeNameLength * sizeof(char));
-  strncpy(tfnwd, dir, dirLength);
-  strncat(tfnwd, filename, (filenameLength + 1));
+  char *tempfileName = (char *)malloc(completeNameLength * sizeof(char));
+  strncpy(tempfileName, dir, dirLength);
+  strncat(tempfileName, filename, (filenameLength + 1));
     
-  return tfnwd;
+  return tempfileName;
 }
 
 #ifdef __cplusplus
@@ -429,12 +429,12 @@ SWIGEXPORT void JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_b
     arg2 = (char *)(*jenv)->GetStringUTFChars(jenv, jarg2, 0);
     if (!arg2) perror("ERROR: couldn't set api trace because given path was wrong.");
   }
-  FILE *f = 0;
-  f = fopen(arg2, "w");
-  if(f==NULL) {
+  FILE *file = 0;
+  file = fopen(arg2, "w");
+  if(file == NULL) {
     perror("ERROR: couldn't set api trace because it couldn't open trace file.");   
   }
-  boolector_set_trapi(arg1,f);
+  boolector_set_trapi(arg1,file);
 }
 
 
@@ -3460,26 +3460,24 @@ SWIGEXPORT jint JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_b
 SWIGEXPORT jstring JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_boolector_1help_1dump_1smt2(JNIEnv *jenv, jclass jcls, jlong jarg1) {
   jstring jresult = 0;
   Btor *arg1 = (Btor *) 0 ;
-  char *result = 0 ;
-  char tfn[] = "boolector_help_dump_smt2_tempfile-XXXXXX";
-  FILE *f = 0;
-  int fd = -1;
-  char * buffer = 0;
-  long length = 0;
-  char *tfnwd = addTemppathToFilename(tfn);
+  char tempFilenameTemplate[] = "boolector_help_dump_smt2_tempfile-XXXXXX";
+  FILE *file = 0;
+  int fileDescr = -1;
+  char *buffer = 0;
+  long fileLength = 0;
+  char *tempfileName = addTemppathToFilename(tempFilenameTemplate);
 
-  fd = mkstemp(tfnwd);
-  if(fd == -1) {
-    free(tfnwd);
+  fileDescr = mkstemp(tempfileName);
+  if(fileDescr == -1) {
+    free(tempfileName);
     perror("ERROR CREATING TEMPORARY FILE FOR SMT2 DUMPING");
     SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "FileDescriptor for file used in boolector_help_dump_smt2 may not be NULL");
     return 0;
   }
 
-  f = fdopen(fd,"w+");
-  if(f==NULL) {
-    unlink(tfnwd);
-    free(tfnwd);
+  file = fdopen(fileDescr,"w+");
+  if(file == NULL) {
+    unlink(tempfileName);
     perror("ERROR OPENING FILE FOR SMT2 DUMPING");
     SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "File for boolector_help_dump_smt2 may not be NULL");
     return 0;
@@ -3491,32 +3489,34 @@ SWIGEXPORT jstring JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJN
   arg1 = *(Btor **)&jarg1; 
     
   //write
-  boolector_dump_smt2(arg1, f);
-    
+  boolector_dump_smt2(arg1, file);
+  
+  unlink(tempfileName);
   //read
-  if (f) {
-    fseek (f, 0, SEEK_END);
-    length = ftell (f);
-    fseek (f, 0, SEEK_SET);
-    buffer = malloc (length);
-    if (buffer) {
-      if(fread (buffer, 1, length, f) != (unsigned long)length) {
-        perror("ERROR READING FILE INTO BUFFER");
-        SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "Buffer for boolector_help_dump_smt2 may not be NULL");
-        return 0;
-      }
+  if(file) {
+    fseek(file, 0, SEEK_END);
+    fileLength = ftell(file);
+    rewind(file);
+    buffer = (char *)malloc((fileLength + 1) * sizeof(char));
+    if(!buffer) {
+      free(buffer);
+      perror("ERROR CREATING BUFFER");
+      SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "Buffer for boolector_help_dump_smt2 may not be NULL");
+      return 0;
+    }
+    if(fread (buffer, 1, fileLength, file) != (unsigned long)fileLength) {
+      free(buffer);
+      perror("ERROR READING FILE INTO BUFFER");
+      SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "Error reading file into buffer in boolector_help_dump_smt2.");
+      return 0;
     }
   }
 
-  if (buffer) {
-    result = buffer;
-  }  
+  buffer[fileLength] = '\0';
     
-  if (result) jresult = (*jenv)->NewStringUTF(jenv, (const char *)result);
+  jresult = (*jenv)->NewStringUTF(jenv, (const char *)buffer);
   
-  unlink(tfnwd);
-  free(tfnwd);
-  fclose (f);
+  fclose(file);
   free(buffer);
   return jresult;
 }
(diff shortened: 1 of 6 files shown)


kfriedberger:
Yes, cleanup is ok. I will test this later.

kfriedberger:
Even with a newly compiled version of Boolector and its bindings, I still get an error when executing `ant unit-tests`:
<pre>
# A fatal error has been detected by the Java Runtime Environment:
#  SIGSEGV (0xb) at pc=0x00007f0e664080dd, pid=9476, tid=9480
# JRE version: OpenJDK Runtime Environment (11.0.6+10) (build 11.0.6+10-post-Ubuntu-1ubuntu118.04.1)
# Java VM: OpenJDK 64-Bit Server VM (11.0.6+10-post-Ubuntu-1ubuntu118.04.1, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# C  [libc.so.6+0x950dd]
[timeout occurred during error reporting in step "printing problematic frame"] after 30 s.
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport %p %s %c %d %P %E" (or dumping to /home/fri/workspace/java-smt/core.9476)
</pre>
The dump does not provide any more information.

baierd:
Yes can confirm. I'll investigate further.

kfriedberger:
@baierd Could you collect and provide some information about this bug here?
- In which location does the segfault happen? This does not need to be the source of the bug, but just the effective location. Is the bug clearly related to the dumping of formulae?
- Optional: There seems to be no stack trace in the dump. Why? Does this only affect me?
- Can we get a much smaller determinstic example that is better than executing all JUnit tests?

baierd:
There was definitely a memory corrupting bug in the wrapper that is now fixed. BUT it still crashes, but less now (at least it feels like less). I have to run the tests or the class in which it happens multiple times in a row (5-10+) to get a crash (before it was more like 1-2 times).
The problem seems to be a memory corruption and, at least in my tests, 2 methods seem to trip the landmine set by the memory problem:

Both in the class `VariableNamesTest`:
`testBoolVariableDump()` and `testEqBoolVariableDump()` with either the above SigSegv, 
`malloc(): corrupted top size` 
or 
`java: malloc.c:2379: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.` 
randomly.

Both methods are the only ones using toString() in the tests (for Boolector) thats why it seems natural that the method `boolector_help_dump_node_smt2` of the wrapper has something to do with it.

Does your Ubuntu trip the problem more often than my Arch @kfriedberger ?

I am at the moment wrapping all methods in the wrapper with the coffeecatch-wrapper. That may result in some more information in the case of a SigSegv.

baierd:
Found it.
Its the `addTemppathToFilename()` method that corrupts the memory somehow.
I'll roll out a fix this evening.

baierd:
Should be fixed now.
I can't reproduce a SigSegv in the unit tests or my own tests.

kfriedberger:
I can confirm the fix. I did not get any more SegFaults.

However, I noticed several warnings when building Boolector with the new wrapper:
<pre>
[exec] /home/fri/workspace/boolector/src/interface_wrap.c: In function ‘addTemppathToFilename’:
[exec] /home/fri/workspace/boolector/src/interface_wrap.c:260:31: warning: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to provide an explicit length? [-Wsizeof-pointer-memaccess]
[exec]    memset(tempfileName,0,sizeof(tempfileName));
[exec]                                ^
src/CMakeFiles/boolector.dir/btorsat.c.o/home/fri/workspace/boolector/src/interface_wrap.c: In function ‘Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_boolector_1help_1dump_1node_1smt2’:
src/CMakeFiles/boolector.dir/sat/btorcms.cc.o/home/fri/workspace/boolector/src/interface_wrap.c:3754:25: warning: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to provide an explicit length? [-Wsizeof-pointer-memaccess]
[exec]    memset(buffer,0,sizeof(buffer));
[exec]                          ^
[exec] /home/fri/workspace/boolector/src/interface_wrap.c: In function ‘Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_boolector_1uf_1assignment_1helper’:
[exec] /home/fri/workspace/boolector/src/interface_wrap.c:3790:50: warning: ‘return’ with a value, in function returning void
[exec]    if(arg3 == 0 || arg4 == 0 || arg5 == 0) return ((void*)0) ;
[exec]                                                   ^
[exec] /home/fri/workspace/boolector/src/interface_wrap.c:3772:25: note: declared here
[exec]  SWIGEXPORT void JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_BtorJNI_boolector_1uf_1assignment_1helper(JNIEnv *jenv, jclass jcls, jlong jarg1, jlong jarg2) {
[exec]                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[exec] /home/fri/workspace/boolector/src/interface_wrap.c:3816:10: warning: ‘return’ with a value, in function returning void
[exec]    return outerJNIArray;
[exec]           ^~~~~~~~~~~~~
</pre>

@baierd Could you take a look at the warnings?

baierd:
Some are present for a while now.
The first two  `argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination;` are OK, because we simply set the whole allocated String to `NULL` before writing to it. That can however be handeled without a warning.
The `return ((void*)0)` warning doesn't make sense for the C compiler because thats a `NULL` in Java. 
The last one is the same as the one before, just as return value.
I can change those last two into NullPointerExceptions.

I'll make another commit to get rid of them.

baierd:
@kfriedberger 
I fixed all compile warnings/notes.
Should be ready to be deployed.

kfriedberger:
@baierd It seems as if the SMTLIB-dump of boolector is not valid in some cases. Additonally, there seem to be additional chars sometimes. Can you please take a look at the added JUnit test?

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -3654,137 +3652,157 @@ SWIGEXPORT jobjectArray JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_B
...
-  arg2 = *(BoolectorNode **)&jarg2; 
+  arg2 = *(BoolectorNode **)&jarg2;
 
-  f = fdopen(fd, "w+");
-  if(f==NULL) {
-    free(tfnwd);
-    unlink(tfnwd);
-    perror("ERROR: COULDNT DUMP NODE BECAUSE IT COULDNT CREATE A DUMP FILE"); 
+  file = fdopen(fileDesrc, "w+");
+  unlink(tempfileName);
```
kfriedberger, line 3712: Why do we need to unlink the file here?

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -3654,137 +3652,157 @@ SWIGEXPORT jobjectArray JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_B
...
   }
-    
-  if (result) jresult = (*jenv)->NewStringUTF(jenv, (const char *)result);
-    
-  unlink(tfnwd); 
-  free(tfnwd);
-  fclose(f); 
+
+  fileLength = ftell(file);
+  rewind(file);
```
kfriedberger, line 3737: Is rewind really needed here?

lib/native/source/libboolector/interface_wrap.c
```diff
@@ -3654,137 +3652,157 @@ SWIGEXPORT jobjectArray JNICALL Java_org_sosy_1lab_java_1smt_solvers_boolector_B
...
     SWIG_JavaThrowException(jenv, SWIG_JavaNullPointerException, "File for boolector_help_dump_node_smt2 may not be NULL");
     return 0;
   }
-    
+
   //write
-  boolector_dump_smt2_node(arg1, f, arg2);
-    
+  boolector_dump_smt2_node(arg1, file, arg2);
+  rewind(file);  //Just to be sure
```
kfriedberger, line 3721: Is rewind really needed here?
Do we need to flush the file after boolector has written to it?

lib/native/source/libboolector/interface_wrap.c
baierd: We don't, we could do it later, but it doesn't matter because we have the filepointer. The file only gets deleted after we close the opened file and Boolector does not close the file itself. I read that for safety reasons you should do this as soon as possible though.

lib/native/source/libboolector/interface_wrap.c
baierd: I think so, yes.
The stream is set to the end of the file before calling `ftell()` to read the length. After that we should put it back to the beginning for `fread()` later.

lib/native/source/libboolector/interface_wrap.c
baierd: I am honestly not sure if `rewind()` is needed here. It worked fine without but i wanted to be sure that the stream is at the beginning of the file.
I can take another look at the Boolector code to determin if its really needed though.

baierd: No we don't need to flush the file as it is deleted after we close the file (tempfile).
Every call to this method should give us a completely new and empty file to write to.

kfriedberger: But we `read` from it. Is content, which might not yet be flushed, already available for reading?

baierd:
> @baierd It seems as if the SMTLIB-dump of boolector is not valid in some cases. Additonally, there seem to be additional chars sometimes. Can you please take a look at the added JUnit test?

I'm gonna take a look.
At first glance it seems like fragments of the last chars are carried over into new lines at the end?

lib/native/source/libboolector/interface_wrap.c
baierd: Shouldn't that be the job of the system (giving us an empty file)?
But thats easy to test^^

kfriedberger: Boolector dumps the String into a file, then we flush (or do currently not flush), and then we read the file into a Java-String. The question is whether we need to flush in the middle or not.

lib/native/source/libboolector/interface_wrap.c
kfriedberger: Is the old file descriptor not pointing to the start of the file any more? I have never though about it. Is a C-based file a real `stram`? Then this is ok.

lib/native/source/libboolector/interface_wrap.c
baierd: Yes but a temporary file that is newly created every time with a unique filedescriptor. If mkstemp can't create a unique firedescr. it gives back -1 and we check that. So it should not be the cause.

lib/native/source/libboolector/interface_wrap.c
baierd: In C a file is a stream, yes.
The Object of type `FILE` has all the necassary information to control that stream.

lib/native/source/libboolector/interface_wrap.c
kfriedberger: There might be a misunderstanding: I am not talking about the tempfile creation. I just assume that there is already a newly created tempfile and we use it like any other file (or stream).
Is it guaranteed that all "written" data can be "read" again? Or might some data be hanging around in the buffer of the (not yet flushed) write process? Maybe Boolector is already flushing at the end of its method, then we do not need to flush by ourselves.

Why am I asking this? Because the ending of the String from dumping a single variable (see the added JUnit test) contains some strange chars where I cannot detect a reason for. I have seen at least the String endings `"\0"`, `"der;"`, and `""` for the same (!) variable, even while dumping the same node several times within one program execution. A possible reason for this might be that some content of the file, such as its last line, is not yet written to file before the file is read into a String.

lib/native/source/libboolector/interface_wrap.c
kfriedberger: Ok, I see.
Would it be nicer to use `if ( fseek(file, 0L, SEEK_SET) != 0 ) {` instead of `rewind(file)`?
This would allow some error handling.

lib/native/source/libboolector/interface_wrap.c
baierd: You were right, a flush of the file before use fixes the problem.

Why doesn't C guarantee that a newly used file is empty though?!

kfriedberger: Now I am confused. I thought about flushing **after** calling `boolector_dump_smt2_node`, not **before**. I assumed that the filepointer points to the start of the opened file and Boolector starts to write there.
Now the control flow is such that:
- we create a file
- we flush the file (why? There is no content.)
- Boolector writes to the file
- we do not flush the file (I really expected that flush is needed here!)
- JNI reads the file into a String.
Strange. I will test and report whether it works.

baierd: My bad, i misunderstood the behavior of `fflush()`, you are right of course.

baierd: I just recompiled commits [Fix SigSegv](https://github.com/sosy-lab/java-smt/pull/185/commits/68cb0d564394ab020f40e8c612fac8b1a5ba9530) and [Fixed compile warnings](https://github.com/sosy-lab/java-smt/pull/185/commits/ca096fe7b07396eb3707d1c6956751714670b510) and tested them both.
The problem seems to be present in the first, but not in the later, thats why my improper flush "worked", it was already fixed, at least on my system.
Can you confirm?

kfriedberger:
@baierd There is still some problem. Could you take another look?

baierd:
On it.

kfriedberger:
The number of SegFaults is clearly reduced with the latest fix. The JUnit tests run non-deterministically successfully more than half of the time.

baierd:
@kfriedberger 
I can't reproduce any SegFaults now. Could you test in your system please?
Also, is your "TMPDIR" set or is it void?

kfriedberger:
Seems to work on my laptop now. I have not set any tmp-directory in my environment, i.e., `echo $TMPDIR` is empty. I will test on another system and wait for the CI now.

kfriedberger:
Hurray, seems to be fixed.
@baierd can we merge this branch? or are there currently more plans about fixes for Boolector?

baierd:
@kfriedberger 
I could clean/fix the parse method if you want, but we don't use it at the moment.
If not we can merge.

PR #188, opened 2020-04-23
Concurrency tests

baierd:
This branch is by no means done!

I just wanted a first impression of @kfriedberger concerning the complexity and because i break with our test paradigm quite a bit.

Main idea is to test SolverContext concurrency with pre-made contexts (each thread gets a single unique context that exists to work concurrenlty), each thread creating its own SolverContext and work concurrently, test parallel optimization (we need to recompile OptimathSAT for that) and test parallel stack usage in a single context.

The main reason for not using `SolverBasedTest0` is that control of the creation and closing of SolverContexts is needed.
Methods `initSolver `and `closeSolver`.

I tried to start the threads of a test as synchronous as possible and wait for all of them to finish before continuing, providing maximum concurrency (especially while creating new contexts as i read that that may be problematic in some solvers!)
There is a timeout that filters possible problems like deadlocks or long computation times and all exceptions are saved and will be returned to be checked.
See the method `assertConcurrency()` for that.

`bvConcurrencyTest()` and `intConcurrencyTest` (should be renamed or simply merged with the test) are just split from the tests because they would seem a little massive. 

The optimization test is just a plain copy at the moment that will be reworked.

Tests for concurrent stack use will be added next.

(I tested a very naive, different version first, extending the JUnit parallel test class. I could reuse `SolverBasedTest0` in that, had to sacrifice any control however and we would have to repeat every test x amount of times, x being the number of threads wanted, and pray that nothing bad happens that throws everything off. I didn't like that.)

Input is very appreciated.

(I have seen the compile problems/errors/checks etc. and will address them in the next few days)

### DIFF
--- a/src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
+++ b/src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
@@ -35,10 +35,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.junit.runners.MethodSorters;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
@@ -66,7 +64,6 @@ import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 import org.sosy_lab.java_smt.api.SolverException;
 
 @SuppressWarnings("resource")
-@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 @RunWith(Parameterized.class)
 public class SolverConcurrencyTest {
 
--- a/src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
+++ b/src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
@@ -70,6 +70,35 @@ public class SolverConcurrencyTest {
   private static final int NUMBER_OF_THREADS = 4;
   private static final int TIMEOUT_SECONDS = 30;
 
+  /**
+   * As some Solvers are slower/faster, we choose an appropriate number of formulas to solve here
+   */
+  private final Map<Solvers, Integer> integerFormulaGen =
+      Map.of(
+          Solvers.SMTINTERPOL,
+          12,
+          Solvers.CVC4,
+          16,
+          Solvers.MATHSAT5,
+          16,
+          Solvers.PRINCESS,
+          12,
+          Solvers.Z3,
+          16);
+
+  private final Map<Solvers, Integer> bitvectorFormulaGen =
+      Map.of(
+          Solvers.BOOLECTOR,
+          60,
+          Solvers.CVC4,
+          9,
+          Solvers.MATHSAT5,
+          60,
+          Solvers.PRINCESS,
+          9,
+          Solvers.Z3,
+          50);
+
   private static List<Throwable> exceptionsList;
 
   @Parameters(name = "{0}")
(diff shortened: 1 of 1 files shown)


kfriedberger:
The substring WIP (work in progress) into the title helps with avoidings pre-mature merges.
I will take a look at the rest soon.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  @After
+  public void resetExceptionsList() {
+    exceptionsList.clear();
+  }
+
+  @Test
+  public void testIntConcurrencyWithConcurrentContext() {
+    // Test concurrency of integers (while every thread creates its unique context on its own
```
kfriedberger, line 99: Better use as JavaDoc?

PhilippWendler:
Just mark the pull request as a draft, GitHub has been supporting this for quite some time now.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
```diff
@@ -0,0 +1,437 @@
...
+    // concurrently)
+    assume().withMessage("Solver %s does not support the theory of bitvectors", solverToUse())
+        .that(solverToUse())
+        .isNotEqualTo(Solvers.BOOLECTOR);
+    List<Runnable> runnableList = new ArrayList<>();
+    for (int i = 0; i < NUMBER_OF_THREADS; i++) {
```
kfriedberger, line 105: Could you extract the loop into a utility method `List<Runable> getRunnables(Runnable r)`?
It might be possible to use Java lambdas for the instanciation of a runnable, e.g., something like
<pre>
Runnable test = () -> {
try {
￼            SolverContext context = initSolver();
￼            intConcurrencyTest(context);
￼            closeSolver(context);
￼          } catch (Throwable e) {
￼            exceptionsList.add(e);
￼          }
};
</pre>
This might improve readability.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
```diff
@@ -0,0 +1,437 @@
...
+  @SuppressWarnings("resource")
+  @Test
+  public void testIntConcurrencyWithoutConcurrentContext() throws InvalidConfigurationException {
+    // Test concurrency with already present and unique context per thread
+    assume().withMessage("Solver %s does not support the theory of bitvectors", solverToUse())
+        .that(solverToUse())
+        .isNotEqualTo(Solvers.BOOLECTOR);
+    List<Runnable> runnableList = new ArrayList<>();
+    ConcurrentLinkedQueue<SolverContext> contextList = new ConcurrentLinkedQueue<>();
+    for (int i = 0; i < NUMBER_OF_THREADS; i++) {
```
kfriedberger, line 160: Splitting the loop into two (initSolvers and getRunnables) would make it clearer that initSolvers is sequential and the other is not.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
```diff
@@ -0,0 +1,437 @@
...
+        "Test " + testName + " failed with exception(s): " + exceptionsList,
+        exceptionsList.isEmpty());
+  }
+
+  /**
+   * As some Solvers are slow... very slow.... and we don't want to wait till the heat death of the
+   * universe nor a 0,1 second test for the faster ones we choose an appropriate number of formulas
+   * to solve here
+   */
+  Map<Solvers, Integer> integerFormulaGen =
```
kfriedberger, line 412: that looks like a constant: `private final Map<Solvers, Integer> INTEGER_FORMULA_GEN`

kfriedberger: Overall, the tests look fine and already cover the most critical use case, i.e., using different contexts in parallel.

kfriedberger:
> Just mark the pull request as a draft, GitHub has been supporting this for quite some time now.

Found it :-)

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
baierd: The first idea is problematic because we have different methods in the runnables every time.

I'll take a look into Lambdas but as far as i know that should be possible.

We can split the `List<Runnable> runnableList = new ArrayList<>();` into a `@BeforeClass` and reset after every test to increase readability further.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
baierd: Done

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
baierd: Its a constant now.

src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java
baierd: Done

Issue #125, opened 2018-04-23
Properly defined platform for our native binaries


PhilippWendler:
Currently it is not properly defined which versions of standard libraries we assume for the native binaries we ship, i.e., on which distributions in which versions these binaries are usable. This is bad for users and for developers.

We should decide on a set of libraries and versions that we allow as dependencies. The Python community has a similar problem for Python libraries with bundled native binaries, and has established the [manylinux2010 platform](https://www.python.org/dev/peps/pep-0571/). Binaries built for this platform are expected to run on almost all current Linux distributions. Thus I suggest we adopt this platform and the [set of libraries and versions](https://www.python.org/dev/peps/pep-0571/#the-manylinux2010-policy) it defines.

After this is decided, we should create a proper way to build our binaries automatically that ensures this compatibility, e.g., by a scripts using the official [manylinux2010 Docker image](https://www.python.org/dev/peps/pep-0571/#docker-images) or a VM (would also fix #74).

kfriedberger:
We might aim towards using `Ubuntu 18.04` as a basic platform for building and executing our solvers. Some users currently have `Ubuntu 16.04` that has LTS until 2021. This version could also be chosen as a basic system. With the upcoming change towards `Java 11` older systems (including even `Ubuntu 16.04`)  might also get out-of-scope. 

Additionally, we might want to remove any code that was especially added for older platforms.
For example, our MathSAT5 library seems to be build for and run on `Ubuntu 10.04`.
I do not expect someone to use such an old system any longer.

PhilippWendler:
I would not require something as new as Ubuntu 18.04. Java 11 is easily installable on Ubuntu 16.04 with a PPA. Users will want to use JavaSMT on StarExec, which is probably even older, and this platform is an important use case.

And what benefit would we have of requiring something newer?
If we have a properly defined platform (which we need anyway) than we need to build it using that platform, and it does not produce more effort if that platform is older or newer.

So why not use something like manylinux (currently available as [manylinux2014](https://www.python.org/dev/peps/pep-0599/))?

kfriedberger:
[...]

Issue #74, opened 2016-08-08
Ensure that we do not publish native solver binaries with unwanted library dependencies


PhilippWendler:
The current build targets for publishing solver binaries are suboptimal because they make it easy to (accidentally) build and publish solver binaries that are broken or depend on libraries on which we do not want them to depend.

- `compile-mathsat-bindings` executes `compile.sh`, but does not fail if the latter reports an undefined symbol in the binary. The output of the task instead tells the user to proceed and publish a broken binary. I assume the same is true for building the Z3 bindings.
- `compile.sh` for MathSAT asks the user to check for unwanted dependencies, but this is hidden when the ant target is used [as the documentation says](https://github.com/sosy-lab/java-smt/blob/master/Developers.md#publish-mathsat5). For Z3, there is not even such a check in the script.

PhilippWendler:
An automatic check for unwanted dependencies would be really good, because building the binaries is getting more and more complex. For example, currently we need to build MathSAT on something like Ubuntu 16.04, which has gcc >= 4.9, but we need to build Z3 on Ubuntu 14.04, because it does not have libgomp >= 4.

PhilippWendler:
This [script](https://github.com/borgbackup/borg/blob/master/scripts/glibc_check.py) could be used as inspiration for such a check.

PhilippWendler:
Other scripts that could be an inspiration are in these [LWN comments](https://lwn.net/Articles/771844/).

Of course, if we use a container solution to implement #125 this check is not so important anymore.

PhilippWendler:
AppImages have the same problems and they have some [documentation](https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages#creating-appimages-that-are-compatible-with-many-systems) with hints.

[bingcc](https://github.com/sulix/bingcc) is a GCC wrapper that might allow us to compile binaries on newer systems without introducing a dependency on newer libraries.

PR #189, opened 2020-05-12
Update MathSAT5 to version 5.6.3

kfriedberger:
The new version of MathSAT5 brings some new features and bug fixes.
This PR contains the required changes.

We should add tests for new features.


kfriedberger:
@PhilippWendler 
Currently I am stuck in the build-process. Which version of libgmp is required?

kfriedberger:
@baierd 
Could you create some unit tests with simple queries using sin/pi/log/exp/sqrt (and soon asin/pow) in the MathSAT5ApiTest?
Maybe just something like "x=0 && sin(x)=0"-->SAT and "x=1 && sin(x)=0"-->UNSAT,
just to test whether MathSAT5 uses the correct formulae, nothing with expensive computation.
Some of those functions seem to be broken in the JNI for MathSAT5 and nobody complained.

Additionally, we need to know in which domain they are working, e.g., the formula type of the return value. I assume FP, but have never checked this.
If it is always FP, even for rational input values, then we should add such methods into the FPFormulaManager and maybe move the requested tests upwards from MathSAT5APITest into FPFormulaManagerTest.

PhilippWendler:
> Currently I am stuck in the build-process. Which version of libgmp is required?

I don't know any specific requirement for the GMP version of MathSAT, you would need to check their documentation. But why do you even assume that a mismatching GMP version is the reason for "being stuck in the build process"?

kfriedberger:
I have compiled `gmp-6.2.0` (`./configure && make`) on my Ubuntu 18.04 (`gcc 7.5.0`).
Then I get the following error when running `ant publish-mathsat -D...` :
<pre>
[exec] There was a problem during compilation of
       "org_sosy_1lab_java_1smt_solvers_mathsat5_Mathsat5NativeApi.c"
[exec] /usr/bin/ld: /home/fri/workspace/solver/gmp-6.2.0/.libs/libgmp.a(realloc.o):
       relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5'
       can not be used when making a shared object; recompile with -fPIC
</pre>
I am not familiar with this error and am not sure where my problem is.
Do I need to use a different version of libgmp or gcc or something else?

PhilippWendler:
https://github.com/sosy-lab/java-smt/blob/093ac2589b21def3b38b7723e06bfc0505289a09/lib/native/source/libmathsat5j/compile.sh#L10-L13

kfriedberger:
Thanks for the hint :-)

kfriedberger:
The new MathSAT5 works fine.
There seems to be a minor problem with weird symbol names like `'` (single quote) in FP theory in the JUnit tests. I reported it to Alberto.

baierd:
> @baierd
> Could you create some unit tests with simple queries using sin/pi/log/exp/sqrt (and soon asin/pow) in the MathSAT5ApiTest?
> Maybe just something like "x=0 && sin(x)=0"-->SAT and "x=1 && sin(x)=0"-->UNSAT,
> just to test whether MathSAT5 uses the correct formulae, nothing with expensive computation.
> Some of those functions seem to be broken in the JNI for MathSAT5 and nobody complained.
> 
> Additionally, we need to know in which domain they are working, e.g., the formula type of the return value. I assume FP, but have never checked this.
> If it is always FP, even for rational input values, then we should add such methods into the FPFormulaManager and maybe move the requested tests upwards from MathSAT5APITest into FPFormulaManagerTest.

No problem, i'll take a look.

kfriedberger:
> The new MathSAT5 works fine.
> There seems to be a minor problem with weird symbol names like `'` (single quote) in FP theory in the JUnit tests. I reported it to Alberto.

Alberto responded and promised a bugfix soon.

Additionally, a problem with OptiMathSAT was reported: their own JavaBindings are outdated. They also promised a fix. As we use our own bindings, this problem does not touch us.

kfriedberger:
I am using Ubuntu 18.04, 64bit.
When compiling and linking the new version of OptiMathSAT 1.7.0.1 I get a warning.

#### cmdline: 
<pre>ant publish-optimathsat \
    -Dmathsat.path=/home/fri/workspace/solver/optimathsat-1.7.0.1-linux-64-bit \
    -Dgmp.path=/home/fri/workspace/solver/gmp-6.2.0 \
    -Dmathsat.version=1.7.0.1
</pre>

#### stdout:
<pre>
compile-optimathsat-bindings:
     [exec] Compiling bindings to OptiMathSAT
     [exec] Compiling the C wrapper code and creating the "liboptimathsat5j.so" library
     [exec] Compilation Done
     [exec] Warning: There are the following unresolved dependencies in libmathsat5j.so:
     [exec]    130: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev
     [exec] Result: 1
</pre>

The publication process finished normally,
but the resulting library crashes when loaded with:
<pre>
/usr/lib/jvm/java-11-openjdk-amd64/bin/java: symbol lookup error: /home/fri/workspace/java-smt/lib/java/runtime/liboptimathsat5j.so: undefined symbol: _ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev
</pre>

@PhilippWendler have you ever seen this warning? How can we get rid of it? Do I need to include some further headers in the binding?

PhilippWendler:
This problem results from the fact that the static library for (Opti)MathSAT does not include a statically-linked libstdc++. Our script links the (Opti)MathSAT binary against the present libstdc++, but the function `_ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev` is required by the OptiMathSAT binary and not present in the libstdc++ you used.

Running the name through `c++filt` gives the following signature:
```
std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()
```

On Ubuntu 18.04, `/usr/lib/x86_64-linux-gnu/libstdc++.so.6` has a bunch of similar functions, but not this exact overload. On Ubuntu 20.04, the function is there and marked as `GLIBCXX_3.4.26`, which makes it clear that this OptiMathSAT binary works only on systems with GCC 9.1 or newer.

There are several possible ways to solve this:
- Run our compilation and linking on a system with GCC 9.1 or newer. This allows to statically link the correct libstdc++ into the resulting binary.
- Ask for a binary compiled with an older compiler/libstdc++.
- Ask for a binary with a statically linked libstdc++.

Note that in general, C++ libraries are not guaranteed to be compatible with anything except the compiler and libstdc++ version with which there were compiled. So if we keep linking the libstdc++ into the binary ourselves, we should match the environment in which the MathSAT binary was compiled as closely as possible. For the current OptiMathSAT binary, this seems to be Ubuntu 20.04 with GCC 9.3, whereas for the current MathSAT binary this seems to be Ubuntu 16.04 with GCC 4.8. How do we currently make sure that the correct versions of Ubuntu and GCC are used?

Additionally, if we use a MathSAT binary that was built on a recent system, we need to make sure that the binary is still backwards compatible with the platforms that we want to support and does not for example reference functions of the C standard lib that were introduced only recently. This means that it is important to implement #74 and #125.

baierd:
@kfriedberger 
The `sqrt()` function uses FP and we got a test for that already. 
MathSATs log/sin/asin/pow/exp take no FP inputs but int and rational input always return rationals (pi does as well).
The methods log/asin and everything using pi ends in a endless calculation on my machine.
(They have @Ignore at the moment)

Issue #194, opened 2020-06-30
MathSAT5 does not consistently handle assumptions.


kfriedberger:
_Originally posted by @baierd in https://github.com/sosy-lab/java-smt/issues/192#issuecomment-651909235_

It seems as if MathSAT5 has partial support for assumption solving and that JavaSMT is using this API in a wrong way.
We need to either use the API in a correct way or remove the invalid code from JavaSMT.

baierd:
UnsatCoreOverAssumptions is fixed with commit [169eaf1](https://github.com/sosy-lab/java-smt/commit/169eaf1c54e6e8eaf14a22b41608a92e39cd1440) and the test is re-enabled.

kfriedberger:
The problem was just a forgotten case for a satisfiable query?
The commit looks good.

baierd:
Yes.
We could however add a `checkState(!closed)` as some solvers have it and some don't, and its probably a good idea.

I have a question though: SMTInterpol needs a empty environment for `UnsatCoreOverAssumptions` while the others don't care about that. Is that just a special case for SMTInterpol or is the idea behind `UnsatCoreOverAssumptions` always to check the assumptions only?
(Yes i know SMTInterpol does not support assumptions and we cheat a little by entering them as fixed formulas and pop them afterwards)

kfriedberger:
That is special for SMTInterpol only.

The idea behind assumption solving is that you can annotate parts of the stack (either as direct annotation `SOME_COMPLEX_EXPR :named PRED1` or as additional constraints like `PRED1 == SOME_COMPLEX_EXPR`) and then enable or disable those parts via the given list of assumptions (which contains `PRED1` or `NOT PRED1`).

Issue #192, opened 2020-06-25
New example program for printing supported features


kfriedberger:
It would be nice to have an program in the `example` package that automatically exports the solver capabilities in a simple way.

The program should iterate over all available solvers (alphabetic order) and print their data, for example as a table on cmdline:

| Solver | Version | Theories | Prover-Features |
| --- | --- | --- | --- |
| MathSAT5 | 5.6.3 | Int, Real, FP, BV, Array, ... | Interpolation, UnsatCores |
| ... |

The data should be autogenerated from the existing class instances, i.e., whenever a XFormulaManager is not null, the solver supports the theory X. This might overapproximate the solver's capabilities (if a solver provides the API, but can not solve the query), but is a good starting point for users.

baierd:
Should it include the license as well?

kfriedberger:
No. License is not auto-generated.

kfriedberger:
The programm should detect all features automatically, e.g. by catching exceptions for unsupported features or provers.

kfriedberger:
And then add it into `./runExamples.sh`, as first program.

baierd:
Probably a naive question: since license is important and in every header, why isn't there a datastructure that gives back the license of the current solver?

kfriedberger:
Licensing is a meta-topic that has nothing to do with "features".
We will soon have the Reuse integration that checks for licenses and can report them.

kfriedberger:
But yes, I agree that having the license in the table would be nice to have.

baierd:
Maybe a good opportunity will arise with the Reuse integration. Shouldn't be hard to add to this program.

baierd:
How should AllSat be handled?
It works for every solver in JavaSMT but not every solver can do it natively and we can't discern the two i think.

kfriedberger:
AllSat is not a big feature. Just ignore it.

baierd:
I tried that already because some methods have inconsistent behavior but i ran into a problem. 

Clarification on the inconsistent behavior. For example `UnsatCoreOverAssumptions()` takes an argument. If that's `null` one gets a NullPointerException as expected. Except in MathSAT5, there its fine and it works. I know that that's implementation specific and not really our fault, but weird.

kfriedberger:
It might be better to always expect a nonnull value.
Can you fix the code for MathSAT5, i.e. add some `Preconditions.checkNotNull` into the method.
And then write a JUnit test for the method with `Null` parameter.

[...]

PR #195, opened 2020-07-04
New example program for printing supported features

baierd:
This pull request is for issue #192 and #194 as it seems to be a small problem that was found by developing the example program.

As i found some inconsistencies with MathSAT those will be included here.
I've added a null check in the `UnsatCoreWithAssumptions()` method of MathSAT5 as it did not act like the other solvers.
Furthermore i added a test for this behaviour.
The method `UnsatCoreWithAssumptions()` itself seems to have a problem as seen in #194 and will be checked as well.

A first iteration of the program is included.
Main problem at the moment is that the output of the example program in the script and therefore the terminal puts line breaks on its own depending on the width of the terminal.
@kfriedberger is there a easy method to fix this?


kfriedberger:
The table is quite nice and clearly written, I like it. It is just very broad.
A possible solution would be to use multiple lines per cell, but that might be quite tricky to implement in a simple and understandable way.
And perhaps, we should aim for a more object-oriented approach instead of using lists of lists of Strings.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+
+  /**
+   * Checks the infoArray for the length of the given column.
+   *
+   * @param columnToCheck column of the infoArray you want the length for.
+   * @return length of String in column, but at least 8 (max length of the descriptor Strings).
+   */
+  private int lineLength(List<List<String>> infoArray, int columnToCheck) {
+    if (columnToCheck > 3 || columnToCheck < 0) {
+      return 8;
```
kfriedberger, line 300: This looks like a constant MIN_CELL_WIDTH.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+    int length = 8;
```
kfriedberger, line 302: MIN_CELL_WIDTH?

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+    List<String> features = new ArrayList<>();
+
+    // Optimization: Will throw UnsupportedOperationException in creation of prover if not
+    // available.
+    try (OptimizationProverEnvironment prover =
+        context.newOptimizationProverEnvironment(ProverOptions.GENERATE_MODELS)) {
+      if (prover != null) {
+        features.add("Optimization");
+      }
+    } catch (UnsupportedOperationException e) {
```
kfriedberger, line 175: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // Interpolation: Will throw UnsupportedOperationException in creation of prover if not
+    // available.
+    try (InterpolatingProverEnvironment<?> prover =
+        context.newProverEnvironmentWithInterpolation(ProverOptions.GENERATE_MODELS)) {
+      if (prover != null) {
+        features.add("Interpolation");
+      }
+    } catch (UnsupportedOperationException e) {
```
kfriedberger, line 185: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+
+    // UnsatCore: throws UnsupportedOperationException if not available.
+    // TODO: check behavior with null argument properly
+    try (ProverEnvironment prover =
+        context.newProverEnvironment(ProverOptions.GENERATE_UNSAT_CORE)) {
+      if (prover.getUnsatCore() != null) {
+        features.add("UnsatCore");
+      }
+    } catch (UnsupportedOperationException e) {
+    } catch (Exception e) {
```
kfriedberger, line 196: Why is this needed?

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+      features.add("UnsatCore");
+    }
+
+    // UnsatCoreOverAssumptions: throws NullPointerException if available.
+    try (ProverEnvironment prover =
+        context.newProverEnvironment(GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS)) {
+      prover.unsatCoreOverAssumptions(null);
+    } catch (UnsupportedOperationException e) {
+    } catch (NullPointerException e) {
```
kfriedberger, line 205: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+                  public List<List<BooleanFormula>> getResult() {
+                    return models;
+                  }
+                },
+                ImmutableList.of())
+            != null) {
+          features.add("AllSAT");
+        }
+      } catch (UnsupportedOperationException e) {
+      }
```
kfriedberger, line 233: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // We don't care about the return value, just that it doesn't throw an
+    // UnsupportedOperationException.
+    try (ProverEnvironment prover = context.newProverEnvironment(ProverOptions.GENERATE_MODELS)) {
+      try {
+        prover.isUnsatWithAssumptions(ImmutableList.of());
+        features.add("Assumptions");
+      } catch (UnsupportedOperationException e) {
+      }
```
kfriedberger, line 243: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    // UnsatCore: throws UnsupportedOperationException if not available.
+    // TODO: check behavior with null argument properly
+    try (ProverEnvironment prover =
+        context.newProverEnvironment(ProverOptions.GENERATE_UNSAT_CORE)) {
+      if (prover.getUnsatCore() != null) {
+        features.add("UnsatCore");
+      }
+    } catch (UnsupportedOperationException e) {
```
kfriedberger, line 195: add a comment, why the exception can be ignored.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
```diff
@@ -0,0 +1,311 @@
...
+   * 4.
+   *
+   * @param solver to check for information. Taken from Solvers enum only.
+   * @return String[] of length 4 of the solver you entered. Array content as String in the
+   *     following order: SolverName, SolverVersion, SolverTheories, SolverFeatures. String[] length
+   *     0 if invalid solver.
+   */
+  private List<String> getSolverInformation(Solvers solver)
+      throws SolverException, InterruptedException {
+    List<String> info = new ArrayList<>();
```
kfriedberger, line 134: An `info` object with named members would be nicer .... perhaps.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
baierd: Because i wanted minimal examples to check for `UnsupportedOperationException` and didn't yet check how the method `getUnsatCore()` behaves on an empty stack. This is subject to change.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
baierd: `info` object in a Class?
Could you explain further please?

baierd: > The table is quite nice and clearly written, I like it. It is just very broad.
> A possible solution would be to use multiple lines per cell, but that might be quite tricky to implement in a simple and understandable way.
Yes i know. But i know that its possible to "unlock" the terminal/shell so that you can scroll and that there are no forced line breaks, i just dont know how yet.

> And perhaps, we should aim for a more object-oriented approach instead of using lists of lists of Strings.
Done

baierd:
I know its very broad and thats a problem in the terminal. But i know that its possible to stop the terminal from forcing line breaks somehow, i've seen it. Either i will find out or change the format.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
kfriedberger: The SolverInfo class would allow to split data retrieval and data dumping. Just an idea.
The following draft might help:
```
/** data class containing the info for the solver and can print it */
class SolverInfo {
  - Solvers solver;
  - String version;
  - List<String> supportedTheories;
  - List<String> features;
}
/** helper class for wrapping the contents of a row in the table  */
class Row {
  - List<String> lines;
  - int minNumOfLines;
  - List<Integer> columnWidths; // needed for producing a new empty line with seperators "|   |   |   |"
  + appendCell(List<String> lines); // append cell-lines to lines. Add new line, if number of cell lines is too large.   
  + toString(); // dump nice row, include terminal line as last line: "|---|---|---|"
}
+ List<String> cellToRow(String cellContent, int exactLengthOfCell); // returns a list of lines, formatted to match exact length or less space, does no contain seperator "|"
```

kfriedberger:
Some terminal can disable line breaks, some do not.
We dot want to rely on terminal-specific code, because that might then be broken on windows... :-)

For me, the table would also be ok, as it is now, and it helps a lot as documentation.

baierd:
Do you know a good width that works in most cases?
Something similar as the other examples?

kfriedberger:
Lets try, how Github layouts the table on a normal monitor (table written in MarkDown, with ", " as seperator in theory and feature cell):

| Solver | Version | Theories | Features |
| --- | --- | --- | --- |
| MATHSAT5 | MathSAT5 version 5.6.3 (973d615f065f) (May 12 2020 08:45:26, gmp 6.1.0, gcc 4.8.5, 64-bit, reentrant) | Integer, Bitvector, Rational, Float, Array, UF | Optimization, Interpolation, UnsatCore, UnsatCore /w Assumption, AllSAT, Assumptions |
| SMTINTERPOL | SMTInterpol 2.5-604-g71e72f93 | Integer, Rational, Array, UF | Interpolation, UnsatCore, UnsatCore /w Assumption, AllSAT, Assumptions |
| Z3 | Z3 4.8.8.0 | Integer, Bitvector, Rational, Float, Array, Quantifier, UF | Optimization, UnsatCore, UnsatCore /w Assumption, AllSAT, Assumptions |
| PRINCESS | Princess (2020-05-01 (assertionless)) | Integer, Bitvector, Array, Quantifier, UF | Interpolation, UnsatCore, AllSAT, Assumptions |
| BOOLECTOR   | Boolector 3.2.1 | Bitvector, Array, Quantifier, UF | AllSAT, Assumptions |
| CVC4 | CVC4 1.8-prerelease | Integer, Bitvector, Rational, Float, Array, UF, Seperation-Logic | UnsatCore, AllSAT, Assumptions |

Overall width for columns:
| 15 (full name) | 40 (split at space) | 25 (split at space) | 25 (split at space) |
| --- | --- | --- | --- |

Would this be possible?

baierd:
Looks good.
I'll give it a try.

baierd:
This table should be fine now.
Code needs to be cleaned up though.

kfriedberger:
Thanks for the nice table and formatting code.
I applied some cleanup. It looks fine now.

kfriedberger:
@baierd do you have any more ideas for improvement or can I merge it?

baierd:
If its ok i'd like to try to check if AllSat is overwritten in the solver and check the behaviour of UnsatCore on an empty stack.

kfriedberger:
When merging the master into this branch, Yices2 fails without a clear reason. We need to fix this in the Yices2-bindings.

baierd:
OK i'll take a look.

src/org/sosy_lab/java_smt/example/ListAllSolversWithFeatures.java
baierd: `UnsatCore()` throws different Exceptions based on the solver used (native Exceptions). I added a comment.

baierd:
I still have a problem with `AllSat()`. If i use reflections to get the method or class i run into basicly the same problem. The class of a prover initialized with `GENERATE_ALL_SAT` is not uniform. For example SMTInterpol has a `ProverWithAssumptionsWrapper` and overrides `AllSat()` while CVC4 returns the same class and does not support it.
You can't check for thrown Exceptions either as the `UnsupportedOperationException` is not in the method declaration.

As the current check for AllSat doesn't work and you said its not important, i would simply remove it.

kfriedberger:
Ok, just remove AllSAT from the table.

baierd:
Thats maybe a stupid question, but shouldn't a prover thats initialized with `GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` be able to perform UnsatCore, UnsatCore /w Assumptions and also Assumptions (without UnsatCore). 
If yes we could combine that testcase into 1 prover object.

kfriedberger:
Yes, `UNSAT_CORE_OVER_ASSUMPTIONS` implies the other features. But it is not the other way around, e.g., `UNSAT_CORE` might be supported, but the rest is not.

baierd:
Yices2 works now. I had to reorder the way the features are checked.

Yices had a problem with the config it uses to create a new prover (yices2 context). Once you use `UnsatCore `once you may not use DPLL or push/pop again. If you do it ends in an `IllegalArgumentException `(native code returns `CTX_INVALID_PARAMETER_VALUE` for error in changing the config). 
Is UnsatCore considered non-linear arithmetics? Because in that case Yices2 defaults to MCSat which does not support push/pop. If not i don't know why.

kfriedberger:
Is the reordering now hiding a bug in Yices2?
Is creating and deleting a prover in yices not cleaning up properly and leaving behind traces/options that are unchangable?

UnsatCores have nothing to do with non-linear arithmetics, they are an orthogonal feature.
We require push/pop, and that should be available for other backend solvers.

baierd:
Possibly.
I tried some things:
-creating a static config that is not deleted with push/pop works fine with this program, but fails in the tests severly!
  (Probably because non-linear arithmetics needs another config acording to the API. But i don't know why our tests don't fail 
  then as we use the config with DPLL and push/pop every time. Maybe it simply overrides the settings after using it?)
-calling UnsatCoreWithAssumptions before UnsatCore is fine, the other way fails. Same with Assumptions and UnsatCore.

kfriedberger:
Looks good now. Anything remaining for this branch?

baierd:
I think this is fine. You can merge.

PR #196, opened 2020-07-06
Change licensing information to be reuse compatible.

kfriedberger:


.reuse/templates/header.jinja2
```diff
@@ -0,0 +1,11 @@
+This file is part of JavaSMT,
+JavaSMT is an API wrapper for a collection of SMT solvers.:
```
PhilippWendler, line 2: These two lines are not a correct sentence, and the punctuation `.:` is also weird. This also affects all changed files.

src/org/sosy_lab/java_smt/solvers/boolector/BtorOption.java
```diff
@@ -1,3 +1,11 @@
...
+// JavaSMT is an API wrapper for a collection of SMT solvers.:
+// https://github.com/sosy-lab/java-smt
+//
+// SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
 package org.sosy_lab.java_smt.solvers.boolector;
 
 /** We keep this synchronized with "btortypes.h". */
```
PhilippWendler, line 11: This looks like code that is derived from code in `btortypes.h`? Depending on the license of that file, we might need to keep the original copyright and license notices.

src/org/sosy_lab/java_smt/solvers/boolector/BtorJNI.java
```diff
@@ -1,3 +1,11 @@
+// This file is part of JavaSMT,
+// JavaSMT is an API wrapper for a collection of SMT solvers.:
+// https://github.com/sosy-lab/java-smt
+//
+// SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
 /* ----------------------------------------------------------------------------
  * This file was automatically generated by SWIG (http://www.swig.org).
```
PhilippWendler, line 10: What file is this based on? Generated code typically needs to be considered derived from the input of the generator (e.g., a grammar in case of parser generators). Depending on the license of the original file, we might need to keep the original copyright and license notices.

src/org/sosy_lab/java_smt/example/AllSatExample.java
```diff
@@ -1,3 +1,11 @@
+// This file is part of JavaSMT,
+// JavaSMT is an API wrapper for a collection of SMT solvers.:
+// https://github.com/sosy-lab/java-smt
+//
+// SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
```
PhilippWendler, line 7: Are these and other examples intended to be just for looking at them, or are they intended as template code that can be copied into own projects and changed/extended there?

If the latter is desired, we might want to use a license with less restrictions.

kfriedberger: Of course this code can be copied and modified. Which license is appropriate here? CC?

PhilippWendler: What license do you mean by CC? This is not an SPDX identifier, nor a common abbreviation.

Of course the copyright holder needs to decide this.

The minimal amount of restrictions (i.e., none) are contained in CC-0 and Unlicense. Both are widely used, although the OSI [recommends not using them](https://opensource.org/faq#public-domain).

One solution would be to use something like `Unlicense OR MIT`. This way, users can pick whether they want no restrictions or prefer an OSI-approved license. Of course we could make the disjunction even broader if we want to.

PR #202, opened 2020-07-29
Add Windows Support for some Solvers

kfriedberger:
Currently, JavaSMT can only be used directly on Linux.
Windows support for several solvers is missing.
This PR aims at improving the availability of solvers like MathSAT5 and Z3 on Windows machines.


kfriedberger:
Intermediate report about MathSAT5 for Windows (64 bit):
- MathSAT5 works on Windows, hurray.
- cross-compiling from Ubuntu for Windows is possible and works very simple. I will provide a script for this soon.
- Windows-dlls can not be merged or linked together into one single file (or can they?). Windows expects each dependency to be already available. So we need to load `mpir.dll`, then `mathsat.dll`, then our `mathsat5j.dll`. Our Java code will be updated soon.
- Compiler warnings in our JNI glue code must be fixed. Pointer type `long int` seems to not match the Windows pointer type. Really??
- Callbacks via pointers are possible, but sometimes crash.
- FP-support is available, but buggy? Need to report.
- Model evaluation is buggy? Need to report.
- The Windows version of MathSAT5 is not threadsafe, no reentrant version available. Executing a asynchronous termination-callback might be problematic.
- But the rest seems to work fine.

kfriedberger:
#### Current status:
Z3 and MathSAT5 work quite nice on Windows.
For Z3, we directly use the available binaries.
For MathSAT5, we cross-compile the JNI bindings on Ubuntu for Windows64.

kfriedberger:
The new release of MathSAT5 looks fine and Z3 also seems to work.
Lets reintegrate this branch and continue with further work in the master branch.

Issue #201, opened 2020-07-25
Drop or improve Maven support?


kfriedberger:
### Status Quo:
We push releases of JavaSMT regularly into the Maven repository.
However, most of current maven release are broken (for quite a long time, since d066c19dbf489fad5152f9c6e66364179a5ba431) due to a missing dependency for java-cup (required by Princess).

Side note: Princess actually request a different version of java-cup, but works with the TUM-version as well. This just makes Maven complex, as there is no TUM-version of java-cup in the Maven repository.

### To discuss:
It is nice to have to provide a simple way of using JavaSMT via Maven. However, is there any benefit for us?
Nobody ever complained about the broken dependency, thus the number of users might be small (cause or effect?).
Even for our own software (such as sosy-common), missing Maven dependencies are only noticed when we check it.
There is no CI (#79) and other issues related to Maven have been open for a long time (#88, #67).

- The easiest way would be dropping it completely.
- The more complex way would be a complete revisit of Maven and a proper integration, i.e., such that more solvers are available.

proteusGIT:
I am currently investigating whether to use java-smt.
Since my project already uses maven and not ivy, this would be the best choice for me.
I installed it yesterday and used Z3 on some formula and everything seemed to work (at least for Z3).

kfriedberger:
Our plans are not yet fixed for Maven. This entry is just a status update.

### Cons:
The maintenance effort to support more build systems over a longer time span is quite high and has no direct benefit for our own projects. We need to keep knowledge about Maven in our team.

### Pro:
Maven is one of the most-used build systems for Java software.
Thus, for many other developers, proper Maven support is beneficial.

Lets start with a list of steps that would be needed for a better Maven support:
- [ ] fix the missing JavaCup, e.g., upload it in the namespace of `sosy-lab` or switch to the previous version.
In either case, we need to update the automatically created `pom.xml` file.
- [ ] upload at least a newer Z3 into the Maven repository. The old one there is outdated, unsupported, in the namespace of `sosy-lab`, and is reported by search engines as first match for `z3 maven`. This might be an easy task, as soon as the pom-template for Z3 is written.
- [ ] as soon as we know how to deal with Z3, other solvers should be easy. Our priority might be MathSAT and CVC4.
[...]

PR #211, opened 2020-12-31
Extend Maven Support

kfriedberger:
This PR aims to solve #201 by improving the support and documentation for Maven when using JavaSMT.


baierd:
ocbea16 is my first draft of a documentation/starting point that should help every dev/maintainer if an Maven related issue should arise.

kfriedberger:
@baierd
From the Maven point it looks as good as possible before merging into trunk.
Is there anything very important that you still want to fix in this branch?
Otherwise I would merge this branch soon and then publish a new release, including a Maven release.
That release would be provided to a potentially larger user group.

baierd:
Looks good from my point of view.
The only thing missing that i'd like to add is Windows support for Maven. But that can be done on another branch.

kfriedberger:
Isn't Maven on Windows the same as on Linux?
I already published the Windows binaries for Z3 and they seemed to work.

Ok, I will merge this PR, then update some solvers (Z3 had a new version yesterday) and then provide a full Maven release for extensive testing.

PR #227, opened 2021-05-06
Gradle example project

baierd:
The Gradle example is located in the doc folder besides the Maven examples and supports Gradle 6.x and 7.

It offers several SMT solvers like SMTInterpol, Princess, Z3, Boolector and CVC4. I did not add Mathsat5 at this time as the naming scheme in Maven is currently inconsistent, but this should be fixed in the next MathSAT5 Maven release (as we use a new script now, compared to the current MathSAT release). Yices2 is not added, because we would have to split the project again.

The project tests each solver with the Intager based Sudoku example, except Boolector as it does not support Ints. Furthermore, as the standard run target of the Gradle project, the `SolverOverViewTable` will be printed.

To test the project, enter the Gradle example project folder and enter `gradle build` to build and test and then `gradle run` to run the `SolverOverViewTable`.

There are basicly 2 questions still open: 
- Should a Bitvector based test (i.e. a Bitvector Sudoku) be incorporated and used for Boolector or simply use the Boolean based Sudoku? In the first case it would be better to create a new issue for this and add a `BitvectorSudokuSolver` to the Sudoku example. 

- Is it better to merge now and add a seperate Yices2 Gradle project later, or wait and add it now.


doc/Example-Gradle-Project/gradle/wrapper/gradle-wrapper.jar.license
```diff
@@ -0,0 +1,7 @@
+This file is part of CPAchecker,
```
PhilippWendler, line 1: First, do we really need a JAR file checked in here?

Second, this seems to a file from Gradle, but its `.license` file claims something completely different. Very likely copyright and license info are wrong?

doc/Example-Gradle-Project/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
+
+# SPDX-FileCopyrightText: 2015 the original author or authors
+# SPDX-License-Identifier: Apache-2.0
+
+##############################################################################
+##
+##  Gradle start up script for UN*X
```
PhilippWendler, line 8: Is this file copied from somewhere? If yes, we should document this for maintainability, for example because we might want to update this at some point.

Also, some licenses require that we need to document whether the file was copied unmodified or had changes, this also needs to be thought of.

PhilippWendler (changes requested):

doc/Example-Gradle-Project/gradlew
PhilippWendler: Thanks for the explanation. This is also not common for other build systems AFAIK. Typical ant, maven, cmake, npm, etc. projects all require you to have the build system installed locally.

And specifically, for an example project that is intended to show to Gradle users (which are likely to have Gradle already) how to use JavaSMT and is not really intended for anyone to check out and work on it, there seems even less reason to add this wrapper than for normal Gradle-based projects.

baierd: This file is not supposed to be uploaded anyway. My bad.

doc/Example-Gradle-Project/build.gradle
```diff
@@ -0,0 +1,146 @@
...
+    url "https://www.sosy-lab.org/ivy"
+    patternLayout {
+      artifact "/[organisation]/[module]/[classifier]-[revision].[ext]"
+      //ivy "/[organisation]/[module]/ivy-[revision].xml"
+    }
+    metadataSources {
+       artifact()
+    }
+  }
+  // SMT-Solver Princess repo
```
kfriedberger, line 38: Is this repository still used? Princess switched to a puiblic Maven repo some time ago.

doc/Example-Gradle-Project/build.gradle
```diff
@@ -0,0 +1,146 @@
...
+// You can manage the JavaSMT dependencies via a config if you need to split them from your regular dependencies
+dependencies {
+  // JUnit 4
+  testImplementation 'junit:junit:4.13'
+
+  // Retrieve JavaSMT with Maven and transitive dependencies via a pom.xml
+  implementation 'org.sosy-lab:java-smt:3.11.0'
+
+  // Retrieve Z3 with Maven but without a pom.xml, just the artifacts
+  runtimeOnly group: 'org.sosy-lab', name: 'javasmt-solver-z3', version: '4.8.10', classifier: 'com.microsoft.z3', ext: 'jar'
```
kfriedberger, line 61: Is there a way to use a central variable for the version numbers of SMT solvers?
Maybe something like `z3.version=4.8.10`.

baierd: Yes, i've added that in the latest commit.

doc/Example-Gradle-Project/build.gradle
baierd: Seems like it as it has the newest version, but i've switched to Maven central for Princess in the latest commit.

doc/Example-Gradle-Project/build.gradle
```diff
@@ -11,6 +11,19 @@ apply plugin: 'application'
...
+def javaSMTYices2Version = '3.10.0'
+def princessVersion = '2021-11-15'
+def smtInterpolVersion = '2.5-916-ga5843d8b'
+def boolectorVersion = '3.2.2-gcc3a7091'
+def z3Version = '4.8.10'
+def yices2Version = '2.6.2-396-g194350c1'
+def cvc4Version = '1.8-prerelease-2020-06-24-g7825d8f28'
+def mathsat5Version = '5.6.6'
+def optiMathsat5Version = '1.7.1-sosy0'
+def junit4Version = '4.13'
```
lorenzleutgeb, line 25: You could put (a dense representation of) the solver dependencies in a map. The syntax is a bit lighter to the eye. If you add a little more information than just the versions, you even iterate over the map in the `dependencies` block.

```groovy
def solverVersions = [
'yices2': '3.10.0',
'princess': '2021-11-15',
// ...
];
```

baierd: I like your map better ;D as you can see in the latest commit. Thank you!

You can even use it like this: solverVersions.yices2

doc/Example-Gradle-Project/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,13 @@
...
+# an API wrapper for a collection of SMT solvers:
+# https://github.com/sosy-lab/java-smt
+#
+# SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-bin.zip
```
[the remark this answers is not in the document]
baierd: I updated and built the examples with 7.3.3 now. As this version complains about some things i did because they will become deprecated in the future i think this should be fine now? (I obviously updated everything that was complained about)
+ we removed all wrapper files anyway ;D

kfriedberger:
It looks like the example project for Gradle is sufficiently documented and can be merged.

PR #246, opened 2021-08-09
remove code for using a separate class-loader for Z3.

kfriedberger:
@PhilippWendler  Could you give a comment whether this commit is harmful or harmless?

From my side, it looks good and works as expected (tested for JavaSMT and CPAchecker).
Is there a special case where the separate class-loader would really be helpful?


kfriedberger:
@PhilippWendler any comment? I would like to merge this on weekend.

src/org/sosy_lab/java_smt/solvers/z3/Z3SolverContext.java
```diff
@@ -119,19 +120,22 @@ public static synchronized Z3SolverContext create(
...
+        NativeLibraries.loadLibrary("libz3");
+        NativeLibraries.loadLibrary("libz3java");
       } catch (UnsatisfiedLinkError e2) {
         e1.addSuppressed(e2);
         throw e1;
       }
     }
 
+    // disable Z3's own loading mechanism, see com.microsoft.z3.Native
+    System.setProperty("z3.skipLibraryLoad", "true");
```
PhilippWendler, line 137: I could not find any documentation on this. Is there some? Is the behavior of this option actually guaranteed by Z3? Is it guaranteed to stay? Relying on some internal undocumented and unofficial option could bite us in the future.

kfriedberger: I could also not find further documentation or a public statement, there is only documentation in the code of Z3.

The build-scripts of Z3 insert the corresponding Java code into the class `com.microsoft.z3.Native`:
https://github.com/Z3Prover/z3/blob/34885562e0debe961b73935c6d02e71d6628238d/scripts/update_api.py#L547-L555
(commited in https://github.com/Z3Prover/z3/commit/7f0b5bc129988225c831314202608a26a18170f9)

At least one other project from Github relies on this functionality:
https://github.com/lorenzleutgeb/atlas/blob/f509bcd1b5c7081eb411c016e64b8d1bca45627e/src/main/java/xyz/leutgeb/lorenz/atlas/util/Z3Support.java#L28

PhilippWendler: > there is only documentation in the code of Z3.

I don't see any documentation there?

And without an official statement I would recommend not using that. For example, right now the code will accept any value for the property, but what if future revisions change that and require specific values?

So asking the Z3 developers to make that option officially documented would be good.

kfriedberger: :heavy_check_mark: Documentation requested: https://github.com/Z3Prover/z3/pull/4667#issuecomment-898265754

kfriedberger: :heavy_check_mark: Documentation was added: https://github.com/Z3Prover/z3/pull/5477

PR #249, opened 2021-08-18
Update SMT solver Z3 to version 4.8.17

kfriedberger:
Our currently included version `4.8.10` of Z3 works nicely on several operating systems.
Since a few weeks there is a newer version `4.8.12` and the developers have changed some bits and bytes:
- on Linux (first test: Ubuntu 21.04) I have segmentation faults with operations on FloatingPoint formula, maybe because of `libgmp`.
- on Windows, there are now more dependencies towards the `Visual C++ Redistributable` package, and the DLL files are included along `libz3.dll`. We need to decide whether we can also deliver them as a dependency within JavaSMT (upload them to Ivy and Maven, nicer for the user, list of dependencies gets longer, we need to look at the corresponding license) or just require the user to install the corresponding system package ([`Microsoft Visual C++ Redistributable for Visual Studio 2015-2019`](https://support.microsoft.com/en-us/topic/the-latest-supported-visual-c-downloads-2647da03-1eea-4433-9aff-95f26a218cc0)).

As MathSAT5 for Windows also depends on `Microsoft Visual C++ Redistributable for Visual Studio` (but for `2014`), we might think of a general strategy of including those dependencies for Windows users. Those dependencies do not change as often as solvers, so having them as a separate dependency could be a good idea.


kfriedberger:
Until the questions from above are solved, I do not push the new Z3 library into our Ivy or Maven repository. Thus the CI is expected to fail.

kfriedberger:
Due to [uninitialized variable usage in Z3](https://github.com/Z3Prover/z3/commit/123c44639564455ff09490446b8c0abbca807533), the update to Z3 version 4.8.12.0 will be skipped.
We will consider the next possible version of Z3. I will keep this PR open for this.

kfriedberger:
There is a new version of Z3, namely `4.8.13`.
This version seems to solve the issue from https://github.com/Z3Prover/z3/issues/5487.
However, the optimization prover test returns an invalid model for a query that is similar to this:
```
RationalFormula x = rmgr.makeVariable("x");
prover.addConstraint(rmgr.lessThan(x, rmgr.makeNumber(1)));
int handle = prover.maximize(x); // this should cause a value of (1 - epsilon)
assertThat(prover.check()).isEqualTo(OptStatus.OPT);
Rational epsilon = Rational.ofLongs(1, 1000000L);
Rational approximation = prover.upper(handle, epsilon); // <-- return an approximation of "0", which is too far away from "1".
```
We need to look into this.

kfriedberger:
Reported the bug for Z3: https://github.com/Z3Prover/z3/issues/5720

kfriedberger:
The versions 4.8.12/13/14 are uploaded into our Ivy and Maven repositories.
JavaSMT will still stick with 4.8.11 for now, because of the mentioned bug in optimization.
A user can easily override the version of the Z3 dependency in the Ivy/Maven/Gradle file.

kfriedberger:
The versions 4.8.15/16/17 are uploaded into our Ivy and Maven repositories.
A user can easily override the version of the Z3 dependency in the Ivy/Maven/Gradle file.

It seems as if Z3 in version 4.8.15 fixes the mentioned bug in optimization.
We will investigate this (soon) and update the dependency of JavaSMT.

PR #250, opened 2021-08-24
String formulae

serras:
This PR adds initial support for strings, the only available operations are creation of constants and equality. A concrete implementation of the theory for Z3 is also provided.

Note: there are some attempts at implementing the theory for SMTInterpol. However, [this comment](https://github.com/ultimate-pa/smtinterpol/pull/130#issuecomment-904416305) suggests that the string theory is not really supported. For that reason, those changes have been reverted (but you may want to squash everything).


src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -135,7 +138,8 @@ protected AbstractFormulaManager(
       @Nullable
           AbstractQuantifiedFormulaManager<TFormulaInfo, TType, TEnv, TFuncDecl> quantifiedManager,
       @Nullable AbstractArrayFormulaManager<TFormulaInfo, TType, TEnv, TFuncDecl> arrayManager,
-      @Nullable AbstractSLFormulaManager<TFormulaInfo, TType, TEnv, TFuncDecl> slManager) {
+      @Nullable AbstractSLFormulaManager<TFormulaInfo, TType, TEnv, TFuncDecl> slManager,
+      @Nullable StringFormulaManager strManager) {
```
kfriedberger, line 142: This looks strange: All managers are "Abstract*Managers" and the new StringFormulaManager is not.
This might be an effect of the small number of solvers supporting String theory.
For future extensions, this could be changed.

src/org/sosy_lab/java_smt/api/StringFormulaManager.java
```diff
@@ -0,0 +1,34 @@
...
+   * <p>Please make sure that the given name is valid in SMT-LIB2. Take a look at {@link
+   * FormulaManager#isValidName} for further information.
+   *
+   * <p>This method does not quote or unquote the given name, but uses the plain name "AS IS".
+   * {@link Formula#toString} can return a different String than the given one.
+   */
+  StringFormula makeVariable(String pVar);
+
+  BooleanFormula equal(StringFormula str1, StringFormula str2);
+}
```
kfriedberger, line 34: Is it useful (and possible) to already include more String-based methods here?
According to [SMTLIB2 String theory](http://smtlib.cs.uiowa.edu/theories-UnicodeStrings.shtml) the following methods might be possible: `concat`, `length`, and `compare` (maybe represented as `lessThan`/`greaterThan`).
Regular expressions might still be out of scope, but remain as future work (not in this PR).

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -193,12 +196,16 @@ public Long extractInfo(Formula pT) {
       case Z3_RE_SORT:
       case Z3_UNKNOWN_SORT:
       case Z3_UNINTERPRETED_SORT:
-        // TODO: support for remaining sorts.
-        throw new IllegalArgumentException(
-            "Unknown formula type "
-                + Native.sortToString(z3context, pSort)
-                + " with sort "
-                + sortKind);
+        if (Native.isStringSort(z3context, pSort)) {
```
kfriedberger, line 199: Is there an enum value for `Z3_STRING_SORT`? I could only find `Z3_RE_SORT`, is this the same sort and would match here, too?

src/org/sosy_lab/java_smt/test/ModelEvaluationTest.java
```diff
@@ -144,4 +144,14 @@ public void testGetBooleansEvaluation() throws SolverException, InterruptedExcep
...
+  @Test
+  public void testGetStringsEvaluation() throws SolverException, InterruptedException {
+    requireStrings();
+    evaluateInModel(
+        smgr.equal(smgr.makeVariable("x"), smgr.makeString("hello")),
+        smgr.makeVariable("y"),
+        Lists.newArrayList(null, "hello"),
+        Lists.newArrayList(null, smgr.makeString("hello")));
+  }
 }
```
kfriedberger, line 157: This is currently the only test for String theory in this PR.
Would it be possible to provide a JUnit test for `StringFormulaManager` containing a few more (small) tests?

kfriedberger (changes requested):
I like this PR. Overall it is clean and simple.
It would be nice to merge after fixing a few small points:
- the set of supported methods from String theory. Maybe we should add some more operations, like `concat`, `length`, and `lessThan`/`greaterThan`.
- missing tests.

kfriedberger:
Current status of this PR:
- Z3 and CVC4 support an sufficient range of features in String and Regex theory.

The following parts are still missing or incomplete or can be improved:
- a proper definition of the method `allChar`:
  It is unclear whether SMT solvers support the full range of Unicode or just Ascii, as defined in SMTLIB2 (maybe with more escaped characters?). For the current SMT solvers we have the following situation:
  - Z3 supports Unicode ranges, but returns unknown when solving with full range.
  - CVC4 seems to only accept Ascii, and crashes with full Unicode range.

  I would be in favour of removing the method `allChar` completely. We already provide a way to specify ranges of chars via the API. This should be sufficient.
- model evaluation: currently we have not yet any support and tests for model evaluation.
- more tests, e.g. for String comparison via LT/LEQ/GT/GEQ.
- optional: a way to translate a real Java regex into SMTLIB2 regex. Currently all regex formulas are build from basic parts from scratch.
- optional: a nice example, maybe encode a [regex crossword puzzle](https://regexcrossword.com/) with JavaSMT :-)

serras:
I've removed `allChar` and added conversions from regular expression classes (`Pattern` and `Matcher`).

I don't know very well how to check the current model evaluation. As far as I understand, it should work on Z3, but I don't know how to add more tests in the corresponding file.

As a third question: what do you mean as conversion between the Int and String theories?

kfriedberger:
Hi @serras 

1) The new API methods for Java regex might not work as expected. They will just return a String that is matches "literally", but not "structurally". If implemented, this feature might need a way to reconstruct a regex based on basic elements from SMTLIB.
I did not list the feature for now, but maybe as a future idea (sorry, should have added this as comment :-) ). This (if missing in the implementation) does not block the PR from merging.

2) There seem to be methods in Z3 and CVC4 to convert an expression from Int theory to String theory and also back, if the String consists of digits). These operations should be similar to the existing conversion between Int to Bitvector theory. For example, this could allow a query like `(= (itos (+ 1 2) (charAt "345" 0)))`.

Best.

kfriedberger:
I will take a look at the model evaluation in the next days.
The basic idea is to query the SMT solver for a model and have a way to determine the vairables in the model.
This should also support String values.

kfriedberger:
With the support for formula visitation, this PR looks sufficient for a merge.
I will review the changes again in the afternoon and plan to merge then.

kfriedberger (approved):

Issue #260, opened 2022-01-26
Provide access to solver-internal statistics


kfriedberger:
Several solvers provide some kind of internal statistics for clauses, conflicts, solving time, etc.
There are users requesting access to this data through JavaSMT.

The overall format depends on the solver. We might choose to export data in a default key-value-mapping.
Some solvers provide statistics per context (e.g., MathSAT), other solvers provide statistics per prover (e.g., SMTInterpol and Z3).
We need to take a look at the usefulness and provide either one or two new methods in the API.

For MathSAT, it seems as if the corresponding method was never included in compilation and the native C-bindings need to be touched.

For Boolector, it seems as if statistics are written to stdout instead to String. This is bad behaviour. Maybe we can redirect the stream in the native C-bindings (otherwise we ignore those statistics).

baierd:
When compiling Boolector you can also set the `--time-stats` flag to enable compilation with time statistics.

baierd:
I've added a new C function + JNI method for your use case in Boolector.

kfriedberger:
closed with #261

PR #261, opened 2022-01-27
Provide access to solver-internal statistics (see #260)

kfriedberger:


kfriedberger:
Solvers like Princess or Yices do not provide statistics.
We might merge this merge request  soon.

PR #265, opened 2022-03-31
Improve handling of division and modulo for BV theory.

kfriedberger:


kfriedberger:
Princess in version 2022-03-16 and earlier has an invalid computation of SDIV for negative values.
For example, the formula `a=10 && a/-3=-3` encoded in BV theory with a bitsize of 8 is reported as UNSAT.
The SMTLIB for this example is:
```
(declare-fun a () (_ BitVec 8))
(assert (and (= a (_ bv10 8)) (= (bvsdiv a (_ bv253 8)) (_ bv253 8))))
```
The bug is reported and will be fixed in the next release of Princess.

PR #270, opened 2022-06-18
Update MathSAT5 from 5.6.6 to 5.6.7

kfriedberger:


kfriedberger:
Current status:
- Linux: v5.6.7 seems to fine. Performance evaluation via CPAchecker comes soon.
- Windows: there is a library problem ('procedure not found') with our `mathsat5j.dll` on some (not all) Windows systems.

kfriedberger:
It looks like the issue with the Windows-DLL is a bug in MinGW 9.3 from 2020, i.e. resulting in an incompatibility with MSVC 19.12 shipped with MathSAT 5.6.7). A more recent version (version 10 from 2022) of MinGW helps.

Lets see if we can update the Docker image for Ubuntu 20.04 with MinGW in version 10, and then repeat the compilation.

kfriedberger:
The situation for compiling MathSAT 5.6.7 appropriately is quite funny:
- For compiling the Linux library, we want to use Ubuntu 20.04 as basis with glibc 2.31. An older version of Ubuntu or glibc would need additional hacks to fully work. A newer version of Ubuntu, like 22.04, brings glibc 2.34 or newer, which is not compatible with running on older systems.
- For compiling the Windows library, we need to use MinGW 10, which is not easily installable and available on Ubuntu 20.04, but is only packaged for newer systems, such as Ubuntu 22.04.

The best solution could be to use two different Docker images for compiling for two distinct target systems, i.e. Linux and Windows. I am still trying to find a more elegant solution.

PhilippWendler:
I fear that compiling the Linux binary on Ubuntu 20.04 will make it unusable on older Linux. Is it really safe to do so? I would guess it is less effort to just compile on the oldest supported platform and to check manually whether it is safe.

kfriedberger:
Hi @PhilippWendler.
Yes, using Ubuntu 20.04 as basis will lead to loosing support for older platforms like Ubuntu 18.04.
However, there is good reason to do so:
- Z3 4.8.17.0 already switched to `glibc_2.29` and `GLIBCXX_3.4.26` which is not available on Ubuntu 18.04, but only on later versions. (Some JUnit tests for JavaSMT fail because of this since some weeks).
- The library for MathSAT 5.6.7 uses `_ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev` which is not available with Ubuntu 18.08, even with installing gcc-9/g++-9 from an additional repository.
- Users with an Ubuntu version that is four years old - even if this is an LTS version - do not benefit from several other improvements in the system, and they can of course continue using any previous version of MathSAT and JavaSMT. The number of users still working with Ubuntu 18.04 might be quite small.

Overall, it seems to be the logical choice to use Ubuntu 20.04 as basis.

PhilippWendler:
:scream: 

---

> Users with an Ubuntu version that is four years old - even if this is an LTS version - do not benefit from several other improvements in the system,

So what? If these other improvements are not relevant for the users, what does this have to do with JavaSMT? This is not a convincing argument in this discussion.

> and they can of course continue using any previous version of MathSAT and JavaSMT.

This is not really true, because most users of JavaSMT will just use it as a transitive dependency and not directly, so they do not get a choice in which JavaSMT version to use (for example users of CPAchecker and other tools built on JavaSMT). There is no way for example a user could say that they want to use MathSAT 5.6.6 together with a current version of CPAchecker because they need some bug fix there.

Development of a tool and development of a library are two different things, and maintainers need to have a different mind set. Many things that tool developers can handle more easily are more complex for library developers, and this is such a case.

---

For Z3, one can at least compile Z3 on your own, if the official binary does not work.

For MathSAT, we could ask the MathSAT developers to provide a binary that does not have this dependency, and in general I would recommend to so. But right now we are coming close to the time where we do not need to care about Ubuntu 18.04 anymore, so we might chicken out this time and just declare Ubuntu 18.04 as unsupported.

But of course we should not just look at only Ubuntu. For example, does it work on Debian 10? I think we should definitely aim to keep Debian 10 supported, because otherwise we would force all Debian developers to use only the latest version, which is not even one year old (this would be like requiring Ubuntu 22.04 in the Ubuntu world).

---

And anyway, of course this needs to be clearly communicated to the users. It seems that the current situation highlights quite a few problems:

There seems to be no documentation about the minimum required versions?! There should be a list with the required versions of libraries like glibc, Java, etc., and for common distributions like Ubuntu and Debian an indication from which release on these requirements are fulfilled. The changelog also needs to inform about this.

And of course, these things should be also be tested and checked by CI. I.e., there should be reliable tests that use these platforms and make sure everything works on them.

Note that I have written down and suggested all this already four years ago: #125!

---

> (Some JUnit tests for JavaSMT fail because of this since some weeks).

Really? Why wasn't this detected and worked on already? Does this mean that other bugs and test failures are possibly undetected due to this? I would say that a problem that prevents tests from being executed should always be worked on with a high priority, because if one instead continues other development work, this will just create more effort if tests are currently worthless.

kfriedberger:
> scream

You might be exaggerating. (Maybe let's talk about this topic in detail offline :beers: :wine_glass: )
I understand your point and support most of it.

> > they can of course continue using any previous version of MathSAT and JavaSMT.
> 
> This is not really true, because most users of JavaSMT will just use it as a transitive dependency and not directly, so they do not get a choice in which JavaSMT version to use (for example users of CPAchecker and other tools built on JavaSMT). There is no way for example a user could say that they want to use MathSAT 5.6.6 together with a current version of CPAchecker because they need some bug fix there.

Please also consider, that the API of MathSAT and also of JavaSMT changes in a few methods with each larger release, so a direct one-to-one replacement of the solver is not possible. A current version of CPAchecker could call more methods from JavaSMT and from the solver than available in an older version of them. 

JavaSMT does not (yet plan to) follow a strict strategy like [semantic versioning](https://semver.org/), but mostly use the minor version number for updates. Of course, we provide hints on interface changes in the changelog.

> Development of a tool and development of a library are two different things, and maintainers need to have a different mind set. Many things that tool developers can handle more easily are more complex for library developers, and this is such a case.
> 
> For Z3, one can at least compile Z3 on your own, if the official binary does not work.

This is also true for JavaSMT.

> For MathSAT, we could ask the MathSAT developers to provide a binary that does not have this dependency, and in general I would recommend to so. But right now we are coming close to the time where we do not need to care about Ubuntu 18.04 anymore, so we might chicken out this time and just declare Ubuntu 18.04 as unsupported.
> 
> But of course we should not just look at only Ubuntu. For example, does it work on Debian 10? I think we should definitely aim to keep Debian 10 supported, because otherwise we would force all Debian developers to use only the latest version, which is not even one year old (this would be like requiring Ubuntu 22.04 in the Ubuntu world).

Debian 10 (oldstable) is nearly the same basis and age as Ubuntu 18.04, and its end-of-life is also coming in 2022 (only the LTS stays until 2024, but there might be no security updates :scream:).
Debian 11 was available since 2019 as unstable and testing version.
However, asking Alberto is a good idea. Maybe he can provide a MathSAT library depending on an older version of glibc.

> And anyway, of course this needs to be clearly communicated to the users. It seems that the current situation highlights quite a few problems:
> 
> There seems to be no documentation about the minimum required versions?! There should be a list with the required versions of libraries like glibc, Java, etc., and for common distributions like Ubuntu and Debian an indication from which release on these requirements are fulfilled. The changelog also needs to inform about this.

A list of important library versions per operating system is a good idea.
Since JavaSMT 3.0.0 we provide all releases and binary solvers for Ubuntu 18.04, i.e. depending on glibc 2.27. Java is required in version 11 or newer.

> And of course, these things should be also be tested and checked by CI. I.e., there should be reliable tests that use these platforms and make sure everything works on them.

We have a well-written CI test based on Ubuntu 18.04 that works quite well for all binary solvers, except for Z3 since about 3 weeks.

> Note that I have written down and suggested all this already four years ago: #125!

Great. Someone should start working on that topic. :smiley: 
Seriously, we already use a Docker image based on Ubuntu 18.04 (glibc 2.27) that provides the required compiler and build dependencies for all our binary solvers. We could take a look at some Docker image based on Debian 10 (glibc 2.28) or even manylinux_X_Y (glibc X.Y).

> > (Some JUnit tests for JavaSMT fail because of this since some weeks).
> 
> Really? Why wasn't this detected and worked on already? Does this mean that other bugs and test failures are possibly undetected due to this? I would say that a problem that prevents tests from being executed should always be worked on with a high priority, because if one instead continues other development work, this will just create more effort if tests are currently worthless.

This issue was detected immediatelly, but I did not recognize this as a major problem and did not report it, because of the growing idea to remove support for Ubuntu 18.04 with some upcoming release. No other unit test was affected, because the ones on Ubuntu 20.04 run well.

kfriedberger:
Answer from Alberto: He will try to provide a library depending on only the older version of GLIBC.

PhilippWendler:
MathSAT 5.6.8 was just released and fixes a bug with wrong answers that creates trouble for CPAchecker. Could we please pick up this version and get a release?

kfriedberger (commented):
We skip MathSAT 5.6.7 and directly go to 5.6.8..

This PR changes the API for casting a FP formula to a different formula type.

PR #283, opened 2022-10-03
Fix bug in regex intersection and difference

Swalkyn:
While working with the theory of strings, I saw unexpected behavior when using regex intersection or difference. It turns out that union was used for those operations instead of intersection. I made the correction and added 3 unit tests. I also made a fix to `testRegexAllCharUnicode()`, which was only passing because of this bug.


kfriedberger:
Thanks for finding and fixing this bug.

PR #302, opened 2023-04-25
Add opensmt2

daniel-raffler:
### - Work in progress - ###
- Almost all functionality is supported now
- New interface to select logic is still preliminary
- Some work is still needed on the swig headers (f.ex handling exceptions)


build/ivysettings.xml
```diff
@@ -20,11 +20,18 @@ SPDX-License-Identifier: Apache-2.0
...
-            <artifact pattern="${ivy.repo.url}/[organisation]/[module]/[artifact]-[revision](-[classifier]).[ext]" />
-        </url>
-
+        <chain name="Sosy-Lab">
+            <url name="Sosy-Labs" descriptor="required">
+                <ivy pattern="${ivy.repo.url}/[organisation]/[module]/ivy-[revision].xml" />
+                <artifact pattern="${ivy.repo.url}/[organisation]/[module]/[artifact]-[revision](-[classifier]).[ext]" />
+            </url>
+            
+            <filesystem name="Sosy-Hacks" descriptor="required">
```
kfriedberger, line 29: You could also provide a file `build.properties` with the following content:
```
# ivy.repo.url=https://www.sosy-lab.org/ivy
ivy.repo.url=file:/home/user/workspace/java-smt/repository
```

Then Ant/Ivy would use that directory as repository.

daniel-raffler: Hello Karlheinz,
unfortunately the remote repository is still needed to get the java dependencies for the build.

I can revert the first commit from my branch once the code is ready to be merged. This should get everything back in line with the main branch

build/build-junit.xml
```diff
@@ -57,10 +57,12 @@ SPDX-License-Identifier: Apache-2.0
...
             <sysproperty key="java.awt.headless" value="true" />
             <batchtest fork="true" todir="${junit.dir}">
-                <fileset dir="${class.dir}">
+              <fileset dir="${class.dir}">
                     <include name="**/*Test.*"/>
                     <exclude name="**/*$*Test.*"/>
-                </fileset>
+                    <exclude name="**/*NativeApiTest.*"/>
+                    <exclude name="**/*NativeAPITest.*"/>
+              </fileset>
```
kfriedberger, line 65: unneeded change. please revert.

build/build-junit.xml
```diff
@@ -89,6 +91,8 @@ SPDX-License-Identifier: Apache-2.0
                     <fileset dir="${class.dir}">
                         <include name="**/*Test.*"/>
                         <exclude name="**/*$*Test.*"/>
+                        <exclude name="**/*NativeApiTest.*"/>
+                        <exclude name="**/*NativeAPITest.*"/>
                     </fileset>
```
kfriedberger, line 96: unneeded change. please revert.

solvers_maven_conf/maven_opensmt_pom_template.xml
```diff
@@ -0,0 +1,62 @@
...
+
+  <scm>
+      <url>https://github.com/sosy-lab/java-smt/</url>
+      <connection>scm:git:git://github.com/sosy-lab/java-smt.git</connection>
+      <developerConnection>scm:git:git@github.com:sosy-lab/java-smt.git</developerConnection>
+  </scm>
+
+  <developers>
+      <developer>
+          <name>Karlheinz Friedberger</name>
```
kfriedberger, line 43: well, please do not add people's names to files without asking them :-)

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
```diff
@@ -0,0 +1,278 @@
...
+    Preconditions.checkState(!closed);
+    Preconditions.checkState(!changedSinceLastSatQuery);
+    return super.getModelAssignments();
+  }
+
+  /** Make sure that the assertions only use features supported by the selected sublogic. */
+  private void checkCompatability() throws SolverException {
+
+    Logic osmtLogic = creator.getEnv();
+    PTRef asserted = osmtLogic.mkAnd(new VectorPTRef(getAssertedFormulas()));
```
kfriedberger, line 160: The conjunction here is dangerous, because a solver could internally simplify the assertion and remove symbols from the result.

solvers_maven_conf/maven_opensmt_pom_template.xml
kfriedberger: just keep it as it is now. I will take responsibility for questions in the future.

src/org/sosy_lab/java_smt/SolverContextFactory.java
```diff
@@ -58,6 +60,28 @@ public enum Solvers {
     YICES2
   }
 
+  public enum Logics {
```
kfriedberger, line 63: The logics are only required for OpenSMT. Do we really need them to be in the context-factory? Would a ConfigurationOption in the OpenSMT-context be sufficient?

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
```diff
@@ -0,0 +1,281 @@
...
+
+    // BUGFIX: We need to store the SMTConfig reference to make sure the underlying C++ object does
+    // not get garbage collected
+    osmtConfig = pConfig;
+    osmtSolver = new MainSolver(creator.getEnv(), pConfig, "JavaSmt");
+
+    assertionStack.push(new ArrayList<>()); // create initial level
+  }
+
+  protected static SMTConfig getConfigInstance(
```
kfriedberger, line 70: better naming: "getConfig" instead of "getConfigInstance"

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtEvaluator.java
```diff
@@ -0,0 +1,35 @@
...
+import opensmt.Logic;
+import opensmt.MainSolver;
+import opensmt.Model;
+import opensmt.PTRef;
+import opensmt.SRef;
+import org.sosy_lab.java_smt.basicimpl.AbstractEvaluator;
+
+public class OpenSmtEvaluator extends AbstractEvaluator<PTRef, SRef, Logic> {
+
+  private final MainSolver osmtSolver;
```
kfriedberger, line 21: Is the osmtSolver unused?

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtFormulaCreator.java
```diff
@@ -0,0 +1,462 @@
...
+    List<Logics> supported =
+        Arrays.asList(
+            Logics.QF_AX,
+            Logics.QF_ALIA,
+            Logics.QF_ALRA,
+            Logics.QF_AUFLIA,
+            Logics.QF_AUFLRA,
+            Logics.ALL);
+
+    return supported.contains(logicToUse);
```
kfriedberger, line 100: Linear runtime from List. Set would be constant runtime.
Here ,this is negligible due to size, but keep in mind for other situations.

src/org/sosy_lab/java_smt/test/ModelEvaluationTest.java
```diff
@@ -203,6 +203,7 @@ public void testGetRationalsEvaluation() throws SolverException, InterruptedExce
 
   @Test
   public void testGetBooleansEvaluation() throws SolverException, InterruptedException {
+    // NOTE: Default value for boolean in OpenSMT is 'true'. We patched it to resolve the issue.
```
kfriedberger, line 206: JavaSMT should not change default values of SMT solvers just to its own JUnit test. This might cause overhead in the future when debugging. 
The better solution is to use the default values as given by the SMT solver, and adjust the JUnit test.

lib/native/source/opensmt/api.patch
```diff
@@ -0,0 +1,537 @@
...
+diff --git a/src/logics/FunctionTools.h b/src/logics/FunctionTools.h
+index fc9dbfb7..943a466d 100644
+--- a/src/logics/FunctionTools.h
++++ b/src/logics/FunctionTools.h
+@@ -33,7 +33,7 @@ class TemplateFunction {
+     FunctionSignature signature;
+     PTRef tr_body;
+ 
+-    inline static std::string_view constexpr template_arg_prefix = ".arg";
++    inline static constexpr std::string_view template_arg_prefix = ".arg";
```
kfriedberger, line 220: Is there a reason for changing the order of modifiers in OpenSMT API? This makes the patch quite complex. We should prefer to only patch a minimum number of lines and - if possible - let the OpenSMT team fix other issues in their code.

lib/native/source/opensmt/api.patch
```diff
@@ -0,0 +1,537 @@
...
+-const char* Logic::tk_val_bool_default = "true";
++const char* Logic::tk_val_bool_default = "false";
+ 
+ const char* Logic::tk_true     = "true";
+ const char* Logic::tk_false    = "false";
+@@ -130,12 +131,13 @@ bool Logic::hasQuotableChars(std::string const & name) const
+         0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+         0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+         0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
++        0, 0, 0, 0, 0, 0
```
kfriedberger, line 249: Why do we need to patch a number of zeroes?

lib/native/source/opensmt/api.patch
```diff
@@ -0,0 +1,537 @@
...
++    for (unsigned char c : name) {
++        if (quotable[c])
+             return true;
+     }
+     return false;
+@@ -150,7 +152,7 @@ std::string Logic::disambiguateName(std::string const & protectedName, SRef sort
+         return protectedName;
+     }
+ 
+-    auto isQuoted = [](std::string const &s) { return s.size() > 3 and *s.begin() == '|' and *(s.end()-1) == '|'; };
```
kfriedberger, line 265: Why do we need to patch the number of chars for quoting?

lib/native/source/opensmt/api.patch
```diff
@@ -0,0 +1,537 @@
...
+diff --git a/src/sorts/SSort.h b/src/sorts/SSort.h
+index 990da3b0..14922397 100644
+--- a/src/sorts/SSort.h
++++ b/src/sorts/SSort.h
+@@ -40,7 +40,7 @@ struct SSymRef {
+     inline friend bool operator!= (SSymRef a1, SSymRef a2) {return a1.x != a2.x; }
+ };
+ 
+-const static struct SSymRef SSymRef_Undef = {INT32_MAX};
++static const struct SSymRef SSymRef_Undef = {INT32_MAX};
```
kfriedberger, line 433: Why do we change the ordering of modifiers here?

lib/native/source/opensmt/api.patch
```diff
@@ -0,0 +1,537 @@
...
+     SymId id = symbols.size();
+diff --git a/src/symbols/Symbol.h b/src/symbols/Symbol.h
+index e795bd56..4b6f6018 100644
+--- a/src/symbols/Symbol.h
++++ b/src/symbols/Symbol.h
+@@ -48,20 +48,20 @@ struct SymbolConfig {
+ };
+ 
+ namespace SymConf {
+-    constexpr auto Default = SymbolConfig{false, false, false, SymbolProperty::None};
```
kfriedberger, line 507: Why do we need to remove "auto" as modifier? Does SWIG not support "auto"?

lib/native/source/opensmt/opensmt.i
```diff
@@ -0,0 +1,1419 @@
+// This file is part of JavaSMT,
+// an API wrapper for a collection of SMT solvers:
+// https://github.com/sosy-lab/java-smt
+//
+// SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+%module OsmtNative
```
kfriedberger, line 9: Where does this file come from? Was it auto-generated? Over 1000 lines of code without a clear desciption is quite hard to maintain in the future.

lib/native/source/opensmt/api.patch
kfriedberger: Did they just fill an array with 256 chars with only 250 instanzes of 1 or 0? Oh no ;-)

lib/native/source/opensmt/api.patch
kfriedberger: This seems to happen just to avoid compilation problems later in the prozess, because "static" should come first. Oh no ;-)

lib/native/source/opensmt/api.patch
blishko: We can definitely fix these problems in OpenSMT itself. Could you open issues for these findings?

src/org/sosy_lab/java_smt/SolverContextFactory.java
daniel-raffler: We thought that other solvers may also benefit from a way to manually select a specific logic. However, I can look into moving it to the OpenSMT context later today.

lib/native/source/opensmt/api.patch
daniel-raffler: Hello Martin,
I'll split the patch up and then open a pull request for OpenSMT so you can pick and choose the changes you want to add to OpenSMT.

lib/native/source/opensmt/api.patch
daniel-raffler: Yes, the array is too short :)
I ran into this while debugging one of the test, and while it's not the actual bug, this might also have caused problems for us since JavaSMT does allow all 256 ASCII codes.

lib/native/source/opensmt/api.patch
daniel-raffler: I think this is another bug, but I'll have to check with Martin.

lib/native/source/opensmt/api.patch
daniel-raffler: Yes, SWIG can't parse the header and simply crashes if the order of the modifies is not changed.

lib/native/source/opensmt/api.patch
daniel-raffler: I think this was causing parsing issues with SWIG as well, but I'll have to go back and check.

lib/native/source/opensmt/opensmt.i
daniel-raffler: It's semi-autogenerated :). I ran doxygen an the OpenSMT headers, then copied the method definitions from there and ran them through sed to get the 'ignore' statements. The whole process was quite messy and it required some extra editing. It might be possible to shrink this file significantly by not listing all the methods, however, SWIG will then generate stub classes for a lot of the other internal classes of OpenSMT, which is also messy. I'll see if I can find a better way, but it may take some time.

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtEvaluator.java
kfriedberger: or is the reference required to avoid some garbage collection? then please document this.

lib/native/source/opensmt/api.patch
kfriedberger: I took a look at the OpenSMT code. This specific method seems to be quite weird. It seems to suffer from several issues:
- the array is too short (fixed by adding more zeros)
- the array-access is not guaranteed to be in range [0, 256] (fixed by casting to unsigned char)
- the large 0/1-block is not readable and might require some documentation.
- C++ might not store strings as char-arrays (8bit), but as int32-arrays, such that additional prefix-bits from UTF-8 appear in the comparison (fixed by casting to unsigned char).
- Does C/C++ even guarantee that a char is always 8bit wide? It is the default, but the standard might allow arbitrary sizes for characters and bytes. (Just a meta question about machine-independent code style :-) )

The benefit is that it might be quite fast, if checking millions of characters.

One could replace that whole code with something much simpler:
```
if (name.front() == '|' and name.back() == '|') return false; // Already quoted
for (auto c : name) {
   switch (c) { // check for quotable chars
       case ' ':
       case '"':
       case '#':
       case '$':
       case '\'':
       case ':':
       case ';':
       case '[':
       case '\\':
       case ']':
       case '{':
       case '}':
           return true;
   }
}
return false;
```

This might avoid several low-level C/C++-based issues, and it is surely more readable than the previous code.

@daniel-raffler Could you file a pull-request for that change in OpenSMT?

lib/native/source/opensmt/api.patch
kfriedberger: Is the terminal \0 counted in the string-size in C++? Then 3 would be valid for "||\0".

lib/native/source/opensmt/opensmt.i
kfriedberger: I read through the SWIG-file and it looks quite easy to understand.
I like the idea of only providing a subset of the API via SWIG, i.e., only the required parts.
Keep it as is.

kfriedberger:
Can this PR be set to "ready for review"? I would like to begin an 'official' review and might push some minor changes onto this PR. Overall it already looks quite good.

lib/native/source/opensmt/api.patch
blishko: No, it is not counted in the size of `std::string`. I'll see if I can find out why it is like this, but I agree it should probably be `2`.

lib/native/source/opensmt/api.patch
blishko: `const static` still seems to be legal C++, but it looks like `static const` is much more preferred. We can fix this.

kfriedberger:
I copied over the PR to a branch of JavaSMT.
Further changes from `daniel-raffler:add_opensmt2` will be taken over.
I will open another PR for merging the branch to `master`.

lib/native/source/opensmt/api.patch
daniel-raffler: I've now opened a pull request for OpenSMT: https://github.com/usi-verification-and-security/opensmt/pull/646

Issue #310, opened 2023-06-04
CVC5: parallel prover instances cause issues (found in CPAchecker benchmark)


kfriedberger:
Our expectation in CVC5 being able to provide multiple provers that can be used interleaved (or even parallel) based on a common basis (shared types and formulas) was not fulfilled.
There are multiple errors when applying CVC5 that way.

Examples from CPAchecker when applying kInduction (two provers applied in parallel)
LOG:
```
scripts/cpa.sh -heap 13000M -noout -benchmark -stack 50M -setprop cpa.predicate.memoryAllocationsAlwaysSucceed=true -setprop counterexample.export.allowImpreciseCounterexamples=true -setprop cpa.predicate.encodeFloatAs=INTEGER -kInduction -setprop solver.solver=CVC5 -setprop cpa.predicate.encodeBitvectorAs=BITVECTOR -timelimit 60s -stats -spec ../sv-benchmarks/c/properties/unreach-call.prp -32 ../sv-benchmarks/c/loops/n.c24.i


--------------------------------------------------------------------------------


Running CPAchecker with Java heap of size 13000M.
Running CPAchecker with Java stack of size 50M.
Language C detected and set for analysis (CPAMain.detectFrontendLanguageIfNecessary, INFO)

Using the following resource limits: CPU-time limit of 60s (ResourceLimitChecker.fromConfiguration, INFO)

CPAchecker 2.2.1-svn-43814M / kInduction (OpenJDK 64-Bit Server VM 17.0.7) started (CPAchecker.run, INFO)

Parsing CFA from file(s) "../sv-benchmarks/c/loops/n.c24.i" (CPAchecker.parse, INFO)

Using the following resource limits: CPU-time limit of 60s (Parallel analysis 1:ResourceLimitChecker.fromConfiguration, INFO)

Using unsound approximation of floats with unbounded integers for encoding program semantics. (Parallel analysis 1:PredicateCPA:FormulaManagerView.<init>, WARNING)

Using predicate analysis with CVC5 1.0.5. (Parallel analysis 1:PredicateCPA:PredicateCPA.<init>, INFO)

Using unsound approximation of floats with unbounded integers for encoding program semantics. (Parallel analysis 1:AssumptionStorageCPA:FormulaManagerView.<init>, WARNING)

Using unsound approximation of floats with unbounded integers for encoding program semantics. (Parallel analysis 1:InductionStepCase:PredicateCPA:FormulaManagerView.<init>, WARNING)

[...]

PR #320, opened 2023-07-18
293 cvc5 interpolation

RSoegtrop:
Implements CVC5 Interpolation, following the Definition of Craig-Interpolation.


src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,334 @@
...
+import org.sosy_lab.java_smt.api.InterpolatingProverEnvironment;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
+import org.sosy_lab.java_smt.api.SolverException;
+
+public class CVC5InterpolatingProver extends CVC5AbstractProver<Term>
+    implements InterpolatingProverEnvironment<Term> {
+
+  private final FormulaManager mgr;
+  // private final CVC5FormulaCreator creator;
+  private final Set<Term> assertedFormulaHash = new HashSet<>(); // comparable to SMTInterpols
```
kfriedberger, line 34: Is a plain Set sufficient?

Counterexample:
What happens on the following process: "push(A), push(B), push(A), pop()"? Then "A" is no longer in the set.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  /**
+   * Interpolates a Tuple of Interpolants according to Craig-Interpolation using CVC5-Interpolation.
+   *
+   * @param formulaPair Interpolation Pair
+   * @return Interpolation of the Interpolation Pair following the definition of
+   *     Craig-Interpolation.
+   */
+  private Term getCVC5Interpolation(ArrayList<ArrayList<Collection<Term>>> formulaPair) {
```
kfriedberger, line 168: If a parameter is a simple Pair, then please use two separate arguments for the first and second pair-element.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -234,6 +234,7 @@ private void requireTreeItp() {
     assume()
         .withMessage("Solver does not support tree-interpolation.")
         .that(solver)
+        // .isAnyOf(Solvers.Z3, Solvers.SMTINTERPOL, Solvers.PRINCESS);
```
kfriedberger, line 237: Why this change?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,334 @@
...
+    if (extraAssertions.isEmpty()) {
+      extraAssert = solver.mkTrue();
+    } else {
+      extraAssert = buildConjunctionOfFormulas(extraAssertions);
+    }
+
+    Term phim = buildConjunctionOfFormulasOverArray(assertedInterpols);
+    Term phip = buildConjunctionOfFormulasOverArray(addedInterpols);
+
+    solver.resetAssertions();
```
kfriedberger, line 191: Just asking:
Is this the solver of this ProverEnvironment-instance?
We must delete the entire solver's context to solve interpolation?
So...ah... how do you rebuild the context after interpolation is done?

kfriedberger: Could we use a separate solver-instancee for interpolation and keep the original solver-context unmodified?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,334 @@
...
+    solver.assertFormula(solver.mkTerm(Kind.AND, extraAssert, phim));
+
+    // with Phip negated, CVC5 Interpolation produces an equivalent interpolation to
+    // Craig-Interpolation
+    Term interpolant = solver.getInterpolant(solver.mkTerm(Kind.NOT, phip));
+
+    // restore Assertion state
+    solver.resetAssertions();
+
+    assertedFormulaHash.forEach((n) -> solver.assertFormula(n));
```
kfriedberger, line 201: This should rebuild the solver-context, maybe. 
1) Why is the code between deleting the solver-context and rebuilding it not in a try-finally-block?
2) The rebuild ignores levels from push/pop.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I swapped it with a `Deque` in the newest commit. It should solve it, right?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Done

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
RSoegtrop: I wanted to test out `Tree Interpolation` with CVC5. Forgot to change it back.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Yes, the newest commit uses a separate solver-instance and leaves to original solver-context untouched.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: With the separate solver-instance, rebuilding is not needed anymore.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: No. A one-dimensional data-structure will not help here.
If you really need to keep track of the existing levels, then you might need use something like a Deque<Collection<Term>>, as it is also done in some other ProverEnvironments.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: Just for clarification, let me ask some questions.

Technical part:
- Does CVC5 allow to share formulas (Terms) between distinct Solver instances?  
JavaSMT already has the possibility for several ProverEnvironments for cvC5, so that part should alredy work as expected, without the requirement of translating formulas or synchronizing symbol declarations. Correct?
- Do you need the solver instance only for the interpolation query itself or for the full lifetime of the Proverenvironment? Is there any information kept in the Solver after returning from the interpolation query?
If not, please do not create the Solver-instance on class-level, but only as part of the interpolation-method.
And check whether it can be closed afterwards.

Theoretical part:
- Can you describe/document how the interpolation based on the CVC5-interface should work? 
- Until now, I could not see any check on symbol usage in the interpolant. Does a CVC5-provided interpolant only contain symbols that are shared between both input formulas? Is this specified in the CVC5-API?

Deep theoretical part (for later documentation): :-)
- Where is the guarantee that a sequence of interpolants satisfies the interpolation criteria of sequntial interpolation?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+  @Override
+  public BooleanFormula getInterpolant(Collection<Term> pFormulasOfA)
+      throws SolverException, InterruptedException {
+    Preconditions.checkState(!closed);
+
+    if (pFormulasOfA.isEmpty()) { // Catch trivial case
+      return mgr.getBooleanFormulaManager().makeBoolean(true);
+    }
+
+    Set<Term> formulasOfA = ImmutableSet.copyOf(pFormulasOfA);
```
baierd, line 93: This copy is not needed.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+            .filter(n -> !formulasOfA.contains(n))
+            .collect(ImmutableSet.toImmutableSet());
+
+    if (formulasOfB.isEmpty()) { // Catch trivial case
+      return mgr.getBooleanFormulaManager().makeBoolean(false);
+    }
+
+    // fit the Input to work with getCVC5Interpolation
+
+    ArrayList<Collection<Term>> formAAsList = new ArrayList<>();
```
baierd, line 107: Please remove ArrayList and use e.g. ImmutableList

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+   *
+   * <p>(1CVC5) <=> (1Craig) due to subst of A with phi- and reflexivity.
+   *
+   * <p>(2CVC5) <=> I -> B = True <=> (not I) \/ B = True <=> not (I /\ (not B)) = True <=> (2Craig)
+   * due to subst of (not B) with phi+ and reflexivity.
+   *
+   * <p>Hence, CVC5's Interpolation produces an equivalent Interpolation to Craig Interpolation, if
+   * B is negated during CVC5 Interpolation.
+   *
+   * @param assertedInterpols Asserted Formulas
```
baierd, line 193: Currently asserted or to be asserted formulas?
Why not name them after the formulas you described above?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+
+    // Respect Asserted Formulas not in the Interpolation Pairs
+
+    ArrayList<Collection<Term>> combinedInterpols = new ArrayList<>();
+    combinedInterpols.addAll(assertedInterpols);
+    combinedInterpols.addAll(addedInterpols);
+
+    // checks, that no Assertions in the Assertion stack are left out. Can happen in Tree
+    // Interpolation
+    Collection<Term> extraAssertions = getAssertedTermsNotInCollection(combinedInterpols);
```
baierd, line 209: Why can this happen?
Please create an example and write a test-case for it.
Also, this looks like an assertion/precondition if this is not in general possible.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+    Term extraAssert = new Term();
+
+    Solver interpolationSolver = new Solver(); // Uses a separate Solver instance to leave to
+    // original solver-context unmodified
```
baierd, line 213: Runaway comment.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+   * @return concatenated Formulas with AND as CVC5 Term
+   */
+  private Term buildConjunctionOfFormulasOverArray(
+      ArrayList<Collection<Term>> formula, Solver usingSolver) {
+    Collection<Term> currColTerm = formula.get(0);
+    formula.remove(0);
+    if (formula.size() == 0) {
+      return buildConjunctionOfFormulas(currColTerm, usingSolver);
+    }
+    return usingSolver.mkTerm(
```
baierd, line 270: This might buffer-overflow for very large formulas. Just to keep in mind.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+
+    setSolverOptions(seed, solverOptions, interpolationSolver);
+
+    if (extraAssertions.isEmpty()) {
+      extraAssert = interpolationSolver.mkTrue();
+    } else {
+      extraAssert = buildConjunctionOfFormulas(extraAssertions, interpolationSolver);
+    }
+
+    Term A = buildConjunctionOfFormulasOverArray(assertedInterpols, interpolationSolver);
```
baierd, line 223: You could just build `extraAssert` after this line and add `extraAssert` to a with an conjunction.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,368 @@
...
+    Term B = buildConjunctionOfFormulasOverArray(addedInterpols, interpolationSolver);
+
+    interpolationSolver.resetAssertions();
```
baierd, line 226: Why do you need to reset the assertions of a fresh solver instance?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+
+  /**
+   * Turns a collection of Formulas to a Single Conjunction of the Formulas e.g.: [A,B,C] ->
+   * A/\B/\C.
+   *
+   * @param formulas collection of formulas
+   * @param usingSolver the CVC5 Solver Instance to use
+   * @return concatenated Formulas with AND as CVC5 Term
+   */
+  private Term buildConjunctionOfFormulas(Collection<Term> formulas, Solver usingSolver) {
```
kfriedberger, line 280: This method looks like a very expensive version of 
```
switch(formulas.size()) {
  case 0: return solver.mkBoolean(true);
  case 1: return Iterables.getOnlyElement(formulas);
  default: return solver.mkTerm(Kind.AND, formulas.toArray());
}
```

If possible, take a look into the implementation of CVC5BooleanFormulaManager:makeAnd(Collection).

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+  /**
+   * Returns Asserted Formulas (from the formula Stack), not present in the given collection of
+   * Terms.
+   *
+   * @param pCombinedInterpols asserted Formulas to invert the selection
+   * @return asserted Formulas not in collTerms, but in the formula Stack
+   */
+  private Collection<Term> getAssertedTermsNotInCollection(
+      ImmutableList<Collection<Term>> pCombinedInterpols) {
+    ImmutableList<Term> retTerms =
```
kfriedberger, line 240: Please consider runtime when using such data-structures.
A List requires a full iteration when searching for an element.
The code  few lines later checks for "contains". Very bad choice.

Tipp: You might want to use an ImmutableSet :-)

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+   * Turns a List of Collections of Formulas to a Single Conjunction of the Formulas e.g.:
+   * [[A,B],[C]] -> A/\B/\C.
+   *
+   * @param formula List of Collections of formulas
+   * @param usingSolver the CVC5 Solver Instance to use
+   * @return concatenated Formulas with AND as CVC5 Term
+   */
+  private Term buildConjunctionOfFormulasOverList(
+      ImmutableList<Collection<Term>> formula, Solver usingSolver) {
+    Collection<Term> eachCollectionOfTermsConcat =
```
kfriedberger, line 263: Instead of conjuncting several conjunctions from several lists, it would be much easier to only perform one conjunction of a longer (flattened) list.

The code might look like:
```
return buildConjunctionOfFormulas(FluentIterable.concat(formula).toSet(), usingSolver);
```

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+    solvertoSet.setOption("full-saturate-quant", "true");
+  }
+
+  @Override
+  public Term addConstraint(BooleanFormula pConstraint) throws InterruptedException {
+    Preconditions.checkState(!closed);
+    Term t = creator.extractInfo(pConstraint);
+
+    super.addConstraint(pConstraint);
+    return t; // t is not returned in the Abstract Class
```
kfriedberger, line 82: Are you sure to return the internal reference?

Please consider the following cases:
- What happens when adding the same constraint several times?
- What is Craig-interpolant of "{A} / {A}" with "A" being unsatisfiable?
- What is Craig-interpolant of "{A, B} / {A}" with "B" being unsatisfiable?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+        ImmutableList.copyOf(pPartitionedFormulas.subList(1, pPartitionedFormulas.size()));
+    // current generated Interpolation Tuple
+    ImmutableList<ImmutableList<Collection<Term>>> betweenRes = ImmutableList.of(currA, currB);
+    ImmutableList<ImmutableList<ImmutableList<Collection<Term>>>> result =
+        ImmutableList.of(betweenRes);
+    // iterate through Tree structure
+    for (int i = 1; i < pStartOfSubTree.length - 1; i++) {
+      // if the leave does not change, continue like Sequential Interpolation
+      if (pStartOfSubTree[i] == pStartOfSubTree[i - 1]) {
+        currA =
```
kfriedberger, line 323: Just because there are immutable data-structures and there are streams, this does not enforce you to only use that kind of data-structures. In this method, that kind of logic makes it really hard to understand and follow the program. It might be much easier to just use a plain Lists or Sets and add elements to them.

I still have no clue what this method is doing.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+      result.add(creator.encapsulateBoolean(itp));
+    }
+    assert result.size() == pStartOfSubTree.length - 1;
+    return result;
+  }
+
+  /**
+   * Interpolates a Tuple of Interpolants according to Craig-Interpolation using CVC5-Interpolation.
+   *
+   * <p>CVC5's getInterpolant: There is a Model, such that the Interpolant I, with A -> I = True
```
kfriedberger, line 171: I am still missing the third constraint on interpolants: which symbols appear in the interpolant?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+   * <p>Craig Interpolation: There is a Model, such that the Interpolant psi, with phi- -> psi =
+   * True (1Craig), not(psi /\ phi+) = True (2Craig).
+   *
+   * <p>With A/phi- current set of assertions and B/phi+ Formulas to interpolate.
+   *
+   * <p>CVC5 -> Craig Interpolation:
+   *
+   * <p>(1CVC5) <=> (1Craig) due to subst of A with phi- and reflexivity.
+   *
+   * <p>(2CVC5) <=> I -> B = True <=> (not I) \/ B = True <=> not (I /\ (not B)) = True <=> (2Craig)
```
kfriedberger, line 183: Are you sure, you can convert SAT to UNSAT in that case? I am not yet convinced.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+    mgr = pMgr;
+    solverOptions = pOptions;
+    seed = randomSeed;
+  }
+
+  /**
+   * Sets the same solver Options of the Original Solver to the separate solvertoSet, except for
+   * produce-interpolants which is set here. From CVC5AbstractProver Line 66
+   */
+  private void setSolverOptions(int randomSeed, Set<ProverOptions> pOptions, Solver solvertoSet) {
```
kfriedberger, line 53: Please do not just copy code, but correctly refactor it to be reusable.

In that case:
- Add a new parameter "pSolver" to the super-method and make the method "protected".
- Then simplify this method to only two lines of code:
```
@Override
protected void setSolverOptions(int randomSeed, Set<ProverOptions> pOptions, Solver pSolver) {
  super.setSolverOptions(randomSeed, pOptions, pSolver);
  solvertoSet.setOption("produce-interpolants", "true");
}
```

This is much easier to understand and avoids redundancy.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+            .filter(n -> !pFormulasOfA.contains(n))
+            .collect(ImmutableSet.toImmutableSet());
+
+    if (formulasOfB.isEmpty()) { // Catch trivial case
+      return mgr.getBooleanFormulaManager().makeBoolean(false);
+    }
+
+    // fit the Input to work with getCVC5Interpolation
+
+    ImmutableList<Collection<Term>> formAAsList = ImmutableList.of(pFormulasOfA);
```
kfriedberger, line 107: Is there a reason for having "lists" instead of "sets"?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,370 @@
...
+  public List<BooleanFormula> getSeqInterpolants(
+      List<? extends Collection<Term>> partitionedFormulas)
+      throws SolverException, InterruptedException {
+    Preconditions.checkArgument(
+        !partitionedFormulas.isEmpty(), "at least one partition should be available.");
+
+    List<BooleanFormula> itps = Stream.of(new BooleanFormula[] {}).collect(Collectors.toList());
+    for (int i = 1; i < partitionedFormulas.size(); i++) {
+      itps.add(
+          getInterpolant(
```
kfriedberger, line 125: I am still missing a proof that this interpolation technique delivers sequential interpolants. 
The interpolation strategy does not enforce a consistent internal state (e.g., internal proof) and seems to even build up the solver-stack several times.

If you are already pushing and popping on a solver-stack, please reuse the first sequential interpolants for further computation. This makes the later computation potentially faster, because the existing interpolants reduce workload.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Good Point. Thanks.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Still have to learn about Java "types" ^^ I mostly replaced Immutable Lists with Immutable Sets.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Good Point.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: ArrayLists, which were used in the beginning, led to Refaster complaining -> replace everything with ImmutableLists (I used the first found method for replacing ArrayLists operators with methods applicable for ImmutableList, mainly some kind of Stream, hence the unreadability). The order matters and i need to copy lists here and there, so ImmutableLists seemed like an alternative. 
This method tries to bring Tree Interpolation to Solvers, with no support for it. It returns a List of Interpolation Pairs (dependent on the given partitionedFormulas and startOfSubTree), following Tree Interpolation. In the following steps, a simple Interpolation Call is sufficient.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: CVC5s documentation does not mention which symbols appear in the interpolant. Is there a elegant way to compare appearing symbols between formulas/terms?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Which part seems problematic? This one: 
```
not (I /\ (not B)) = True <=> I /\ (not B) = False <=> (2Craig)
```
?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I see how it is done now. Thanks.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: No :) its replaced.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I took the implementation of `Mathsat5InterpolationProver`. The only comment:
```
Mathsat5InterpolatingProver line 136: the fallback to a loop is sound and returns an inductive sequence of interpolants
```
it sounds sketch but i went with it. 
Could you elaborate on how the interpolation pairs would look, if the already calculated sequential interpolants would be used for further interpolations? 
Lets say we have following partitionedFormulas `[A,B,C,D,E,F]`
1. `itp0 = getInterpolation([A],[B,C,D,E,F])`
2. `itp1 = getInterpolation([itp0],[C,D,E,F])`
3. ...
Something like this?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: Technical part: 
 - Sharing formulas between different solver instances works.
 - I changed it to exist only during the interpolation-method. One thing would be kept in the solver. If the interpolation call was successful, you could call `getInterpolantNext` directly after to produce a different but also valid interpolant. In the case the assertions do not change, i could "redirect" a second interpolation call to this method. Would this be wished behavior? 
 
The rest is discussed below.

kfriedberger: There is currently no requirement for computing more than one interpolant (or one set of sequential/tree interpolants) per SAT-check. So it can be kept simple and without any state-management.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: Is each pair that is returned here defined as the "subtree" and the "rest of the whole tree", based on a single cut-point that is the interpolation position?

If this is the case, then the interpolation criteria might not be fulfilled.
The reasoning is quite similar to the comment lately regarding sequential interpolation with one proof vs several proofs. 

There is a example documented in CPAchecker, that was a problem for another tree-interpolation problem:
https://gitlab.com/sosy-lab/software/cpachecker/-/blob/trunk/src/org/sosy_lab/cpachecker/util/predicates/interpolation/strategy/WellScopedInterpolation.java
You might give it a try and produce tree interpolants for it.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: The FormulaManager can collect variables and UFs from a formula with calling `fmgr.extractVariablesAndUFs(formula)`.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: Your "proof" only transforms the formula and ignores the fact, that the query for the solver changes from "sat" to "unsat". This just feels weird..., I currently have no counterexample, why it should not work. Perhaps it is valid overall. :-)

Maybe discuss with your supervisor or in a SosyLab group meeting for details.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
kfriedberger: You did correctly got the idea for sequential interpolation.

For an SMT solver that uses the same internal proof for all interpolants, it is easy to provide valid sequential interpolants. (It actually is not "easy". We reported some bugs to the solver-developers in the early times, until they got it right... or removed interpolation support from the solver :smile:). 

I found some colorful slides in an archived folder (I am happy to use them again, after 8 years :smile: no guarantee for correctness! ):

The colors denote the first (GREEN) and second (RED) part of the interpolation query, and the interpolant (BLUE). 

image

image
image
image

image
image
image
image
image

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  @Override
+  public List<BooleanFormula> getSeqInterpolants(
+      List<? extends Collection<Term>> partitionedFormulas)
+      throws SolverException, InterruptedException {
+    Preconditions.checkArgument(
+        !partitionedFormulas.isEmpty(), "at least one partition should be available.");
+
+    List<BooleanFormula> itps = Stream.of(new BooleanFormula[] {}).collect(Collectors.toList());
```
kfriedberger, line 106: hm, what about a plain and simple, easy to read, nicely understandable `new ArrayList<>();` instead of that `thing`?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,327 @@
...
+   */
+  private Term getCVC5Interpolation(
+      ImmutableList<Collection<Term>> pFormAAsList, ImmutableList<Collection<Term>> pFormBAsList) {
+
+    // Uses a separate Solver instance to leave the original solver-context unmodified
+    Solver interpolationSolver = new Solver();
+
+    setSolverOptions(seed, solverOptions, interpolationSolver);
+
+    interpolationSolver.resetAssertions();
```
kfriedberger, line 199: Why do we need to reset a fresh solver instance?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
```diff
@@ -0,0 +1,327 @@
...
+
+    Term interpolantA = buildConjunctionOfFormulasOverList(pFormAAsList, interpolationSolver);
+
+    Term interpolantB = buildConjunctionOfFormulasOverList(pFormBAsList, interpolationSolver);
+
+    interpolationSolver.assertFormula(interpolantA);
+
+    Term interpolant =
+        interpolationSolver.getInterpolant(interpolationSolver.mkTerm(Kind.NOT, interpolantB));
```
kfriedberger, line 208: Just for clarification: The interpolant here is created by calling the interpolationSolver, which is closed a few lines later. It the interpolant still valid after closing the solver?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I dropped the idea of implementing `treeInterpolation` for CVC5. Maybe doing it properly for generally any solver would be wiser. Looking at the slides of yours below, it does not look to complicated, something I would be happy to try out.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I added a small check after each interpolation, but only checking for the maps keys. Is it sufficient?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I did not meet with my supervisor yet, but I want to discuss: 
 - `1CVC5 <=> 1Craig`: both formulas need to hold/be valid, so there should be no problem.
 - `2CVC5 <=> 2Craig`: is the swap from valid to "unsat" not represented by `...=True -> ...=False`? In CVC5 it is always valid (for me that means equals `True`, different to "sat") in Craig it is always "unsat" (the formula equals `False`). So there happens a swap by just transforming formulas, or am I mixing things up?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I implemented it the way you suggested. Surprisingly :) the tests do not fail anymore. Thanks a lot! With this approach, no further prove is needed right?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I was under the impression that ArrayLists were not allowed (Refaster just screamed at me in the beginning). I swapped it back to `ArrayList`.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: We do not. Forgot to remove it.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I can only assume that it works. Sharing formulas between different solver instances works. sequentialInterpolation produces correct Interpolations using those interpolants. 
Is there a way to guarantee it?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: I added a precondition to check that CVC5 satisfies the definition of Craig-Interpolation. It held for every test case.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -987,6 +987,43 @@ public <T> void bigSeqInterpolationTest() throws InterruptedException, SolverExc
     }
   }
 
+  @Test
+  @Ignore
+  public <T> void testTrivialInterpolation() throws InterruptedException, SolverException {
```
baierd, line 992: What is the goal of this test?
You are testing the same behavior twice here.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
RSoegtrop: After a small fix in the test, every solver agrees on False and True.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
RSoegtrop: I fixed the test to validate discussed behavior. 
https://github.com/sosy-lab/java-smt/pull/320#discussion_r1307348386

RSoegtrop:
I needed to disable following Tests because CVC5s `getInterpolant` did not terminate:
 - InterpolatingProverTest
    - simpleInterpolation
    - bigSeqInterpolationTest

 - SolverFormulaWithAssumptionsTest: 
    - basicAssumptionsTest

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5InterpolatingProver.java
baierd: As far as we are informed by the CVC5 devs we can share formulas in between solver instances freely. But there seems to be some problem, see #310 . I would however accept that the intended behavior is that the formulas are sharable.

baierd:
So there is 1 test failing iff the internal assertions of getInterpolant() for CVC5 are active. To be concrete, its only this line `    assert solverP.checkSat().isSat() : "Interpolant does not follow Craig Interpolation";` thats problematic. If you comment out this line, the test runs fine.
My suggestion is to disable this line for now, extract the problem and send it to the CVC5 devs.

kfriedberger (approved):
From my side, the code looks valid, under the assumption that CVC5 provides valid interpolants.
PR approved!

Future work:
- Some unit tests should be analyzed in detail whether CVC5 can be improved. 
- Testing of interpolation on larger SMT queries should be done, e.g., within CPAchecker with predicate analysis or impact algorithm.

kfriedberger:
I published a preliminary development-version of JavaSMT including PR (CVC5-interpolation) into Ivy. It is available with version `4.0.3-59-g6a56e821`. For testing, it can be added as dependency into CPAchecker (simply change the version number in CPAchecker's `lib/ivy.xml` and build CPAchecker).

A first look at benchmark results from SV-COMP shows some solved files with CVC5 and predicate analysis, but still room for improvement. The JavaSMT-bindings from this PR seem to work well.

kfriedberger:
I did a quick benchmark on predicate analysis with CPAchecker, with integer-based formula-encoding (no bitvectors!).

**Result are here:**
- all files: https://www.cip.ifi.lmu.de/~friedber/CPAchecker/cvc5interpolation/
- main table: https://www.cip.ifi.lmu.de/~friedber/CPAchecker/cvc5interpolation/results.2023-09-03_13-33-23.table.html

The interesting thing might be the quantile plot:
image

**In short:**
CVC5 provides some useful interpolants, but fails on several tasks, such that there are less false results (plot-line starts not as far left as other solvers) and there are less solved tasks (plot-line does not reach as far to the right as other solvers).

There are several cases where the analysis reports a repeated counterexample path when using CVC5. This might be caused by invalid interpolants and needs to be checked in detail.
For example:
- file: bitvector/byte_add-1.yml
- command line: `scripts/cpa.sh -heap 13000M -noout -benchmark -stack 50M -setprop cpa.predicate.memoryAllocationsAlwaysSucceed=true -setprop counterexample.export.allowImpreciseCounterexamples=true -predicateAnalysis -setprop cpa.predicate.encodeBitvectorAs=INTEGER -setprop solver.solver=CVC5 -timelimit 900s -stats -spec ../sv-benchmarks/c/properties/unreach-call.prp -32 ../sv-benchmarks/c/bitvector/byte_add-1.i
`
- error in the log: `Error: Refinement failed: Counterexample could not be ruled out and was found again`

kfriedberger:
### Analysis of a problematic task from last benchmark execution, including a comparison of MathSAT5 and CVC5

I used the following example, because the input file is quite short and simple:
```
scripts/cpa.sh -predicateAnalysis \
    -setprop cpa.predicate.encodeBitvectorAs=INTEGER \
    -setprop solver.solver=CVC5 -stats \
    -spec ../sv-benchmarks/c/properties/unreach-call.prp -32 \
    ../sv-benchmarks/c/loop-new/count_by_1.i
```

The `refinement failed` error comes from repeated counterexamples and is caused by invalid interpolant sequences in CPAchecker.

A specific interpolation query was suspicious, coming from three input formulas. The query appears when inroling the loop in the program once. Shorted/simplified formulas given here (see unit test added in 7a1676e):
- 0: `i3=0`
- 1: `i3<1000 & i4=i3+1`
- 2: `i4>1000`

CPAchecker (especially predicate analysis) requires sequential interpolants for the refinement. For the given formulas, the interpolation query should be `[0]/[1]/[2]` (three partitions resulting in two interpolants). The default interpolation strategy of CPAchecker does not query such interpolants (due to historic reasons), but queries two interpolants for `[0]/[1,2]` and `[0,1]/[2]`.

For those two queries, MathSAT5 produces `i3<=0` und `i4<=1`. CVC5 produces `true` und `i4=1`. While the response from MathSAT5 is a valid sequence of interpolant, the response from CVC5 is not. The reason is the new internal proof that is used by CVC5 to generate interpolants.

In general, querying individual interpolants from a sequence of formulas is not the best way to go ([nicely documented 6 years ago](https://github.com/sosy-lab/java-smt/blob/d32310b5b21753786f8e3dca1d2709679f337986/src/org/sosy_lab/java_smt/api/InterpolatingProverEnvironment.java#L32-L34)), and CPAchecker has already an option, such that the problematic query is avoided: `cpa.predicate.refinement.strategy=SEQ` (instead of `SEQ_CPACHECKER`.

A new benchmark show better results and avoids this problem:
- table (CVC5 with interpolation strategy, SEQ_CPACHECKER vs. SEQ): <COMING SOON>
- main table (updated): <COMING SOON> 

image

However CVC5 still has some other issues with interpolation:
- exception about formulas of invalid formula type: `CVC5ApiException: expecting a Boolean subexpression` (debugging in CVC5 required)
- several timeouts on quite simple files (optimization required)

Those problems might not be within the scope of this PR.

kfriedberger:
Is there anything left to do before merging? I would like to merge this PR soon.

PR #324, opened 2023-07-31
322 feature request clone proverenvironment with stack

baierd:
I've added two methods to copy a prover to the solver context with an implementation for Z3 including tests.
Note: it seems like Z3 supports this only for the base level (no pushs to the assertion stack).

### DIFF
--- a/src/org/sosy_lab/java_smt/api/SolverContext.java
+++ b/src/org/sosy_lab/java_smt/api/SolverContext.java
@@ -66,6 +66,18 @@ public interface SolverContext extends AutoCloseable {
   ProverEnvironment newProverEnvironment(ProverOptions... options);
 
   /**
+   * Create a new {@link ProverEnvironment} which encapsulates an assertion stack and can be used to
+   * check formulas for unsatisfiability, but retains the current assertion stack of the given
+   * {@link ProverEnvironment}.
+   *
+   * @param proverToCopy An existing {@link ProverEnvironment}, whichs assertion stack is to be
+   *     copied into the new one.
+   * @param options Options specified for the prover environment. All the options specified in
+   *     {@link ProverOptions} are turned off by default.
+   */
+  ProverEnvironment copyProverEnvironment(ProverEnvironment proverToCopy, ProverOptions... options);
+
+  /**
    * Create a fresh new {@link InterpolatingProverEnvironment} which encapsulates an assertion stack
    * and allows generating and retrieve interpolants for unsatisfiable formulas. If the SMT solver
    * is able to handle satisfiability tests with assumptions please consider implementing the {@link
@@ -75,6 +87,22 @@ public interface SolverContext extends AutoCloseable {
    *     {@link ProverOptions} are turned off by default.
    */
   InterpolatingProverEnvironment<?> newProverEnvironmentWithInterpolation(ProverOptions... options);
+
+  /**
+   * Create a fresh new {@link InterpolatingProverEnvironment} which encapsulates an assertion stack
+   * and allows generating and retrieve interpolants for unsatisfiable formulas. The new {@link
+   * ProverEnvironment} retains the current assertion stack of the given {@link ProverEnvironment}.
+   * If the SMT solver is able to handle satisfiability tests with assumptions please consider
+   * implementing the {@link InterpolatingProverEnvironment} interface, and return an Object of this
+   * type here.
+   *
+   * @param proverToCopy An existing {@link ProverEnvironment}, whichs assertion stack is to be
+   *     copied into the new one.
+   * @param options Options specified for the prover environment. All the options specified in
+   *     {@link ProverOptions} are turned off by default.
+   */
+  InterpolatingProverEnvironment<?> copyProverEnvironmentWithInterpolation(
+      ProverEnvironment proverToCopy, ProverOptions... options);
 
   /**
    * Create a fresh new {@link OptimizationProverEnvironment} which encapsulates an assertion stack
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
@@ -45,6 +45,7 @@ abstract class Z3AbstractProver<T> extends AbstractProverWithAllSat<T> {
   protected final long z3context;
   private final Z3FormulaManager mgr;
 
+  // m_n_obj in Z3
   protected final long z3solver;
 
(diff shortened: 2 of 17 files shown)


src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,6 +65,18 @@ enum ProverOptions {
...
+   * Create a new {@link ProverEnvironment} which encapsulates an assertion stack and can be used to
+   * check formulas for unsatisfiability, but retains the current assertion stack of the given
+   * {@link ProverEnvironment}.
+   *
+   * @param proverToCopy An existing {@link ProverEnvironment}, whichs assertion stack is to be
+   *     copied into the new one.
+   * @param options Options specified for the prover environment. All the options specified in
+   *     {@link ProverOptions} are turned off by default.
+   */
+  ProverEnvironment copyProverEnvironment(ProverEnvironment proverToCopy, ProverOptions... options);
```
kfriedberger, line 78: LEaving the options open for the user might introduce problems, when it comes to featuers like unsat-core and interpolation, because we need to provide tokens/ids for all pushed assertions.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -76,6 +88,22 @@ enum ProverOptions {
...
+   * implementing the {@link InterpolatingProverEnvironment} interface, and return an Object of this
+   * type here.
+   *
+   * @param proverToCopy An existing {@link ProverEnvironment}, whichs assertion stack is to be
+   *     copied into the new one.
+   * @param options Options specified for the prover environment. All the options specified in
+   *     {@link ProverOptions} are turned off by default.
+   */
+  InterpolatingProverEnvironment<?> copyProverEnvironmentWithInterpolation(
+      ProverEnvironment proverToCopy, ProverOptions... options);
```
kfriedberger, line 105: Leaving the options open for the user might introduce problems, when it comes to featuers like unsat-core and interpolation, because we need to provide tokens/ids for all pushed assertions.

kfriedberger: Why is there no copy-method for the OptimizationProver? The current API is incomplete.

kfriedberger: It might be better to provide a copy-method/copy-constructor directly in the ProverEnvironment, because it knows best how to copy its content. The current naming scheme is quite verbous.

src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
```diff
@@ -73,6 +74,34 @@ abstract class Z3AbstractProver<T> extends AbstractProverWithAllSat<T> {
 
     logfile = pLogfile;
     mgr = pMgr;
+    postProcessProverCreation(pSolverOptions);
+  }
+
+  Z3AbstractProver(
```
kfriedberger, line 80: This looks like redundant code in both constructors..

src/org/sosy_lab/java_smt/test/SolverContextTest.java
```diff
@@ -119,4 +122,66 @@ public void testFormulaAccessAfterClose() {
...
+  @Test
+  public void testProverCopyWithStackAndAssertions() throws InterruptedException, SolverException {
+    requireProverCopying();
+
+    assume()
+        .withMessage(
+            "Solver %s does not support prover copying for non-base-level provers",
+            solverToUse())
+        .that(solverToUse())
+        .isNotEqualTo(Solvers.Z3);
```
kfriedberger, line 161: I do not understand this test: Is the whole copying-thing working or not?

baierd: Yes. For a stack that has not been pushed according to the error message. Which makes it unusable for the use-case of Thomas. I've updated the discussion in the Z3 repo to find out if i made a mistake or if this is intended behavior.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: So you would suggest to always copy the options 1:1?

src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
baierd: How so?

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: I agree with you on all things.
However, Z3 needs permanent saving of references for the copied prover. Since we can't guarantee that the original prover is not closed, we need to save them in the context.
Of course we could do this via a method as well.

src/org/sosy_lab/java_smt/api/SolverContext.java
kfriedberger: If the method should copy a stack and (as optimization) also the solver-internal data, then I would assume that the Prover requires identical options.
However, if we keep this open for the user and are able to copy all assertions directly onto another Prover, and limit the internal optimization to only the case of identical options, then we could cover more potential use-cases, e.g., if a user wants to enable unsat-cores or interpolants later. We need to specify this in the API.

src/org/sosy_lab/java_smt/api/SolverContext.java
kfriedberger: If Z3 requires permanent memory-tracking for the base-solver and does not allow closing it, then the copy-method is broken by design. Are you sure that there is no bug in using the copy-method?

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: I removed the options for now. We may want to test if it is possible to alter the options in copied provers first.

baierd:
By the way, with your additions in #325 , we could easily provide a default implementation for this feature.

src/org/sosy_lab/java_smt/api/SolverContext.java
kfriedberger: If we copy a solver, I would assume that the new solver-instance increments all internal references on its own to avoid garbage-collection in case the old solver-instance is closed. I do not see a requirement for JavaSMT to keep track of the old solver-instance.

kfriedberger:
Another question: what happens to the returned tokens (formula-names or partition-ids or random ids) related to added/pushed constraints for interpolation? Are the old tokens also valid in the new environment, or is a user required to get new ones (and how should that happen?)?

PR #325, opened 2023-07-31
Unify constraints tracking across all solvers

kfriedberger:


src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -22,8 +33,10 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
   protected final boolean generateUnsatCores;
   private final boolean generateUnsatCoresOverAssumptions;
   protected final boolean enableSL;
+  protected boolean closed = false;
 
   private final Set<Evaluator> evaluators = new LinkedHashSet<>();
+  private final List<Collection<BooleanFormula>> assertedFormulas = new ArrayList<>();
```
baierd, line 39: Is there a benefit to `ArrayList` compared to the `Deque` that we use in most cases where an assertion stack is needed?

src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
```diff
@@ -208,27 +215,31 @@ protected long addConstraint0(BooleanFormula f) throws InterruptedException {
...
-  public void push() throws InterruptedException {
+  protected void push0() throws InterruptedException {
     Preconditions.checkState(!closed);
-    try {
-      Native.solverPush(z3context, z3solver);
-    } catch (Z3Exception exception) {
-      throw creator.handleZ3Exception(exception);
+    super.push();
+    if (storedConstraints != null) {
+      storedConstraints.push(storedConstraints.peek());
```
baierd, line 222: This puts the previous level on the new level, is this on purpose?

baierd:
There is also a optimization test failing for appveyor for some reason.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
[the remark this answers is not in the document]
kfriedberger: There is no negative impact and I was thinking about the stack-cloning and perhaps even a method for assertion-retrieval for a given level, i.e., what was given to the solver can be extracted back to the user.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -22,8 +33,10 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
...
 
   private static final String TEMPLATE = "Please set the prover option %s.";
```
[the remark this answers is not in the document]
kfriedberger: The name is condensed from 5 subclasses with identical name :-) I will think about the name.

src/org/sosy_lab/java_smt/solvers/z3/Z3AbstractProver.java
kfriedberger: This is on purpose, because:
- we use the same strategy in several places (all places are part of this PR).
- we always require the complete union of all constraints when it comes to unsat-cores or interpolants.
- the PersistantMap is quite memory efficient and shares common sub-mappings in most cases.

kfriedberger:
The failing AppVoyer-tests are strange, because I did not modify any Windows-specific code.

baierd:
Maybe a fluke? Can you restart them?

kfriedberger:
There were some inconsistencies in the Z3 implemention for the optimization-prover. Should be fixed by now.

PR #334, opened 2023-10-08
Add opensmt2 (from branch to master)

kfriedberger:


daniel-raffler (commented):
(Edit: This was meant to be a comment on [4543d4d](https://github.com/sosy-lab/java-smt/pull/334/commits/4543d4d2671dc114f036395f47ef4a3045135ed6))

Hello Karlheinz,
note that OpenSMT does not fully implement QF_AUFLIRA as mixed integer-real terms are not supported. The logic was added specifically for us as "the most general logic OpenSMT can support" (see [here](https://github.com/usi-verification-and-security/opensmt/discussions/614#discussioncomment-6326541)). Maybe I could ask the developers to rename it to ALL?

kfriedberger:
Hi @daniel-raffler .

'ALL' does not include all posible logics, but only the ones supported by OpenSMT.
I already saw that commit earlier and wanted to keep the naming consistent.
That way, the user of JavaSMT does exaclty see which logics are available.

kfriedberger:
The last commit a41bece does improve responses from CPAchecker in several benchmark tasks. However, OpenSMT suffers from SegmentationFaults in some cases, where the given query uses logic combination.

Example: 
```
scripts/cpa.sh \
    -heap 13000M -noout -benchmark -stack 50M -timelimit 120s -stats -32 \
    -setprop cpa.predicate.memoryAllocationsAlwaysSucceed=true \
    -setprop counterexample.export.allowImpreciseCounterexamples=true \
    -predicateAnalysis -setprop cpa.predicate.encodeBitvectorAs=INTEGER \
    -setprop solver.solver=OPENSMT \
    -setprop cpa.predicate.encodeFloatAs=UNSUPPORTED \
    -setprop solver.opensmt.logic=QF_LIA \
    -spec ../sv-benchmarks/c/properties/unreach-call.prp \
    ../sv-benchmarks/c/loops/eureka_01-1.c
``` 
This example crashes in:
```
C  [libopensmt.so+0x1fbb73]  LASolver::getLAVar_single(PTRef)+0x23
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  org.sosy_lab.java_smt.solvers.opensmt.api.OsmtNativeJNI.MainSolver_check(JLorg/sosy_lab/java_smt/solvers/opensmt/api/MainSolver;)J+0
j  org.sosy_lab.java_smt.solvers.opensmt.api.MainSolver.check()Lorg/sosy_lab/java_smt/solvers/opensmt/api/sstat;+9
j  org.sosy_lab.java_smt.solvers.opensmt.OpenSmtAbstractProver.isUnsat()Z+61
j  org.sosy_lab.java_smt.basicimpl.withAssumptionsWrapper.BasicProverWithAssumptionsWrapper.isUnsat()Z+8
``` 

The optimal solution would be to improve OpenSMT, such that it returns an an error for such cases instead of SegFault.

kfriedberger:
There are a few minor points that need to be fixed in the future, but this can be done outside of this PR:
- avoid segmentation faults with invalid API-usage (most part are internally in OpenSMT)
- avoid giving integer tokens for interpolation assertiona towards the user, or validate them when using them (this point might be a weakness of several native solvers: the user can insert arbitrary values and access memory)
- re-visit model evaluation for arrays (perhaps a partial model is better than aborting on arrays)

kfriedberger:
Some benchmarks in CPAchecker show that the current bindings for OpenSMT work well:

Predicate-Int:
image

kInduction-Int:
image

PR #344, opened 2023-12-19
Added a solver-independent parser-interpreter and code-generator for SMT-LIB2 to JavaSMT

KJanelle:
Integrating a parser-interpreter and a code-generator for SMT-LIB2 into JavaSMT as well as the usage of the Princess binary as a solver.

Theories supported by parser and generator:

    Boolean
    Integer
    Real
    Bitvector
    Array
    Uninterpreted Functions

Generator does not support modular congruence, because SMT-LIB2 does not (as far as I was able to find out).
Parser wraps (define-fun ...) and (let ...) for JavaSMT.
(as const ...) is interpreted as regular array, since JavaSMT does not support otherwise.

Value assignments in models from the Princess binary always use (define-fun), their conversion from SMT-LIB2 to JavaSMT is implemented accordingly.

argumentsInterpretation in Model.ValueAssignment is always empty, since it's not possible to gather the necessary information from the model String returned by the Princess binary.

Main source for the supported operators was https://smtlib.cs.uiowa.edu/theories.shtml.

As the existing solvers process constraints differently, not all tests work for all solvers (e.g. the tests for Z3 only succeed if they are carried out individually). The focus was on testing the usage with the Princess binary.


.checkstyle
baierd, line 1: Please restore this file.

.gitignore
baierd, line 1: Please revert this file to the original.

.idea/ant.xml
baierd, line 1: Revert this file to the state in the main branch please.

.idea/compiler.xml
baierd, line 1: Revert this file to the state in the main branch please.

.idea/encodings.xml
baierd, line 1: Revert this file to the state in the main branch please.

.idea/inspectionProfiles/Project_Default.xml
baierd, line 1: Revert this file to the state in the main branch please.

.idea/JavaSMT.iml
```diff
@@ -1,15 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
-<!--
-This file is part of JavaSMT,
-an API wrapper for a collection of SMT solvers:
-https://github.com/sosy-lab/java-smt
-
```
baierd, line 7: Re-Add the header please.

.idea/JavaSMT.iml
```diff
@@ -313,6 +311,15 @@ SPDX-License-Identifier: Apache-2.0
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="module-library">
+      <library>
+        <CLASSES>
+          <root url="jar://$USER_HOME$/Downloads/antlr-4.13.1-complete.jar!/" />
```
baierd, line 317: Delete/Change this entry please. It refers to your system. It should look like the other ones with `MODULE_DIR` instead of `USER_HOME`.

.idea/misc.xml
```diff
@@ -1,25 +1,61 @@
...
-This file is part of JavaSMT,
-an API wrapper for a collection of SMT solvers:
-https://github.com/sosy-lab/java-smt
-
-SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
-
-SPDX-License-Identifier: Apache-2.0
--->
-
 <project version="4">
```
baierd, line 1: Restore the deleted information (e.g. project version) please.

.idea/modules.xml
baierd, line 1: Revert this file to the state in the main branch please.

SMT2_UF_and_Array.smt2
baierd, line 1: Please add a comment in the beginning of the file, explaining its usage.

build.xml
baierd, line 1: Revert this file to the state in the main branch please.

build/build-publish-solvers.xml
baierd, line 1: Please translate everything to englisch and add a comment which princess version you refer to here! Because its not our default version. Also, add a comment that we will try to ask the dev to provide this version of Princess himself.

build/build-publish.xml
baierd, line 1: Revert this file to the state in the main branch please.

lib/ivy.xml
```diff
@@ -163,6 +163,8 @@ SPDX-License-Identifier: Apache-2.0
         <dependency org="org.sosy_lab" name="javasmt-solver-cvc4" rev="1.8-prerelease-2020-06-24-g7825d8f28" conf="runtime-cvc4->solver-cvc4" />
         <dependency org="org.sosy_lab" name="javasmt-solver-cvc5" rev="1.0.5-g4cb2ab9eb" conf="runtime-cvc5->solver-cvc5" />
         <dependency org="org.sosy_lab" name="javasmt-solver-boolector" rev="3.2.2-g1a89c229" conf="runtime-boolector->solver-boolector" />
+        <!-- <dependency org="org.sosy_lab" name="package-princess" rev="2023-06-19" conf="solver-princess" /> -->
```
baierd, line 166: Remove this please.

baierd (commented):
Also please delete all files not related to JavaSMT (result files of your benchmarking w CPAchecker etc.), as well as all files deleted by the buildsystem that you added.

src/org/sosy_lab/java_smt/SolverContextFactory.java
```diff
@@ -284,6 +290,15 @@ private SolverContext generateContext0(Solvers solverToCreate)
         return PrincessSolverContext.create(
             config, shutdownNotifier, logfile, (int) randomSeed, nonLinearArithmetic);
 
+      case PRINCESS_BINARY:
```
daniel-raffler, line 293: This is exactly the same as in the 'Princess' branch. We probably don't need a new name for the binary solver backend. Instead we could just set the option `generateSMTLIB2` to true, and then use `ProverOptions.useBinary` whenever a new ProverEnvironment is created. Maybe it would also be possible to turn this into a single option?

We may also support more binary backends in the future?

src/org/sosy_lab/java_smt/basicimpl/AbstractArrayFormulaManager.java
```diff
@@ -39,8 +39,11 @@ public <TI extends Formula, TE extends Formula> TE select(
     //    Rational
 
     final TFormulaInfo term = select(extractInfo(pArray), extractInfo(pIndex));
-
-    return (TE) getFormulaCreator().encapsulate(elementType, term);
+    TE result = (TE) getFormulaCreator().encapsulate(elementType, term);
+    if (Generator.isLoggingEnabled()) {
+      ArrayGenerator.logSelect(result, pArray, pIndex);
```
daniel-raffler, line 44: Logging for the SMTLIB2 generator is done using the same global generator objects. This is an issue when more than one solver instance tries to output SMTLIB2 code at the same time.
We should consider reimplementing this using a delegate. At  the very least the logged SMTLIB2 output needs to be linked to the SolverContext that created it.

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
```diff
@@ -82,6 +84,12 @@ public void close() {
    */
   @Override
   public boolean isUnsat() throws SolverException, InterruptedException {
+    try {
+      Generator.dumpSMTLIB2();
```
daniel-raffler, line 88: This will write the generated SMTLIB2 file to a file called `out.smt2`. We may want add a method to get the script as a string instead.

src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
```diff
@@ -74,19 +80,29 @@ protected PrincessAbstractProver(
...
-    if (result.equals(SimpleAPI.ProverStatus$.MODULE$.Sat())) {
-      wasLastSatCheckSat = true;
-      return false;
-    } else if (result.equals(SimpleAPI.ProverStatus$.MODULE$.Unsat())) {
-      return true;
-    } else if (result.equals(SimpleAPI.ProverStatus$.MODULE$.OutOfMemory())) {
-      throw new SolverException(
-          "Princess ran out of stack or heap memory, try increasing their sizes.");
+    if (useBinary) {
+      binaryModel.getOutput();
```
daniel-raffler, line 91: Here we use the Princess binary to read back the SMTLIB2 script we just dumped to `Out.smt2`. The output from the binary is then redirected to a file called `Model.smt2`, and later read back in to parse the result. We may look into using pipes instead to avoid problems when multiple solver instances are used simultaneously.

Issue #339, opened 2023-11-09
StackOverflowError using MathSAT5


baierd:
When using Mathsat with the following commandline on CPAchecker we end up with an StackOverflowError whos source is not entirely clear. We should investigate this.

`scripts/cpa.sh -svcomp24 -stats -spec test/programs/benchmarks/properties/unreach-call.prp test/programs/benchmarks/bitvector/gcd_2.c`

```
Exception in thread "main" java.lang.StackOverflowError
    at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_solve(Native Method)
    at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_check_sat(Mathsat5NativeApi.java:156)
    at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5AbstractProver.isUnsat(Mathsat5AbstractProver.java:103)
    at org.sosy_lab.cpachecker.util.predicates.smt.BasicProverEnvironmentView.isUnsat(BasicProverEnvironmentView.java:61)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager$Interpolator.checkInfeasabilityOfTrace(InterpolationManager.java:1097)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager$Interpolator.buildCounterexampleTrace(InterpolationManager.java:973)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.buildCounterexampleTrace0(InterpolationManager.java:446)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.lambda$buildCounterexampleTrace$0(InterpolationManager.java:328)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.callWithTimelimit(InterpolationManager.java:337)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.buildCounterexampleTrace(InterpolationManager.java:327)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.interpolate(InterpolationManager.java:401)
    at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.interpolate(InterpolationManager.java:380)
[...]

PR #345, opened 2023-12-19
339 stackoverflowerror using mathsat5

daniel-raffler:
Hello everyone,
this is my fix for bug #339. The issue was in the implementation of the termination call back for MathSAT in our JNI bindings. In order to check whether the shutdown manager has been triggered we need to provide MathSAT with a way of calling back our Java code. For this the current Jenv pointer is stored when a new prover environment instance is created. Unfortunately this pointer is only valid for the current thread, and if isUnsat is later called from any other thread we get a StackOverflowError. The solution is to only install the callback right before the native call to MathSAT in isUnsat. This way we can be sure to be on the right thread when the JEnv is saved. Once MathSAT returns we can then remove the callback again.

In addition to the bug fix this pull request also includes a new test class "SolverThreadLocalityTest" that is meant to find similar errors in other solvers.


src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
```diff
@@ -97,18 +97,32 @@ private long buildConfig(Set<ProverOptions> opts) {
...
   protected abstract void createConfig(Map<String, String> pConfig);
 
+  private <T> T exec(Callable<T> closure) throws SolverException, InterruptedException {
+    long hook = context.addTerminationTest(curEnv);
+    T value = null;
+    try {
+      value = closure.call();
+    } catch (Throwable t) {
+      Throwables.propagateIfPossible(t, IllegalStateException.class, SolverException.class);
+      Throwables.propagateIfPossible(t, InterruptedException.class);
```
PhilippWendler, line 107: There is a bug here: This will silently swallow all checked exceptions except for the mentioned classes. One always has to make sure to end this list of `propagateIfPossible` with a catch all that throws any remaining exception (e.g., by wrapping it in an `AssertionError`).

Furthermore: `IllegalStateException` is unchecked and is thus redundant here. The JavaDoc of `propagateIfPossible` says that this method is discouraged and mentions a replacement.

But of course, as argued above, all this could go away anyway.

src/org/sosy_lab/java_smt/test/SolverThreadLocalityTest.java
```diff
@@ -0,0 +1,327 @@
...
+import org.sosy_lab.java_smt.api.FormulaManager;
+import org.sosy_lab.java_smt.api.IntegerFormulaManager;
+import org.sosy_lab.java_smt.api.InterpolatingProverEnvironment;
+import org.sosy_lab.java_smt.api.SolverContext;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
+import org.sosy_lab.java_smt.api.SolverException;
+import org.sosy_lab.java_smt.solvers.opensmt.Logics;
+
+public class SolverThreadLocalityTest extends SolverBasedTest0.ParameterizedSolverBasedTest0 {
+  ExecutorService executor = Executors.newFixedThreadPool(2);
```
PhilippWendler, line 35: Resources like this should be properly handled with `@Before` and `@After`. As written each execution of this test leaks threads!

src/org/sosy_lab/java_smt/test/SolverThreadLocalityTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+    try (SolverContext newContext = factory.generateContext()) {
+      try (BasicProverEnvironment<?> prover =
+          newContext.newProverEnvironment(ProverOptions.GENERATE_MODELS)) {
+        // Trying to add a formula from our global context to the newly created solver context.
+        prover.push(formula);
+        assertThat(prover).isUnsatisfiable();
+      }
+    } catch (InvalidConfigurationException e) {
+      throw new RuntimeException(e);
```
PhilippWendler, line 324: In tests, it is easier to let all kinds of exceptions propagate instead of wrapping them in an unchecked exception, and it provides shorter stacktraces.

daniel-raffler:
Hello Philipp,
thanks a lot for the review! I think I've now addressed all of your points. Is there anything else you would like me to change?

PhilippWendler:
I guess Daniel will do a full review, I just stumbled upon this because I was interested in this issue (good find btw.!) and left comments where I noticed something.

I guess you could get rid of the `boolean result` that you have in the two changed methods.

kfriedberger (approved):
The bugfix is a great outcome, well done! The tests are well written, too.

The steps for debugging and narrowing the problem (cf. commit history of this PR) are very interesting.

:+1:

daniel-raffler:
Thanks for your help, Karlheinz! I've talked to Daniel and he will merge the branch in the next few days.

PR #346, opened 2023-12-20
314 adding the abstract numeric domains of the apron library as smt solver

winnieros:
Integration of the APRON Numeric Abstract Domain Library as a quasi SMT solver supporting Integer, Rational and Boolean.
Problems with the implementation:

- Apron doesn't support UFs, no evaluation with CPAchecker possible
- limited functionalities of Boolean
- Abstract interpretation uses overapproximation of the value set which leads to unsound results; that does not go in line with SMT; no possibility was found to inform the user about possible overapproximation
- Model is not trustworthy; ex.: `x = 3, y = 1/2` gives `]-infinity, infinity[` as value set for `x = y +3`


src/org/sosy_lab/java_smt/solvers/apron/ApronModel.java
```diff
@@ -0,0 +1,320 @@
...
+        return getIntVarValue((ApronIntVarNode) pNode);
+      } else if (pNode instanceof ApronRatVarNode) {
+        return getRatVarValue((ApronRatVarNode) pNode);
+      } else {
+        return getComplexValue(pNode);
+      }
+    }
+  }
+
+  private ApronNode getComplexValue(ApronNode pNode) {
```
baierd, line 227: The method is quite big, maybe it would be good to split it up.

src/org/sosy_lab/java_smt/solvers/apron/ApronModel.java
```diff
@@ -0,0 +1,320 @@
...
+
+  private ApronNode getIntVarValue(ApronIntVarNode varNode) {
+    ApronIntCstNode nullValue = new ApronIntCstNode(BigInteger.ZERO);
+    String varName = varNode.getVarName();
+    for (ValueAssignment assignment : model) {
+      if (varName.equals(assignment.getName())) {
+        return (ApronNode) assignment.getValueAsFormula();
+      }
+    }
+    return nullValue;
```
baierd, line 307: It seems like you can just return and create the `new ApronIntCstNode(BigInteger.ZERO);` here. 
Also goes for the method below.

src/org/sosy_lab/java_smt/solvers/apron/ApronRationalFormulaManager.java
```diff
@@ -0,0 +1,260 @@
...
+      return new ApronRatCstNode(num, den);
+    }
+    return new ApronRatCstNode(num, BigInteger.ONE);
+  }
+
+  @Override
+  protected ApronNode negate(ApronNode pParam1) {
+    ApronRatUnaryNode unaryNode = new ApronRatUnaryNode(pParam1,
+        Texpr1UnNode.OP_NEG);
+    return unaryNode;
```
baierd, line 112: Why not return it directly?

src/org/sosy_lab/java_smt/solvers/apron/ApronRationalFormulaManager.java
```diff
@@ -0,0 +1,260 @@
...
+  protected ApronNode sumImpl(List<ApronNode> operands) {
+    if (!operands.isEmpty()) {
+      ApronNode first = operands.remove(0);
+      for (ApronNode operand : operands) {
+        first = new ApronRatBinaryNode(first, operand,
+            Texpr1BinNode.OP_ADD);
+      }
+      return first;
+    }
+    return null;
```
baierd, line 132: Is `null` the correct return here?
Maybe throw an exception instead?

daniel-raffler:
@baierd The code looks good so far. Can you upload the solver backend to Ivy? (Note that the binaries will depend on system supplied MPFR and GMP libraries. If this is a problem we can probably still change our build script to include them with the installation.)

This is what I used to build the binaries:
```ant publish-apron -Dapron.path=/home/daniel/workspace/apron-JavaSMT -Dapron.customRev=0.9.14```

(We need the `apron` version from your fork for this: `git clone git@github.com:baierd/apron-JavaSMT.git`)

daniel-raffler:
> (We need the `apron` version from your fork for this: `git clone git@github.com:baierd/apron-JavaSMT.git`)

I could also create a patch from your changes and include it with JavaSMT. This is probably easier to maintain than keeping the fork updated?

src/org/sosy_lab/java_smt/solvers/apron/ApronRationalFormulaManager.java
daniel-raffler: I'm returning 0 now.

baierd:
@kfriedberger we should decide if we want to add this Apron based SMT solver to JavaSMT or not. It has the same limitations as dReal due to being based on abstract interpretation.

However, we could try using [ELINA](https://github.com/eth-sri/ELINA) instead of Apron, which seems to be more precise and faster. However, both projects seem to be abandoned.

PR #348, opened 2024-01-07
342 add a debug mode

daniel-raffler:
Adds a debug mode that helps find common user errors
- We check that solver structures are only accessed from the thread that created them
- We make sure that formulas are not used outside of their context
- Configuration options can be used to enable/disable the checks

EDIT:
Added a patch for testing: https://gist.github.com/daniel-raffler/6bfc3044232ea1cd542f43d9442f5495
This will run all JavaSMT tests in debug mode


daniel-raffler:
Hello everyone,
I've run some benchmark to test the performance impact of adding debug mode as a new feature. Here are the results for BMC(k=1) with CPAchecker on the sv-comp24 test set:
Screenshot from 2024-04-27 03-03-35
For reference, these are the results we get when running the same analysis with the JavaSMT version from the master branch:
Screenshot from 2024-04-27 03-03-47
As you can see the performance impact is significand, however, only when debug mode is actually selected. When the option is not being used there is no additional overhead.

daniel-raffler:
The results for memory use show the same picture:
Screenshot from 2024-04-27 03-04-14
Screenshot from 2024-04-27 03-04-24

src/org/sosy_lab/java_smt/delegate/debugging/DebuggingSolverContext.java
```diff
@@ -0,0 +1,239 @@
...
+import org.sosy_lab.java_smt.api.ProverEnvironment;
+import org.sosy_lab.java_smt.api.SolverContext;
+import org.sosy_lab.java_smt.api.visitors.DefaultFormulaVisitor;
+import org.sosy_lab.java_smt.api.visitors.TraversalProcess;
+
+// TODO: Figure out what to do about boolector (does not support visitors..)
+// TODO: Add an option that checks if sorts were created on the context
+
+@Options(prefix = "solver.debugMode")
+public class DebuggingSolverContext extends DefaultFormulaVisitor<TraversalProcess>
```
kfriedberger, line 39: This class mixes three contepts, which are SolverContext, Visitor, and a central point for DebugAssertions. It might be easier to use a seperated classes for those three concepts, i.e., a separated visitor and a specialized DebuggingContext that contains the checks. Those three classes can share fields if required.

kfriedberger (changes requested):
Overall, a valid solution for checking threads and contexts.
I wrote a small request for changes. Looks good so far.

src/org/sosy_lab/java_smt/delegate/debugging/DebuggingSolverContext.java
daniel-raffler: I've split up the class now. `DebuggingSolverContext` now only contains the methods of the JavaSMT interface, a new class `DebuggingContext`is used for the options and the bookkeeping data, and the visitor method was moved to an anonymous class that is created directly in `addFormulaTerm`. Hopefully this will make the code a little easier to read.

kfriedberger (approved):
Looks good to me.  
This PR can be merged.

PR #349, opened 2024-01-12
Add user propagator feature

ThomasHaas:
This PR adds the user propagator feature to JavaSMT (~ user-defined theory solvers).
Currently, the only SMT backend that supports this feature is Z3, but CVC5 also seems to be working on adding support for user propagators.
For now, only the most important (and most general) user propagator features are exposed.

The PR contains 3 example implementations: `SimpleUserPropagator` and two propagators for `Nqueens`.
Both `NQueens` propagators are enumerating all models of an `Nqueens` instance, but one of them also implements the placement constraints inside the user propagator (the other works on the original encoding).
The user propagator gives tremendeous speed-up when enumerating the models: we measure up-to two orders of magnitude speed-up (a couple of minutes rather than hours for large instances).

The speed-up can be easily observed on Nqueens with `N=11`: iterating all models with the classical approach takes around 22 seconds (on my machine), whereas the user propagator solutions need only about 2 seconds. The difference gets larger with larger `N`.

~**Problems:** The user propagator feature needs the recent version of Z3 and does not work with the current version used by JavaSMT.~ Thanks @kfriedberger for updating the Z3 version.

*Edit:* I just noticed that I accidentally copied the license headers into some of the source files.
How are the licensing rules? Shall I remove the headers or replace them using my name?

### DIFF
--- a/src/org/sosy_lab/java_smt/api/PropagatorBackend.java
+++ b/src/org/sosy_lab/java_smt/api/PropagatorBackend.java
@@ -1,0 +1,116 @@
+/*
+ *  JavaSMT is an API wrapper for a collection of SMT solvers.
+ *  This file is part of JavaSMT.
+ *
+ *  Copyright (C) 2007-2016  Dirk Beyer
+ *  All rights reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.sosy_lab.java_smt.api;
+
+/**
+ * The PropagatorBackend class is used by {@link UserPropagator} to implement custom user
+ * propagators.
+ * It contains functions to interact with the SAT/SMT core during solving, for example, it provides
+ * the ability to propagate conflicts.
+ */
+public interface PropagatorBackend {
+
+  /**
+   * Registers an expression to be observed by a {@link UserPropagator}.
+   * Whenever the value of a registered expression becomes known by the solver, this will
+   * get reported to the user propagator by invoking the callback
+   * {@link UserPropagator#onKnownValue} (if notification was enabled via
+   * {@link #notifyOnKnownValue}).
+   * Similarly, equalities between registered expressions are reported via
+   * {@link UserPropagator#onEquality} (if enabled via
+   * {@link #notifyOnEquality()}.
+   * @param theoryExpr The expression to observe.
+   */
+  void registerExpression(BooleanFormula theoryExpr);
+
+  /**
+   * Raises a conflict caused by a set of conflicting assignments. Shall only be called
+   * from within a callback by {@link UserPropagator} such as
+   * {@link UserPropagator#onKnownValue} and
+   * {@link UserPropagator#onFinalCheck()}.
+   *
+   * <p> Effectively causes the solver to learn the implication "/\ conflictLiterals => false"
+   *
+   * @param conflictLiterals A set of inconsistent assignments.
+   */
+  void propagateConflict(BooleanFormula[] conflictLiterals);
+
+  /**
+   * Propagates a consequence implied by a set of assigned literals. Shall only be called
+   * from within a callback by {@link UserPropagator} such as
+   * {@link UserPropagator#onKnownValue} and
+   * {@link UserPropagator#onFinalCheck()}.
+   *
+   * <p> Effectively causes the solver to learn the implication "/\ assignedLiterals => consequence"
+   *
+   * @param assignedLiterals A set of assigned literals.
+   * @param consequence The consequence implied by the assigned literals.
+   */
+  void propagateConsequence(BooleanFormula[] assignedLiterals, BooleanFormula consequence);
+
+  /**
+   * Propagates a consequence as well as a set of equalities implied by a set of assigned literals.
+   * Shall only be called from within a callback by {@link UserPropagator} such as
+   * {@link UserPropagator#onKnownValue} and
+   * {@link UserPropagator#onFinalCheck()}
+   *
+   * <p> Effectively causes the solver to learn the implication "/\ assignedLiterals =>
+   *    (consequence /\ forall i: equalitiesLHS[i] == equalitiesRHS[i])"
+   *
+   * @param assignedLiterals A set of assigned literals.
+   * @param equalitiesLHS The left-hand sides of implied equalities.
+   * @param equalitiesRHS The right-hand sides of implied equalities.
+   * @param consequence The consequence implied by the assigned literals.
+   */
+  void propagateConsequenceWithEqualities(BooleanFormula[] assignedLiterals,
+                                          BooleanFormula[] equalitiesLHS,
+                                          BooleanFormula[] equalitiesRHS,
+                                          BooleanFormula consequence);
+
+  /**
+   * Enables tracking of expression values for the associated {@link UserPropagator} via
+   * {@link UserPropagator#onKnownValue(BooleanFormula, BooleanFormula)}.
+   *
+   * <p>This function is typically called from {@link UserPropagator#initialize()} if the
+   * theory solver needs to listen to the value of expressions registered by
+   * {@link #registerExpression}.
+   */
+  void notifyOnKnownValue();
+
+  /**
+   * Enables tracking of equalities for the associated {@link UserPropagator} via
+   * {@link UserPropagator#onEquality(BooleanFormula, BooleanFormula)}.
+   *
+   * <p>This function is typically called from {@link UserPropagator#initialize()} if the
+   * theory solver needs to observe equalities between expressions registered by
+   * {@link #registerExpression}.
+   */
+  void notifyOnEquality();
+
+  /**
+   * Enables the final callback {@link UserPropagator#onFinalCheck()} that is invoked
+   * when the solver finds a full satisfying assignment.
+   *
+   * <p>This function is typically called from {@link UserPropagator#initialize()} if the
+   * theory solver needs to perform final consistency checks.
+   */
+  void notifyOnFinalCheck();
+}
--- a/src/org/sosy_lab/java_smt/api/PropagatorBackend.java
+++ b/src/org/sosy_lab/java_smt/api/PropagatorBackend.java
@@ -34,9 +34,7 @@ public interface PropagatorBackend {
    * get reported to the user propagator by invoking the callback
    * {@link UserPropagator#onKnownValue} (if notification was enabled via
    * {@link #notifyOnKnownValue}).
-   * Similarly, equalities between registered expressions are reported via
-   * {@link UserPropagator#onEquality} (if enabled via
-   * {@link #notifyOnEquality()}.
+   *
    * @param theoryExpr The expression to observe.
    */
   void registerExpression(BooleanFormula theoryExpr);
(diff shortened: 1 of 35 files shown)


kfriedberger:
Hi Thomas,
thanks for your PR. The user-propagator is very interesting.

### Z3
We will update Z3 to the latest version soon. Afterwards, I will take a look at this PR.

### License
JavaSMT is open-source under Apache-2.0 and there are currently no plans to change the license. For students and other contributers, the simplest approach is to just copy the existing default licencing information and use a header-text like
```
// This file is part of JavaSMT,
// an API wrapper for a collection of SMT solvers:
// https://github.com/sosy-lab/java-smt
//
// SPDX-FileCopyrightText: {{YEAR}} Dirk Beyer <https://www.sosy-lab.org>
//
// SPDX-License-Identifier: Apache-2.0
```
I am not sure whether external persons that contribute larger parts of code have to sign some kind of copyright/licensing-transfer, e.g., to allow future changes in the name of Sosy-Lab. We had such a process internally once, long time ago. For such details, we need to ask @PhilippWendler or @dbeyer.

ThomasHaas:
Thanks for the quick update to Z3's version.

ThomasHaas:
For the Nqueens example, we copied the already existing Nqueens code and modified it to accomodate for user propagators.
I can try to merge both examples into one to reduce redundancy if that is desired.

ThomasHaas:
I just fixed some issues in Z3's user propagators for Java ([PR](https://github.com/Z3Prover/z3/pull/7088)).
Without those changes, it is currently not possible to implement the `decide`-callback which allows the user to react to and modify decisions of the solver.
I guess it is unlikely that you update the Z3 version of JavaSMT by building from sources, right? :)

kfriedberger:
We can provide Z3 from sources for Linux (64bit) via Ivy and Maven, but we do currently not want to compile Z3 for other platforms. Z3 itself has releases regularly, so other platforms might need to wait for some specific features only a little bit longer.

Thanks for your patience. I still had not yet the time for reviewing the complete PR.

ThomasHaas:
No problem, it's not too urgent. I will add the remaining feature(s) with the new Z3 release.

ThomasHaas:
Z3 released a new version with my fixes to the user propagator. In case you update to the new Z3 version, I will add the remaining feature in this PR. But I'm also fine with postponing that feature into a future PR.

kfriedberger:
Z3 was updated to version 4.12.5.

ThomasHaas:
I added the last missing feature: listening to and modifying decisions made by the solver.

ThomasHaas:
Any estimate on when this will get reviewed? :)

src/org/sosy_lab/java_smt/api/UserPropagator.java
```diff
@@ -0,0 +1,114 @@
...
+   * <p>
+   * Note: This callback is only invoked if the user propagator enabled it via
+   * {@link PropagatorBackend#notifyOnDecision()}.
+   * @param expr The expressions whose value gets decided (usually a literal).
+   * @param value The decision value.
+   */
+  void onDecision(BooleanFormula expr, boolean value);
+
+  /**
+   * Connects this user propagator with a {@link PropagatorBackend}. The backend is used
```
kfriedberger, line 80: Can this method be called mutliple times?
If yes, is any existing propagator replaced by the new one?
If no, do we want to throw an exception?
PLease document this behaviour.

src/org/sosy_lab/java_smt/api/UserPropagator.java
```diff
@@ -0,0 +1,114 @@
...
+   *   <li>
+   *     The callback {@link UserPropagator#onDecision} gets invoked right before the
+   *     solver decides on the value of a registered expression (if notification was enabled via
+   *     {@link PropagatorBackend#notifyOnDecision()}).
+   *   </li>
+   * </ul>
+   *
+   * @param theoryExpr The expression to observe.
+   */
+  void registerExpression(BooleanFormula theoryExpr);
```
kfriedberger, line 111: Does this expression need to be a symbol (variable) or can it be a complex boolean formula?
Can we also register non-boolean symbols?

Examples:
- Can we register to "a&&b"? Would the propagator then be called for each operation involving "a&&b" in that exact combination/operation?
- Can we register "counter==42"? Would the propagator then be called for each operation involving "counter"?

src/org/sosy_lab/java_smt/api/PropagatorBackend.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.Optional;
+
+/**
+ * The PropagatorBackend class is used by {@link UserPropagator} to implement custom user
+ * propagators.
+ * It contains functions to interact with the SAT/SMT core during solving, for example, it provides
+ * the ability to propagate conflicts and to influence the decision-making.
+ */
+public interface PropagatorBackend {
```
kfriedberger, line 19: This class is for internal usage only. A user does not need to implement it himself, but JavaSMT internally provides solver-specific implementation for it. Correct?

src/org/sosy_lab/java_smt/api/UserPropagator.java
kfriedberger: Can we add a user-propagator at each point in time, e.g., after pushing several formulas? Or should we create a user propagator before creating any symbols? For Z3, it seems to be sufficient to just instantiate the uSerPropagator right before callung the SMT/SAT-check.

src/org/sosy_lab/java_smt/basicimpl/AbstractUserPropagator.java
```diff
@@ -0,0 +1,59 @@
...
+import org.sosy_lab.java_smt.api.UserPropagator;
+import org.sosy_lab.java_smt.api.PropagatorBackend;
+
+public abstract class AbstractUserPropagator implements UserPropagator  {
+
+  protected @Nullable PropagatorBackend backend;
+
+  @Override
+  public void onFinalCheck() {
+
```
kfriedberger, line 35: This does nothing? Is this the intended behaviour of the abstract super-class?
Maybe at least insert a check for "Preconditions.notNull(backend)".

src/org/sosy_lab/java_smt/api/UserPropagator.java
ThomasHaas: You can register any boolean expression, even `counter==42`. However, the propagator will notify on the value of the boolean expression and not on any variables inside. Basically, the propagator works on the boolean skeleton of the formula.
So you cannot track the precise value of `counter` and only indirectly know about its value by tracking atoms/predicates over `counter.`

Also, the propagator is more on the semantic level rather the syntactic one. If your whole asserted formula is `a || b` but you registered `a && b`, the propagator will still notify you once its partially found model fixes the value of `a && b`.
In this case, the solver will likely do this:
```
Decide a = false
Unit propagate b = true
Report (a&&b) == false to UserPropagtor
```
You can run the `SimpleUserPropagator` example I added which showcases various interactions and prints the solver steps involved.

src/org/sosy_lab/java_smt/basicimpl/AbstractUserPropagator.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public final void injectBackend(PropagatorBackend backend) {
+    this.backend = Preconditions.checkNotNull(backend);
+  }
+
+  @Override
+  public void registerExpression(BooleanFormula theoryExpr) {
+    Preconditions.checkState(backend != null);
```
kfriedberger, line 55: For Null-Checks there exists a method Preconditions.checkNotNull that might be better.

src/org/sosy_lab/java_smt/api/PropagatorBackend.java
ThomasHaas: Yes, it is implemented internally only (and could me marked `sealed` in new Java Versions).
The user only interacts with this backend via this theory solver.

src/org/sosy_lab/java_smt/basicimpl/AbstractUserPropagator.java
ThomasHaas: I didn't want to force the user to implement methods if he does not need them, so by default they are all empty.
There are dedicated methods to enable if the callback should be called or not.

src/org/sosy_lab/java_smt/basicimpl/AbstractUserPropagator.java
ThomasHaas: I tend to use `checkNotNull` for arguments rather than internal state . I think the `IllegalStateException` is more explicit about the fact that this method was called at the wrong time/state.

src/org/sosy_lab/java_smt/api/UserPropagator.java
```diff
@@ -0,0 +1,114 @@
...
+   * @param backend The propagator backend.
+   */
+  void injectBackend(PropagatorBackend backend);
+
+  /**
+   * This method is similar to a constructor but is guaranteed to get invoked only after
+   * {@link #injectBackend} was successfully called.
+   * The user can enable notifications by accessing the injected {@link PropagatorBackend}.
+   */
+  void initialize();
```
kfriedberger, line 91: Would it make sense to inline th method "initialize" into "injectBackend" directly to simplify the user-directed API and name both together as "initializeWithBackend(backend)"?

src/org/sosy_lab/java_smt/solvers/z3/Z3TheoremProver.java
```diff
@@ -169,8 +184,12 @@ public void close() {
 
       Native.solverReset(z3context, z3solver); // remove all assertions from the solver
       Native.solverDecRef(z3context, z3solver);
+      if (propagator != null) {
+        propagator.close();
```
kfriedberger, line 188: set propagator to null after cleanup? Otherwise a second call to "close" would result in a NullPointer.

src/org/sosy_lab/java_smt/api/UserPropagator.java
ThomasHaas: I have not tried out what the limitations are, but it is definitely fine to add the user propagator anytime before the first SAT-check. You can create the whole solver formula and then add the propagator (you need to register the expressions after adding the propagator though)
Whether one can add them after the first SAT-check is something I didn't test.

src/org/sosy_lab/java_smt/solvers/z3/Z3UserPropagator.java
```diff
@@ -0,0 +1,185 @@
...
+    for (int i = 0; i < formulaArray.length; i++) {
+      if (formulaArray[i] != null) {
+        formulaInfos[i] = creator.extractInfo(formulaArray[i]);
+      }
+    }
+    return formulaInfos;
+  }
+
+  private BooleanFormula encapsulate(long z3Expr) {
+    // We fill the lowest 3 bits that are always 0 due pointer alignment.
```
kfriedberger, line 173: This is an interesting observation. It might not be universally true, but valid for most systems. Lets keep it. :-)

src/org/sosy_lab/java_smt/api/UserPropagator.java
ThomasHaas: Hmm, I separated those two to make it easier for the user to avoid any mistakes, since there is basically only a single reasonable implementation for `injectBackend` and that one can be inherited from `AbstractUserPropagator`.
I can also merge them, if you think that is better.

src/org/sosy_lab/java_smt/solvers/z3/Z3UserPropagator.java
ThomasHaas: Do you support 32-bit versions of Z3? If so, this code could break. I just added this little trick because the standard values result in a lot of collisions, especially on smaller maps.

src/org/sosy_lab/java_smt/api/UserPropagator.java
ThomasHaas: After thinking more about this, I think it is always bad to register the same user propagator multiple times.
I made this point clear in the documentation and the implementation provided by `AbstractUserPropagator` will throw an error if the method is called multiple times.

src/org/sosy_lab/java_smt/api/UserPropagator.java
ThomasHaas: I combined both methods now.

src/org/sosy_lab/java_smt/solvers/z3/Z3TheoremProver.java
ThomasHaas: I set it to null now, but in practice it would not be an issue because all of this code can only be executed once due to the surrounding condition.

src/org/sosy_lab/java_smt/solvers/z3/Z3UserPropagator.java
kfriedberger: A user can provide its own version of Z3, including 32bit.
Are you sure about collisions here? Java should already use some bitwise and/or modulo operator internally in the hashmap.

ThomasHaas: When profiling, I just noticed that the hashmap code for tree nodes was run all the time and tree nodes are only created if there are a lot of collisions (linear buckets get replaced by tree buckets if the bucket gets too full).
The hashcode for long is computed as `(l >> 32) ^ l` but the upper range tends to be constant so that the lower 3 bits end up being constant as well. The hashmap itself uses a power of two module so it retains only the lower bits of the hashcode.
As this would be really bad on its own, the hashmap tries to do some bit-mixing on the hashcode before truncating it down, but apparently it does not work too well here. That being said, my profiling was restricted to a single internal benchmark and the Nqueens examples, so I don't have extensive data.

A simple solution is to just shift by 2 rather than 3 to also make it correct on 32 bit systems (assuming pointers are well-aligned).

kfriedberger: Thanks for the nice explanation.

A sound way could be a full bitwise rotation when computing a hashable value: `(l << 61) | (l >> 3)`.
That should work well on your observation and is independent of a specific architecture. :-)

ThomasHaas: I didn't think about bitwise rotation... a much nicer solution indeed :).

ThomasHaas:
What is your policy regarding API-breaking changes?
I would consider the user propagator to be quite experimental and changes to the API are to be expected in the future.

kfriedberger:
We try to keep the number of API changes to a minimum, but we do change if neccessary.
For any breaking changes in the API, we need to increment the version number.

ThomasHaas:
I see, that makes sense.

kfriedberger:
I am currently going through the code and trying a few things.

There are still some smaller questions, but the overall idea is clear.

The following questions are out of scope for this PR, but we might want to keep track on this topic:
- We have an internal ALL-SAT-routine in the ProverEnvironment. This could be replaced by a user-propagator, at least for Z3, to have a potentially faster ALL-SAT iteration?
- There is a certain encoding in CPAchecker for converting C-code with pointers to SMT formulas. A user propagator could serve as callback mechanism and solve pointer comparisons, etc, directly and faster than using a pure SMT-based encoding. (@sosy-lab/cpachecker Does someone want to try this?)

ThomasHaas:
>     * We have an internal ALL-SAT-routine in the ProverEnvironment. This could be replaced by a user-propagator, at least for Z3, to have a potentially faster ALL-SAT iteration?

I believe this is true, since the Nqueens example is precisely such an ALL-SAT iteration. Whether it is universally more faster than adding blocking clauses from the outside is something I cannot promise though.
In my particular use-case (the reason I wanted this feature in the first place), just moving my theory reasoning into the final check of a user propagator did end up being slower by factor 2-3 (this is not an ALL-SAT problem though).
Also, this will only iterate through boolean models (not sure if your ALL-SAT iteration also gives models over complex theories).

Btw. thanks for the cleanup.

kfriedberger (approved):
The change looks good so far and it is in a mergable state.

kfriedberger:
Additonal info: our ALL-SAT receives some important predicates from the user, aka a set of boolean symbols, over which the models should be computed. We ignore symbols from other theories in the iteration.

ThomasHaas:
> Additonal info: our ALL-SAT receives some important predicates from the user, aka a set of boolean symbols, over which the models should be computed. We ignore symbols from other theories in the iteration.

That works just fine. Just register those predicates with the user propagator, collect all the assignments via the `onKnownValue`-callback and in the `onFinalCheck`-callback raise a conflict on the assignment over those predicates.

A possible disadvantage for user propagators is that they may iterate the same model multiple times so that the propagator needs to explicitly track the models found so far (which could be exponentially many). This is because clauses learned by conflicts are eligible for clause deletion.
I discussed a related issue with Nikolaj (https://github.com/Z3Prover/z3/discussions/7109) and he added a way to generate persistent clauses, but this option is not available right now since JavaSMT has no way to pass parameters to Z3.
For ALL-SAT specifically, deletable clauses might still be preferred though.

PR #355, opened 2024-02-04
Replace Z3 phantom reference map by doubly-linked list.

ThomasHaas:
I played around with the user propagator feature I implemented in #349 and ran into a performance problem related to phantom references. The issue is that the user propagator creates a `BooleanFormula` wrapper whenever the SMT solver reports a variable assignment, causing a phantom reference to get created.
This quickly generates hundreds of thousands of phantom references (or even millions?) and the `HashMap` that is used to manage phantom references degrades heavily: 80%(!) of all solving time was spent just accessing the map (this was 4 minutes from 5 minutes of total solving time!).

In this PR I propose a simple (general-purpose) solution by replacing the map with a doubly-linked list.
Doubly-linked lists provide the necessary `O(1)` deletion and insertion time while scaling to arbitrary sizes.
Furthermore, the memory footprint of the list should be strictly smaller than that of the previously used `IdentityHashMap`.

On my test benchmark, the solving time was reduced from 5 minutes down to 70 seconds, and the phantom reference overhead was just 10% rather than 80% of the total time. Interestingly, my profiler reported the 10% to be mostly from `ReferenceQueue.poll` rather than the actual cleanup code (the profiler might be inaccurate though).

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
@@ -965,8 +965,6 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
 
   /** Closing the context. */
   public void forceClose() {
-    cleanupReferences();
-
     // Force clean all ASTs, even those which were not GC'd yet.
     if (usePhantomReferences) {
       Z3AstReference cur = referenceListHead.next;
@@ -979,6 +977,12 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
       // Bulk delete everything between head and tail
       referenceListHead.next = tail;
       tail.prev = referenceListHead;
+
+      // Remove already enqueued references.
+      // NOTE: Together with the above list deletion, this will guarantee that no more ast
+      // references are reachable by the GC making them all eligible for garbage collection and
+      // preventing them from getting enqueued into the reference queue in the future.
+      while (referenceQueue.poll() != null);
     }
   }
 
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
@@ -387,22 +387,22 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
     return fpSort;
   }
 
-  private static class Z3AstReference extends PhantomReference<Z3Formula> {
-    private long z3Ast;
+  private static final class Z3AstReference extends PhantomReference<Z3Formula> {
+    private final long z3Ast;
     private @Nullable Z3AstReference prev;
     private @Nullable Z3AstReference next;
 
     // To generate dummy head and tail nodes
-    public Z3AstReference() {
-      super(null, null);
+    private Z3AstReference() {
+      this(null, null, 0);
     }
 
-    public Z3AstReference(Z3Formula referent, ReferenceQueue<? super Z3Formula> q, long z3Ast) {
+    private Z3AstReference(Z3Formula referent, ReferenceQueue<? super Z3Formula> q, long z3Ast) {
       super(referent, q);
       this.z3Ast = z3Ast;
     }
 
-    public void insert(Z3AstReference ref) {
+    private void insert(Z3AstReference ref) {
       assert next != null;
       ref.prev = this;
       ref.next = this.next;
@@ -410,7 +410,7 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
       this.next = ref;
     }
 
-    public void cleanup(Long environment) {
+    private void cleanup(long environment) {
       Native.decRef(environment, z3Ast);
       assert (prev != null && next != null);
       prev.next = next;
@@ -418,7 +418,7 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
     }
   }
 
-  private <T extends Z3Formula> T storePhantomReference(T out, Long pTerm) {
+  private <T extends Z3Formula> T storePhantomReference(T out, long pTerm) {
     if (usePhantomReferences) {
       referenceListHead.insert(new Z3AstReference(out, referenceQueue, pTerm));
     }
@@ -966,6 +966,20 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
   /** Closing the context. */
   public void forceClose() {
     cleanupReferences();
+
+    // Force clean all ASTs, even those which were not GC'd yet.
+    // Is a no-op if phantom reference handling is not enabled.
+    if (referenceListHead != null) {
+      Z3AstReference cur = referenceListHead.next;
+      assert cur != null;
+      while (cur.next != null) {
+        Native.decRef(environment, cur.z3Ast);
+      }
+      Z3AstReference tail = cur;
+      // Bulk delete everything between head and tail
+      referenceListHead.next = tail;
+      tail.prev = referenceListHead;
+    }
   }
 
   /**
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
@@ -23,11 +23,9 @@ import com.microsoft.z3.enumerations.Z3_decl_kind;
 import com.microsoft.z3.enumerations.Z3_sort_kind;
 import com.microsoft.z3.enumerations.Z3_symbol_kind;
 import java.lang.ref.PhantomReference;
-import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.math.BigInteger;
 import java.util.ArrayList;
-import java.util.IdentityHashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -120,8 +118,7 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
   /** Automatic clean-up of Z3 ASTs. */
   private final ReferenceQueue<Z3Formula> referenceQueue = new ReferenceQueue<>();
 
-  private final IdentityHashMap<PhantomReference<? extends Z3Formula>, Long> referenceMap =
-      new IdentityHashMap<>();
+  private @Nullable Z3AstReference referenceListHead = null;
 
   // todo: getters for statistic.
   private final Timer cleanupTimer = new Timer();
@@ -259,14 +256,6 @@ class Z3FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
         new Z3ArrayFormula<>(getEnv(), pTerm, pIndexType, pElementType), pTerm);
   }
 
-  private <T extends Z3Formula> T storePhantomReference(T out, Long pTerm) {
-    if (usePhantomReferences) {
-      PhantomReference<T> ref = new PhantomReference<>(out, referenceQueue);
-      referenceMap.put(ref, pTerm);
-    }
-    return out;
-  }
-
   @SuppressWarnings("unchecked")
   @Override
   public <T extends Formula> T encapsulate(FormulaType<T> pType, Long pTerm) {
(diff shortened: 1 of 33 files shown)


src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -386,16 +387,53 @@ public Long getFloatingPointType(FormulaType.FloatingPointType type) {
     return fpSort;
   }
 
+  private static class Z3AstReference extends PhantomReference<Z3Formula> {
+    private long z3Ast;
```
PhilippWendler, line 391: Should be `final`?

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -386,16 +387,53 @@ public Long getFloatingPointType(FormulaType.FloatingPointType type) {
...
+    private @Nullable Z3AstReference prev;
+    private @Nullable Z3AstReference next;
+
+    // To generate dummy head and tail nodes
+    public Z3AstReference() {
```
PhilippWendler, line 396: `public` here and for the methods makes no sense because the whole class is `final`.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -386,16 +387,53 @@ public Long getFloatingPointType(FormulaType.FloatingPointType type) {
...
+
+    public void insert(Z3AstReference ref) {
+      assert next != null;
+      ref.prev = this;
+      ref.next = this.next;
+      ref.next.prev = ref;
+      this.next = ref;
+    }
+
+    public void cleanup(Long environment) {
```
PhilippWendler, line 413: Using `Long` here and in `storePhantomReference` doesn't really provide a benefit. The current callers have `Long`s anyway so at least it doesn't cause additional boxing, but both methods only unbox the `Long`, so accepting `long`  works the same and helps to avoid confusion and accidental boxing the future.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -928,12 +966,6 @@ private long goalToAST(long z3context, long goal) {
   /** Closing the context. */
   public void forceClose() {
     cleanupReferences();
-
-    // Force clean all ASTs, even those which were not GC'd yet.
-    // Is a no-op if phantom reference handling is not enabled.
-    for (long ast : referenceMap.values()) {
-      Native.decRef(getEnv(), ast);
-    }
```
PhilippWendler, line 936: What is the reason for deleting this?

ThomasHaas:
>> Furthermore, the memory footprint of the list should be strictly smaller than that of the previously used IdentityHashMap.

> Really?

I was expecting this response :).

> So if I did not overlook something, the list is not strictly smaller than the map, if the latter happens to be >= 75% full and the Longs are referenced anyway.

It is true that I accidentally calculated in the boxing of the longs, but those do already exist in all cases where the function is used. Nevertheless, the load factor of `IdentityHashMap` is 66.67% (it resizes if `3*size > 2*capacity`) guaranteeing that for each stored object there is at least one free cell, hence the size per object is `3n <= size/object <= 4n`.
The proposed list is always `3n` in memory size and thus smaller. It might be the case that a fully filled hashmap is identical in size to the list, but for all practical purposes I would still claim that the list is strictly smaller :).

> But I see a major additional argument: The memory consumption of the list shrinks again once fewer elements exist. IdentityHashMap does not implement any shrinking of the table, so if the program creates a lot of Z3Formula objects and frees them afterwards, the memory consumed by the table is kept occupied until the whole solver context is freed.

Good point. This also means the upper bound `size/object <= 4n` does not even hold. The maps memory consumption is unbounded.

> Of course, having add/remove be guaranteed O(1) instead of just amortized is also nice.

Yes, it performs A LOT better in extreme cases where the hashmap's operations start to approach linear time.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
ThomasHaas: This was by accident because I misinterpreted the purpose. I added it back.

However, I noticed that if the `PhantomReferences` are already enqueued, then they will never get removed from the queue after closing (until the context object itself gets collected).
Unfortunately, `ReferenceQueue.clear` does not exist and one has to poll all entries to clear them.
Should I add this or does this not really matter?

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
ThomasHaas: I agree.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
ThomasHaas: Changed everything to private.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
ThomasHaas: Yes.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -931,8 +969,16 @@ public void forceClose() {
 
     // Force clean all ASTs, even those which were not GC'd yet.
     // Is a no-op if phantom reference handling is not enabled.
-    for (long ast : referenceMap.values()) {
-      Native.decRef(getEnv(), ast);
+    if (usePhantomReferences) {
+      Z3AstReference cur = referenceListHead.next;
+      assert cur != null;
+      while (cur.next != null) {
+        Native.decRef(environment, cur.z3Ast);
```
kfriedberger, line 976: "cur = cur.next" missing?

ThomasHaas: Oh, yeah...

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
PhilippWendler: You mean those references that have been enqueued before `close()` is called? These could in principle be cleaned up, but note that there will likely also be references that are enqueued after `close()` is called, and there is nothing we can do about them, so I do not think we need to bother with the already enqueued ones.

Once the solver context becomes unreachable and is cleaned up by GC, the queue and all `PhantomReference`s are also cleaned up. So this affects only users who close the context but then keep a reference to it for a longer time.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -931,8 +969,17 @@ public void forceClose() {
 
     // Force clean all ASTs, even those which were not GC'd yet.
     // Is a no-op if phantom reference handling is not enabled.
```
PhilippWendler, line 971:
```suggestion
```
The second line of the context is now redundant because of the explicit condition.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -931,8 +969,17 @@ public void forceClose() {
...
+      Z3AstReference cur = referenceListHead.next;
+      assert cur != null;
+      while (cur.next != null) {
+        Native.decRef(environment, cur.z3Ast);
+        cur = cur.next;
+      }
+      Z3AstReference tail = cur;
+      // Bulk delete everything between head and tail
+      referenceListHead.next = tail;
+      tail.prev = referenceListHead;
```
PhilippWendler, line 982: This doesn't really change anything, right?

All the `Z3AstReference` deleted here will either be already in the `ReferenceQueue` or will be added to it in the future, unless the whole context is freed before. So in no case the reference through the list should matter.

Deleting here produces an inconsistent doubly-linked list, but this should also not matter because `cleanupReferences` should never be called again.

Not sure whether I find it cleaner to keep the deletion or to remove it.

(If it is removed, the `while` loop can be rewritten with `for`.)

ThomasHaas: > All the `Z3AstReference` deleted here will either be already in the `ReferenceQueue` or will be added to it in the future, unless the whole context is freed before. So in no case the reference through the list should matter.

No, `PhantomReferences` that are not referenced anymore are themselves eligible for garbage collection and will never get enqueued because the GC won't find them anymore!
Otherwise we would not even have to bother with storing the phantom references in the first place.

It is true though that as long as there is a single reference enqueued, the whole list is reachable through that single reference and hence cannot be GC'd early. This also means that all references will eventually get enqueued...

Then I suggest to bulk delete the list, and then(*) poll the `ReferenceQueue` until it is empty.
This will guarantee that all phantom references can be GC'd and no reference will ever get enqueued again.

> 
> Deleting here produces an inconsistent doubly-linked list, but this should also not matter because `cleanupReferences` should never be called again.
> 
It is true that`cleanupReferences` should not be called again. Even if the list was kept consistent, this would still cause double decrements of the native reference counters and hence be incorrect (this behaviour matches the previous one). If I clear the queue as suggested above, this problem will also be gone! Furthmore the list is guaranteed to be fully consistent afterwards!

(*) The ordering is important. If one clears the queue first, it may get refilled during the deletion of the list

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
ThomasHaas: See my answer to the bulk deletion point you raised further below.

ThomasHaas:
I noticed that the `Z3FormulaCreator` is full of calls to `clearReferences`, pretty much everywhere where objects get created.
How about moving the call to `clearReferences` to inside `storePhantomReference` so that all code that generates phantom references also clears them? This would make all existing calls to ´clearReferences` obsolete.

kfriedberger (approved):
I could not detect a breaking example or unit test. The code itself looks valid.
However, I am not sure whether we need/want to merge this PR.

Executing a command like
`java -cp bin:lib/java/runtime-z3/common.jar:lib/java/runtime-z3/com.microsoft.z3.jar org.sosy_lab.java_smt.example.nqueens_user_propagator.NQueens 11 2`
or
`java -cp bin:lib/java/runtime-z3/common.jar:lib/java/runtime-z3/com.microsoft.z3.jar org.sosy_lab.java_smt.example.nqueens_user_propagator.NQueens 11 3`
showed a no speedup, e.g., the example runs for about 6.0-6.5s, with and without this change.
Perhaps the speedup was related to other optimizations implemented for Z3 or user propagation.

ThomasHaas:
> I could not detect a breaking example or unit test. The code itself looks valid. However, I am not sure whether we need/want to merge this PR.
> 
> Executing a command like `java -cp bin:lib/java/runtime-z3/common.jar:lib/java/runtime-z3/com.microsoft.z3.jar org.sosy_lab.java_smt.example.nqueens_user_propagator.NQueens 11 2` or `java -cp bin:lib/java/runtime-z3/common.jar:lib/java/runtime-z3/com.microsoft.z3.jar org.sosy_lab.java_smt.example.nqueens_user_propagator.NQueens 11 3` showed a no speedup, e.g., the example runs for about 60-65s, with and without this change. Perhaps the speedup was related to other optimizations implemented for Z3 or user propagation.

Have you enabled phantom references (they are disabled by default)?
Also, I made the user propagator robust(*) against this feature by reusing existing wrappers and creating only half as many wrappers as before.
Even if I disable the optimization, Nqueens (N=13) does not run into performance problems but it still causes the reference map to reach a size of 7+ million entries, meaning it will consume over 1GB of memory until the whole context is dropped (deu to the lack of shrinking).

(*) The robustness is only for formulas reported by the user propagator itself. If the user decides to create new formulas from inside the callbacks, they can still end up with millions of phantom references and a possibility of encountering bad map performance.

kfriedberger:
First: I forgot to enable phantom references. Thanks for the hint.

Second: I copied the time numbers and forgot the decimal separator (too late in the night :-) ). My older private laptop requires about 6.0-6.5s for the example. I measured again on a bigger machine just now and get about 2s when executing NQueens of size 11 with method 2 or 3. For n=12, the big machine requires about 20s with map-stored phantom references, and also 20s with the list-based references.

Issue #360, opened 2024-03-19
Missing features and/or potential bugs (willing to fix)


leventeBajczi:
Hi!

I'm in the process (https://github.com/ftsrg/theta/pull/258) of integrating JavaSMT with our Theta model checker. By doing so, I have discovered some pain points of the integration, which I'd like to both log as a potential issue and discuss in terms of what I can do to help these get resolved. I'm more than happy to try and create a PR about the changes, but wanted to discuss first.

I know that most of these problems are easily solved using some workarounds, but I think there is value in getting them implemented / fixed.

In no particular order the missing features (keep in mind that it is entirely possible that I just missed some :) ):

1. The conversion between `Rational` and `Integer` sorts in this direction is given by the `.floor()` operation, but there is no inverse operation, despite Z3 having a `.mkInt2Real`, SMT-LIB defining a `to_real`, etc. I can see some signs of support for this operator (e.g., [here](https://github.com/sosy-lab/java-smt/blob/dd87728958851fccd7960aeb7898354a3035ee80/src/org/sosy_lab/java_smt/api/FunctionDeclarationKind.java#L92)), but I see no way of creating this formula otherwise. 

2. There is no support for rotating (either left or right) a bitvector. Some solvers (including Z3 and CVC4) have support for these operators, and for others, it can be (expensively) emulated. However, emulating it for supporting solvers is way too big of an overhead.

3. `fp.rem` is not supported. 

4. I see no way of creating array literals, therefore I cannot translate the following assertion: `(assert (= 0 (select ((as const (Array Int Int)) 0) 0)))`. 

5. I see no way of creating a floating point literal from a IEEE-754 bit pattern without first creating the bitvector literal, then using the `fromIeeeBitvector`. 

*And the (potential) bug:*

6.  Modulo creates REM operator instead of MOD for bitvectors. See relevant code [here](https://github.com/sosy-lab/java-smt/blob/dd87728958851fccd7960aeb7898354a3035ee80/src/org/sosy_lab/java_smt/solvers/z3/Z3BitvectorFormulaManager.java#L130-L136). This is potentially dangerous, because (at least) I would think that a `.modulo()` function will use the `bvsmod` operator, and not one of the `{s,u}rem` operators. Consider the following as an example for why this matters:

``` smtlib
(declare-fun a () (_ BitVec 2))
(declare-fun b () (_ BitVec 2))

(assert (not (= (bvsmod a b) (bvsrem a b))))

[...]

PR #362, opened 2024-03-23
Add rotation operations to BitvectorFormulaManager

leventeBajczi:
As discussed in https://github.com/sosy-lab/java-smt/issues/360, this PR ~adds~ enhances the rotateRight and rotateLeft operations in the bitvector formula manager. 

I added implementations for solvers that do not directly support rotation-by-bitvector, by using shift operations. Therefore, all bitvector-supporting solvers can also use rotation now.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractBitvectorFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractBitvectorFormulaManager.java
@@ -352,28 +352,6 @@ public abstract class AbstractBitvectorFormulaManager<TFormulaInfo, TType, TEnv,
     throw new UnsupportedOperationException("Solver does not support BV rotation.");
   }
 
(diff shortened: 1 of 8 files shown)


leventeBajczi:
All tests seem to pass, but Princess is very slow in the `bvRotateByBV` test. Maybe it is worth disabling that test for Princess as to not make testing take too long (and `bvIsIdenticalAfterFullRotation` tests Princess's rotation capabilities to some degree)

leventeBajczi:
Yes, even in the AppVeyor the test experienced a timeout with princess. I'll disable it.

kfriedberger:
The idea to use mod/rem for computing the shift is very good. 
I adapted your solution and applied it in 12ceedfa6a.

Princess is quite slow for BV theory, and BV-rotation seems to be beyond its limits, for now.

I will close this PR without merging.

PR #363, opened 2024-03-23
Add fp.rem operation to FloatingPointFormulaManager

leventeBajczi:
As discussed in https://github.com/sosy-lab/java-smt/issues/360, this PR adds the fp.rem operation in the floatingpoint formula manager.

I also added some light tests, as to prove that the operation behaves as expected. Let me know if I should add more.


kfriedberger:
The PR looks good. I will wait for the project-internal tests and then merge this.
Those tests run on the Gitlab-runner and are only executed for commits from project-internal developers.

PR #364, opened 2024-03-23
Add new floating point literal constructor using IEEE754 bitpattern

leventeBajczi:
As discussed in #360, currently, floating point literals cannot be directly created if the user has access to the bit pattern of the value. Therefore, this PR adds support for this using the new `makeNumber(BigInteger exponent, BigInteger mantissa, boolean signBit, FloatingPointType type)` method.

I added a test to verify that no trivial mistakes were made. Let me know if I should add more.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3FloatingPointFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3FloatingPointFormulaManager.java
@@ -83,7 +83,14 @@ class Z3FloatingPointFormulaManager
         Native.mkNumeral(
             z3context, mantissa.toString(), Native.mkBvSort(z3context, type.getMantissaSize()));
 
-    assert Native.getSort(z3context, signBv) != Native.getSort(z3context, mantBv);
+    assert Native.getBvSortSize(z3context, Native.getSort(z3context, signBv)) == 1
+        : "SignBV should be 1 bit long";
+    assert Native.getBvSortSize(z3context, Native.getSort(z3context, expoBv))
+            == type.getExponentSize()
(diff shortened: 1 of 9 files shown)


src/org/sosy_lab/java_smt/solvers/z3/Z3FloatingPointFormulaManager.java
```diff
@@ -70,6 +71,30 @@ protected Long makeNumberImpl(double pN, FloatingPointType pType, Long pRounding
...
+
+    long signBv = Native.mkBvNumeral(z3context, 1, new boolean[] {signBit});
+    long expoBv =
+        Native.mkNumeral(
+            z3context, exponent.toString(), Native.mkBvSort(z3context, type.getExponentSize()));
+    long mantBv =
+        Native.mkNumeral(
+            z3context, mantissa.toString(), Native.mkBvSort(z3context, type.getMantissaSize()));
+
+    assert Native.getBvSortSize(z3context, Native.getSort(z3context, signBv)) == 1
```
leventeBajczi, line 86: Note that currently, that test fails for Z3, because it overwrites the created constant's sort. I don't know how to solve this problem, I suspect a bug in Z3. If I remove line 82, then signBv's size remains 1, but as-is, its size is 23 (= mantissa size).

leventeBajczi: It does not matter which constructor method I use on z3's side (`string` or `bool[]`).

leventeBajczi: Nevermind, I was just missing some incRefs and decRefs.

PR #367, opened 2024-03-24
Added array literal creation to API and implementation

leventeBajczi:
As mentioned in #360, currently, array literals are not supported by the API. This PR adds two new methods to this end, one without a default value, and one with a default (else) value. I also added some tests to verify that these work normally.

### DIFF
--- a/src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java
@@ -61,6 +61,30 @@ public interface ArrayFormulaManager {
   <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
       String pName, ArrayFormulaType<TI, TE> type);
 
+  /** Create a new array constant with uninitialized values. */
+  <TI extends Formula, TE extends Formula, FTI extends FormulaType<TI>, FTE extends FormulaType<TE>>
+      ArrayFormula<TI, TE> makeArray(FTI pIndexType, FTE pElementType);
+
+  /** Create a new array constant with uninitialized values. */
+  <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
+      ArrayFormulaType<TI, TE> type);
+
+  /**
+   * Create a new array constant with uninitialized values.
+   *
+   * @param elseElem: The default value of all entries in the array.
+   */
+  <TI extends Formula, TE extends Formula, FTI extends FormulaType<TI>, FTE extends FormulaType<TE>>
+      ArrayFormula<TI, TE> makeArray(TE elseElem, FTI pIndexType, FTE pElementType);
+
+  /**
+   * Create a new array constant with values initialized to elseElem.
+   *
+   * @param elseElem: The default value of all entries in the array.
+   */
+  <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
+      TE elseElem, ArrayFormulaType<TI, TE> type);
+
   /** Make a {@link BooleanFormula} that represents the equality of two {@link ArrayFormula}. */
   <TI extends Formula, TE extends Formula> BooleanFormula equivalence(
       ArrayFormula<TI, TE> pArray1, ArrayFormula<TI, TE> pArray2);
--- a/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ArrayFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ArrayFormulaManager.java
@@ -47,6 +47,15 @@ public class CVC5ArrayFormulaManager extends AbstractArrayFormulaManager<Term, S
   }
 
   @Override
+  protected <TI extends Formula, TE extends Formula> Term internalMakeArray(
+      Term elseElem, FormulaType<TI> pIndexType, FormulaType<TE> pElementType) {
+    final ArrayFormulaType<TI, TE> arrayFormulaType =
+        FormulaType.getArrayType(pIndexType, pElementType);
+    final Sort cvc5ArrayType = toSolverType(arrayFormulaType);
+    return solver.mkConstArray(cvc5ArrayType, elseElem);
+  }
+
+  @Override
   protected Term equivalence(Term pArray1, Term pArray2) {
     return solver.mkTerm(Kind.EQUAL, pArray1, pArray2);
   }
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5ArrayFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5ArrayFormulaManager.java
@@ -8,6 +8,7 @@
 
 package org.sosy_lab.java_smt.solvers.mathsat5;
 
+import static org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_make_array_const;
 import static org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_make_array_read;
 import static org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_make_array_write;
 import static org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.msat_make_equal;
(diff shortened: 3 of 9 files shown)


src/org/sosy_lab/java_smt/solvers/z3/Z3ArrayFormulaManager.java
```diff
@@ -45,6 +46,22 @@ protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
     return getFormulaCreator().makeVariable(z3ArrayType, pName);
   }
 
+  @Override
+  protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
+      FormulaType<TI> pIndexType, FormulaType<TE> pElementType) {
+    long func =
+        getFormulaCreator()
+            .declareUFImpl(
+                "__unnamed_arr", toSolverType(pElementType), List.of(toSolverType(pIndexType)));
```
kfriedberger, line 55: Is this extra UF really necessary?
How can we create two arrays with default values, when they have distinct types? (test missing)
Is this name whitelisted as normal variable name and excluded in model generation? (test missing)

src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java
```diff
@@ -61,6 +61,30 @@ <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> store(
   <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
       String pName, ArrayFormulaType<TI, TE> type);
 
+  /** Create a new array constant with uninitialized values. */
+  <TI extends Formula, TE extends Formula, FTI extends FormulaType<TI>, FTE extends FormulaType<TE>>
+      ArrayFormula<TI, TE> makeArray(FTI pIndexType, FTE pElementType);
+
+  /** Create a new array constant with uninitialized values. */
+  <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
```
kfriedberger, line 69: Do we really need two methods for doing the same thing? Just packing two types into one array-type could be done on user-side. We might want to reduce the API at this place and only provide one method.

leventeBajczi: I took the other makeArray() methods as the basis for this API change (https://github.com/sosy-lab/java-smt/blob/7a9ec2262b0fe34b588515a43b2add10d1afe4c4/src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java#L53-L62). However, I see no other point (besides adhering to the style of existing methods) in keeping both of these, so if you agree, I can remove the one taking two types instead of a single array type.

src/org/sosy_lab/java_smt/solvers/z3/Z3ArrayFormulaManager.java
leventeBajczi: I don't see another way of creating a non-initialized array literal. I added some extra tests to verify that these do not create problems with e.g. vars with the same name. It turns out that only Z3 works well with such literals (without a default value), and only Z3, MathSAT and CVC5 work with initialized literals.

leventeBajczi:
I can see that the `SolverConcurrencyTest` seems to be failing. However, I cannot reproduce the problem locally. Can you please take a look?

baierd:
`SolverConcurrencyTest` is flaky. If its not reproducable locally or continuously in the CI you may ignore that test class.

.idea/ant.xml
kfriedberger, line 1: Please do not change basic project files.

src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java
```diff
@@ -61,6 +61,30 @@ <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> store(
   <TI extends Formula, TE extends Formula> ArrayFormula<TI, TE> makeArray(
       String pName, ArrayFormulaType<TI, TE> type);
 
+  /** Create a new array constant with uninitialized values. */
+  <TI extends Formula, TE extends Formula, FTI extends FormulaType<TI>, FTE extends FormulaType<TE>>
```
kfriedberger, line 65: I do not yet see the benefit of having uninitialized unnamed arrays in the API.
Please conosider the following usages:
- Create such an array and then use the Java instance twice. Should these be two distinct arrays or the same array? (There is no difference to having a array variable.)
- Create two distinct uninitialized arrays. Those arrays are distinct and can hold distinct or equal elements. (There is no difference to having two array variables.)

Overall, the user can simply choose a unique name and create an array directly.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ArrayFormulaManager.java
```diff
@@ -46,6 +46,15 @@ protected <TI extends Formula, TE extends Formula> Term internalMakeArray(
...
   }
 
+  @Override
+  protected <TI extends Formula, TE extends Formula> Term internalMakeArray(
+      Term elseElem, FormulaType<TI> pIndexType, FormulaType<TE> pElementType) {
+    final ArrayFormulaType<TI, TE> arrayFormulaType =
+        FormulaType.getArrayType(pIndexType, pElementType);
+    final Sort cvc5ArrayType = toSolverType(arrayFormulaType);
+    return solver.mkConstArray(cvc5ArrayType, elseElem);
+  }
```
kfriedberger, line 56: I took this part of the code and merged it to master.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5ArrayFormulaManager.java
```diff
@@ -48,6 +49,15 @@ protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
...
   }
 
+  @Override
+  protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
+      Long elseElem, FormulaType<TI> pIndexType, FormulaType<TE> pElementType) {
+    final ArrayFormulaType<TI, TE> arrayFormulaType =
+        FormulaType.getArrayType(pIndexType, pElementType);
+    final Long mathsatArrayType = toSolverType(arrayFormulaType);
+    return msat_make_array_const(mathsatEnv, mathsatArrayType, elseElem);
+  }
```
kfriedberger, line 59: I took this part of the code and merged it to master.

src/org/sosy_lab/java_smt/solvers/z3/Z3ArrayFormulaManager.java
```diff
@@ -45,6 +46,22 @@ protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
...
+    return Native.mkAsArray(z3context, func);
+  }
+
+  @Override
+  protected <TI extends Formula, TE extends Formula> Long internalMakeArray(
+      Long elseElem, FormulaType<TI> pIndexType, FormulaType<TE> pElementType) {
+    return Native.mkConstArray(z3context, toSolverType(pIndexType), elseElem);
+  }
```
kfriedberger, line 63: I took this part of the code and merged it to master.

src/org/sosy_lab/java_smt/test/ArrayFormulaManagerTest.java
```diff
@@ -180,4 +180,155 @@ public void testFloatIndexFloatValue() throws SolverException, InterruptedExcept
...
+    // select(store(arr, i, j)) == j
+    assertThatFormula(bvmgr.equal(num4, amgr.select(amgr.store(arr, num2, num4), num2)))
+        .isTautological();
+    // same, but checking whether unnamed arrays work well (i.e., not global)
+    assertThatFormula(bvmgr.equal(num5, amgr.select(amgr.store(arr, num2, num5), num2)))
+        .isTautological();
+    // init array cannot read any value besides default
+    assertThatFormula(bvmgr.equal(numM1, amgr.select(arr, num2))).isUnsatisfiable();
+    assertThatFormula(bvmgr.equal(num0, amgr.select(arr, num2))).isTautological();
+  }
```
kfriedberger, line 333: I took this part of these tests as basis for further tests that are merged to master.

.idea/ant.xml
leventeBajczi: ah sorry, I didn't pay attention when adding files to the commit (and IntelliJ is adamant on changing these all the time..)

src/org/sosy_lab/java_smt/api/ArrayFormulaManager.java
leventeBajczi: That's probably right.

leventeBajczi:
I believe you're right that uninitialized arrays do not make a lot of sense. I also see that for the initialized arrays you merged the modifications to master manually - I'm closing this PR. Thanks for the help!

Issue #381, opened 2024-06-25
`interpolation can only be done over previously asserted formulas` when re-adding constraints


leventeBajczi:
Following my discussion with @baierd, I'm providing a (somewhat) minimal working example for the bug where re-adding the same constraint twice makes it impossible to query interpolants.

**Setup**: add one constraint (`A`), add a second constraint (`B1`) making the system UNSAT, then re-add the second constraint (`B2`), and query interpolants between either `A` and `B1`, or `A` and `B2`. 
**Experienced behavior**: Some solvers report `interpolation can only be done over previously asserted formulas`.
**Expected behavior**: Re-adding constraints should not interfere with the ability to calculate interpolants.

The affected solvers (What I think are considered bugs are in **bold**):

|        	| OpenSMT 	| Yices2 	| Z3     	| MathSAT5 	| SMTInterpol 	| Princess 	| Boolector 	| CVC5     	| CVC4   	|
|--------	|---------	|--------	|--------	|----------	|-------------	|----------	|-----------	|----------	|--------	|
| test 1 	| _error_ 	| no itp 	| no itp 	| **FAIL** 	| **FAIL**    	| **FAIL** 	| no itp    	| **FAIL** 	| no itp 	|
| test 2 	| _error_ 	| no itp 	| no itp 	| PASS     	| PASS        	| PASS     	| no itp    	| PASS     	| no itp 	|
| test 3 	| _error_ 	| no itp 	| no itp 	| PASS     	| **FAIL**    	| **FAIL** 	| no itp    	| PASS     	| no itp 	|

I would separately emphasize the third test's success with MATHSAT5 and CVC5, which makes it even harder to understand what's happening.

The reproduction test (junit5):

``` java
    public static Stream<Arguments> data() {
        return Arrays.stream(Solvers.values()).map(Arguments::of);
    }

    static SolverContext createSolver(Solvers solver) throws InvalidConfigurationException {
        Configuration config = Configuration.defaultConfiguration();
        LogManager logger = BasicLogManager.create(config);
        ShutdownManager shutdown = ShutdownManager.create();

        SolverContext context = SolverContextFactory.createSolverContext(
                config, logger, shutdown.getNotifier(), solver);
        return context;
    }

    @ParameterizedTest
    @MethodSource("data")
    void first(Solvers solver) throws InterruptedException, SolverException, InvalidConfigurationException {
        SolverContext context = createSolver(solver);
[...]

PR #382, opened 2024-06-29
#381 improve interpolation behaviour

kfriedberger:
see #381 for details

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
@@ -2,7 +2,7 @@
 // an API wrapper for a collection of SMT solvers:
 // https://github.com/sosy-lab/java-smt
 //
-// SPDX-FileCopyrightText: 2022 Dirk Beyer <https://www.sosy-lab.org>
+// SPDX-FileCopyrightText: 2024 Dirk Beyer <https://www.sosy-lab.org>
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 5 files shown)


src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -1179,4 +1179,70 @@ public <T> void testInvalidToken() throws InterruptedException, SolverException
     assertThrows(
         IllegalArgumentException.class, () -> stack.getInterpolant(ImmutableList.of((T) p3)));
   }
+
+  @Test
+  @SuppressWarnings("unused")
```
baierd, line 1184: The unused suppression's should be removed ;D

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -103,17 +113,20 @@ protected void popImpl() {
...
+  @CanIgnoreReturnValue
+  protected String addConstraint0(BooleanFormula pF) {
     Preconditions.checkState(!closed);
     setChanged();
     Term exp = creator.extractInfo(pF);
     if (incremental) {
       solver.assertFormula(exp);
     }
-    return null;
+    String id = "ID_" + ID_GENERATOR.getFreshId();
```
baierd, line 127: I totally get what you are doing here. But we should create a new class (InterpolationPoint?) that wraps this ID to make the intention clear for user and future developers.

baierd (changes requested):
This fixes the issue nicely. And besides the 2 small change requests this can be merged with no issues.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
kfriedberger: InterpolationPoint is a quite special use-case. Some solvers also use those references for unsat-cores or model generation. We can think about it independently of this issue. Maybe name the reference to something like FormulaBacktrackPoint.

Issue #308, opened 2023-06-04
Princess: match error in model evaluation (found in CPAchecker benchmark)


kfriedberger:
LOG:
```
scripts/cpa.sh -heap 13000M -noout -benchmark -stack 50M -setprop cpa.predicate.memoryAllocationsAlwaysSucceed=true -setprop counterexample.export.allowImpreciseCounterexamples=true -setprop cpa.predicate.encodeFloatAs=INTEGER -bmc -setprop solver.solver=PRINCESS -setprop cpa.predicate.encodeBitvectorAs=INTEGER -setprop cpa.loopbound.maxLoopIterations=10 -timelimit 60s -stats -spec ../sv-benchmarks/c/properties/unreach-call.prp -32 ../sv-benchmarks/c/nla-digbench-scaling/freire1_valuebound1.c


--------------------------------------------------------------------------------


Running CPAchecker with Java heap of size 13000M.
Running CPAchecker with Java stack of size 50M.
Language C detected and set for analysis (CPAMain.detectFrontendLanguageIfNecessary, INFO)

Using the following resource limits: CPU-time limit of 60s (ResourceLimitChecker.fromConfiguration, INFO)

CPAchecker 2.2.1-svn-43814M / bmc (OpenJDK 64-Bit Server VM 17.0.7) started (CPAchecker.run, INFO)

Parsing CFA from file(s) "../sv-benchmarks/c/nla-digbench-scaling/freire1_valuebound1.c" (CPAchecker.parse, INFO)

Using unsound approximation of ints with unbounded integers and floats with unbounded integers for encoding program semantics. (PredicateCPA:FormulaManagerView.<init>, WARNING)

Using predicate analysis with Princess 2022-11-03. (PredicateCPA:PredicateCPA.<init>, INFO)

Using unsound approximation of ints with unbounded integers and floats with unbounded integers for encoding program semantics. (AssumptionStorageCPA:FormulaManagerView.<init>, WARNING)

Starting analysis ... (CPAchecker.runAlgorithm, INFO)

Creating formula for program (AbstractBMCAlgorithm.run, INFO)

Starting satisfiability check... (AbstractBMCAlgorithm.boundedModelCheck, INFO)

Error found, creating error path (AbstractBMCAlgorithm.analyzeCounterexample0, INFO)

Exception in thread "main" scala.MatchError: _0 (of class ap.parser.ISortedVariable)
[...]

PR #391, opened 2024-09-13
Add support for Strings and Rationals to the Princess backend

daniel-raffler:
Hello everyone,
the goal of this pull request is to add two more theories for strings and rational numbers to the Princess backend. It is a continuation of the work that went into https://github.com/sosy-lab/java-smt/pull/257. Some of the issues that blocked us from merging it back then have since been resolved, and at least the rational formula manager is actually starting to look pretty good now. Here is a comparison between Princess and some other solvers when running BMC on the `ReachSafety-Floats` task with option `cpa.predicate.encodeFloatAs=RATIONAL` set to test the rational formula manager:

Screenshot from 2024-10-17 14-14-36

It's a bit harder to evaluate how well the String manager is working now. The Ostrich team has been very helpful and they quickly fixed a couple of bugs that we've reported. However, since CPAchecker doesn't use String theory there is no easy way of trying it out.

## Linked issues:

Princess (Rationals)
- https://github.com/uuverifiers/princess/issues/7
- https://github.com/uuverifiers/princess/issues/8
- https://github.com/uuverifiers/princess/issues/11
- https://github.com/uuverifiers/princess/issues/12
- https://github.com/uuverifiers/princess/issues/13
- https://github.com/uuverifiers/princess/issues/14
- https://github.com/uuverifiers/princess/issues/15
- https://github.com/uuverifiers/princess/issues/16

Ostrich (Strings)
- https://github.com/uuverifiers/princess/issues/9
- https://github.com/uuverifiers/ostrich/issues/87
- https://github.com/uuverifiers/ostrich/issues/88
- https://github.com/uuverifiers/ostrich/issues/89

I've added (partial) workarounds for the two Princess issues, but some work may still be needed. For Ostrich the biggest limitation is that the solver can only handle constraints where (at least) one of the sides is a singleton (= constant) string. Something like "a str.<= b" where both `a` and `b` are variables is already impossible. This is a known issue and they're planning a redesign for the solver that will allow it to handle such constraints. In the mean time we might want to add some additional error messages to make sure that the user knows why their program is failing.


src/org/sosy_lab/java_smt/test/SolverTheoriesTest.java
```diff
@@ -767,6 +767,33 @@ public void testMakeBitVectorArray() {
     }
   }
 
+  @Test
+  public void testNestedIntegerArray() {
+    requireArrays();
+    requireIntegers();
+
+    IntegerFormula _i = imgr.makeVariable("i");
```
baierd, line 775: `_i` is not a valid variable name according to our style guide.

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Override
+  protected IExpression makeVariableImpl(String varName) {
+    return getFormulaCreator().makeVariable(PrincessEnvironment.FRACTION_SORT, varName);
+  }
+
+  @Override
+  protected IExpression floor(IExpression number) {
+    throw new AssertionError("floor is not supported in Princess");
```
baierd, line 112: This looks like a `UnsupportedOperationException`.

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Override
+  protected IExpression multiply(IExpression number1, IExpression number2) {
+    FormulaType<?> sort1 = getFormulaCreator().getFormulaType(number1);
+    FormulaType<?> sort2 = getFormulaCreator().getFormulaType(number1);
+
+    IExpression result = PrincessEnvironment.rationalTheory.mul((ITerm) number1, (ITerm) number2);
+
+    if (result instanceof IIntLit && ((IIntLit) result).value().equals(IdealInt.apply(0))) {
```
baierd, line 122: A comment for explaining this line would be nice.

src/org/sosy_lab/java_smt/solvers/princess/PrincessStringFormulaManager.java
```diff
@@ -0,0 +1,309 @@
...
+  PrincessStringFormulaManager(PrincessFormulaCreator pCreator) {
+    super(pCreator);
+  }
+
+  /**
+   * Tries to parse an escaped unicode character
+   *
+   * <p>Returns the original String if parsing is not possible.
+   */
+  private static String literalOrSkip(String pToken) {
```
baierd, line 38: Do we test this sufficiently?

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -155,6 +163,11 @@ private ImmutableList<ValueAssignment> getAssignments(
     // then handle assignments for non-array cases.
     // we expect exactly one assignment per model entry.
 
+    // FIXME: Princess may return multiply mappings for UFs if the arguments are rational, f.ex
```
baierd, line 166: Would you be so kind and open an issue for this so that we don't forget?

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -240,14 +253,76 @@ public String toString() {
   @Override
   public void close() {}
 
+  /** Tries to determine the Sort of a Term. */
+  private Sort getSort(IExpression pTerm) {
+    // Just using sortof() won't work as Princess may have simplified the original term
+    // FIXME: This may also affect other parts of the code that use sortof()
```
baierd, line 259: Would you be so kind and open an issue for this so that we don't forget?
Also, is there an option to disable this maybe?

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -240,14 +253,76 @@ public String toString() {
...
+    if (pTerm instanceof ITimes) {
+      ITimes times = (ITimes) pTerm;
+      return getSort(times.subterm());
+    } else if (pTerm instanceof IPlus) {
+      IPlus plus = (IPlus) pTerm;
+      return getSort(plus.apply(0));
+    } else if (pTerm instanceof IFormula) {
+      return creator.getBoolType();
+    } else {
+      // TODO: Do we need more cases?
```
baierd, line 269: It would be good to look into this before merging.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -240,14 +253,76 @@ public String toString() {
...
+    }
+    return pTerm;
+  }
+
   @Override
-  protected IExpression evalImpl(IExpression formula) {
-    IExpression evaluation = evaluate(formula);
-    if (evaluation == null) {
-      // fallback: try to simplify the query and evaluate again.
-      evaluation = evaluate(creator.getEnv().simplify(formula));
```
baierd, line 248: What happens when `simplify()` is called for rationals?

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
```diff
@@ -508,32 +540,69 @@ private static String getName(IExpression var) {
   }
 
   static FormulaType<?> getFormulaType(IExpression pFormula) {
+    // TODO: We could use Sort.sortof() here, but it sometimes returns `integer` even though the
```
baierd, line 543: We should extract this into an issue so that we don't forget.

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
```diff
@@ -661,6 +730,15 @@ static <T> Seq<T> toSeq(List<T> list) {
...
+    return PrincessEnvironment.toSeq(
+        exprs.stream().map(e -> (ITerm) e).collect(Collectors.toList()));
+  }
+
+  static Seq<ITerm> toITermSeq(IExpression... exprs) {
+    return toITermSeq(Arrays.asList(exprs));
+  }
+
   IExpression simplify(IExpression f) {
     // TODO this method is not tested, check it!
```
baierd, line 743: Native solver tests seem appropriate considering there is some assumptions that we put on the behavior of the solver.

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
```diff
@@ -104,6 +112,30 @@ class PrincessEnvironment {
 
   public static final Sort BOOL_SORT = Sort$.MODULE$.Bool();
   public static final Sort INTEGER_SORT = Sort.Integer$.MODULE$;
+  public static final Sort NAT_SORT = Sort.Nat$.MODULE$;
+
+  static final StringTheory stringTheory =
+      new OstrichStringTheory(
+          toSeq(new ArrayList<>()),
+          new OFlags(
+              OFlags.$lessinit$greater$default$1(),
```
baierd, line 121: A comment explaining these would be nice.

daniel-raffler:
> Looks like this was not easy. Good job!

Thanks!

> Also, is it possible to use Ostrich without Princess in theory?

No, unforunately not. Ostrich does include a script to run it directly from the command line, however, internally this will still just call Princess.

From our perspective it's probably best to see Ostrich as the String Theory module of Princess.

.idea/JavaSMT.iml
```diff
@@ -340,6 +340,24 @@ SPDX-License-Identifier: Apache-2.0
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="module-library">
+      <library>
+        <CLASSES>
+          <root url="jar://$MODULE_DIR$/lib/java/runtime-princess/ostrich_2.13.jar!/" />
```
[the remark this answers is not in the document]
daniel-raffler: Yes, those are both dependencies of `ostrich_2.13.jar` and we never call them directly. Here is the section from the `ivy.xml`:
```
<dependencies>
	<dependency org="org.scala-lang" name="scala-library" rev="2.13.8" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
	<dependency org="io.github.uuverifiers" name="ostrich-ecma2020-parser_2.13" rev="1.3" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
	<dependency org="io.github.uuverifiers" name="princess_2.13" rev="2023-06-19" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
	<dependency org="org.sat4j" name="org.sat4j.core" rev="2.3.1" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
	<dependency org="org.scalacheck" name="scalacheck_2.13" rev="1.14.0" force="true" conf="test->runtime(*),master(*)"/>
	<dependency org="dk.brics.automaton" name="automaton" rev="1.11-8" force="true" conf="compile->compile(*),master(*);runtime->runtime(*)"/>
</dependencies>
```
In fact we may want to remove `automaton.jar` as well as it is just another dependency?

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
daniel-raffler: Do you mean `simplify` here?

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
daniel-raffler: I've opened a new Princess issue for this: https://github.com/uuverifiers/princess/issues/16

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
daniel-raffler: This is linked to https://github.com/sosy-lab/java-smt/pull/391#discussion_r1803154040. Once the Princess issue has been fixed we can simply use `Sort.sortOf`.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
daniel-raffler: I've given it a try now, and unfortunately `simplify()` doesn't seem to solve the issue. The entire code here is meant to work around https://github.com/uuverifiers/princess/issues/7 and https://github.com/uuverifiers/princess/issues/8 in Princess and we might be able to simplify it again once these bugs have been closed.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
daniel-raffler: I've had another look at this and couldn't find any problems with the model. This was more of a note to myself while working on `evalImpl` where we do have to fix at least some of the values. However, this does not seem to be a problem for the model and I've now removed the note.

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
daniel-raffler: Fixed!

src/org/sosy_lab/java_smt/test/RationalFormulaManagerTest.java
```diff
@@ -37,11 +35,9 @@ public class RationalFormulaManagerTest extends SolverBasedTest0.ParameterizedSo
   @Test
   public void rationalToIntTest() throws SolverException, InterruptedException {
     requireRationals();
-    assume()
-        .withMessage("Solver %s does not support floor operation", solverToUse())
-        .that(solverToUse())
-        .isNotEqualTo(Solvers.OPENSMT);
+    requireRationalFloor();
```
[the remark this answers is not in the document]
daniel-raffler: Yes, it should be fixed now :)

daniel-raffler:
Princess/Ostrich has released new versions of their binaries that include the latest bug fixes for this branch.

@kfriedberger Could you handle uploading the files to our repository? We'll need version `2024-11-08` for Princess ([maven](https://central.sonatype.com/artifact/io.github.uuverifiers/princess_2.13/dependencies)) and `1.4.1` for Ostrich ([maven](https://central.sonatype.com/artifact/io.github.uuverifiers/ostrich_2.13)). I believe some of the dependencies will also have to be updated and uploaded to our mirror.

daniel-raffler:
@baierd Could you have another look at this branch to see if there are any remaining issues?

kfriedberger:
Hi @daniel-raffler .
I uploaded `Princess 2024-11-08` and `Ostrich 1.4.1` to the Sosy-Lab Ivy repository.

*PS:*
You can also do the Ivy publication on your own. The requirement is a Sosy-Lab SVN account with write-access to Ivy repository.
The SVN repository is at https://svn.sosy-lab.org/software/ivy/ and the command lines are documented in `Libraries.txt`.
Please note that that repository must be used in "append-only" style, i.e. you must never ever modify an existing library or XML file.

build/ivysettings.xml
```diff
@@ -15,14 +15,20 @@ SPDX-License-Identifier: Apache-2.0
...
+        <chain name="Repos">
+            <url name="Sosy-Lab">
+                <ivy pattern="${ivy.repo.url}/[organisation]/[module]/ivy-[revision].xml"/>
+                <artifact pattern="${ivy.repo.url}/[organisation]/[module]/[artifact]-[revision](-[classifier]).[ext]"/>
+            </url>
+            <url name="Maven-Repo" m2compatible="true">
+                <ivy pattern="https://repo1.maven.org/maven2/[organisation]/[module]/ivy-[revision].xml"/>
+                <artifact pattern="https://repo1.maven.org/maven2/[organisation]/[module]/[revision]/[artifact]-[revision].[ext]"/>
+            </url>
+        </chain>
```
kfriedberger, line 31: Do we need this change or is this just for development?

src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
```diff
@@ -46,6 +47,15 @@ abstract class PrincessAbstractProver<E> extends AbstractProverWithAllSat<E> {
...
 
+  /**
+   * Values returned by {@link Model#evaluate(Formula)}.
+   *
+   * <p>We need to record these to make sure that the values returned by the evaluator are
+   * consistant. Calling {@link #isUnsat()} will reset this list as the underlying model has been
+   * updated.
+   */
+  protected final List<IFormula> evaluatedTerms = new ArrayList<>();
+
```
kfriedberger, line 58: This might be a bigger issue for Princess. 
Can we add a plain test for Princess (without this code) to see where Princess fails to be consistant?
Can we ask the authors to provide a consistant model?

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -237,29 +244,72 @@ public String toString() {
...
-      return out.isEmpty() ? null : out.get();
-    } else if (formula instanceof IFormula) {
-      Option<IExpression> out = model.evalExpression(formula);
-      return out.isEmpty() ? null : out.get();
+  @Override
+  protected @Nullable IExpression evalImpl(IExpression expr) {
+    // Extending the partial model does not seem to work in Princess if the formula uses rational
+    // variables. To work around this issue we (temporarily) add the formula to the assertion
+    // stack and then repeat the sat-check to get the value.
+    if (expr instanceof ITerm) {
```
kfriedberger, line 287: How expensive is this workaround? 

Can we ask the Princess developers for a proper evaluation? I would assume that Princess could implement this internally, including caching for repeated calls, memoization of assignments comparing across different evaluations, etc.

build/ivysettings.xml
daniel-raffler: No, it was for development only. Once everything is in our repository it can be removed.

daniel-raffler:
> Hi @daniel-raffler . I uploaded `Princess 2024-11-08` and `Ostrich 1.4.1` to the Sosy-Lab Ivy repository.

Great work! Thanks for your help.

> **PS:** You can also do the Ivy publication on your own. The requirement is a Sosy-Lab SVN account with write-access to the SVN repository at https://svn.sosy-lab.org/software/ivy/ . The command lines are documented in `Libraries.txt`. Please note that that repository must be used in "append-only" style, i.e. you must never ever (*) modify or remove an existing library or XML file.
> 
> (*) except in very rare and really critical cases 😄

Thanks for the tip! I'll have to look into this and see if I actually have write-access to the SVN repository.

build/ivysettings.xml
daniel-raffler: I've restored the original version now.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
daniel-raffler: > How expensive is this workaround?

I've just run bmc on the `sv-comp` task set as a benchmark:

Screenshot From 2024-11-17 17-31-55
The brown line uses Princess with the `main`, and the green(?) line is for the updated version from this branch. Since we're not using floating point or string formulas, most of the difference between the two should be down to the changes in `evalImpl` (and possibly some changes in Princess itself). Note that the old version (from `main`) actually crashes on quite a number of the tasks due to https://github.com/sosy-lab/java-smt/issues/308. The workaround has the side-effect of also fixing this issue. If we filter for just the tasks that return the correct result with the Princess version from `main` the graph looks like this:

Screenshot From 2024-11-17 17-32-45
From those results it doesn't look like the work-around is adding too much overhead, especially as the problem size increases. It might still be worth looking into other benchmarks that don't use `bmc` and see if there is any significant difference there.

> Can we ask the Princess developers for a proper evaluation? I would assume that Princess could implement this internally, including caching for repeated calls, memoization of assignments comparing across different evaluations, etc.

That would of course be ideal. We already have two Princess tickets (https://github.com/uuverifiers/princess/issues/7 and https://github.com/uuverifiers/princess/issues/8) about issues with the evaluation of rational formulas, and another one (https://github.com/uuverifiers/princess/issues/17) for the crashes on integer formulas that I mentioned earlier (see #308). The first two are already quite old and I don't know what the current status is. Maybe we could just ask for an update? For the last one I was told that they will need more time for a fix as the issues doesn't seem to be entirely trivial. Because of that I would prefer to merge this branch with the workaround now and then come back later once the issues have been fixed in Princess and remove the workaround again.

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
kfriedberger: see next line 😄

src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
daniel-raffler: > This might be a bigger issue for Princess. Can we add a plain test for Princess (without this code) to see where Princess fails to be consistant? Can we ask the authors to provide a consistant model?

I'm afraid in this case the problem here is really with the work-around, and not with Princess. We effectively emulate `(get-values (<expr>))` with a sequence of commands:
```
(push 1)
(assert (= <new_variable> <expr>))
(check-sat)
(get-model) 
// Now get the value from the entry for <new_variable>
(pop 1)
```
The problem is than the sat check causes the solver to forget the old model, and because of this it can choose a different assignment for the variables. For instance, if the only assertion on the stack is `x < y` then calling `evaluate(x)` twice may return different values as the solver simply chooses a different model each time. You can see this by replacing  the code in `PrincessAbstractProver.getEvaluatedTerms()` with `ImmutableList.of()`. This should cause tests in `ModelTest` and `SolverAllSatTest` to fail as the model is not kept consistent across multiply calls to `evaluate()`.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
```diff
@@ -281,35 +288,40 @@ private ITerm simplifyRational(ITerm pTerm) {
 
   @Override
   protected @Nullable IExpression evalImpl(IExpression expr) {
+    // TODO creating our own utility variables might eb unexpected from the user.
+    //   We might need to exclude such variables in models and formula traversal.
```
daniel-raffler, line 292: This should be fine as Princess keeps symbols "per stack level". Once the level is popped at the end of this method the symbol is deleted and won't show up in models in the future.

src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
kfriedberger: The implementation of `PrincessAbstractProver.isUnsat()` should not depend on earlier model evaluations, because it can produce a completely different model. The current implementation seems to work well, even if its runtime is not optimal. We can improve this in the future.

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
kfriedberger: can be done after merging

src/org/sosy_lab/java_smt/solvers/princess/PrincessModel.java
kfriedberger: Yes, good catch. I would still prefer to not need to use any additional symbols.

src/org/sosy_lab/java_smt/solvers/princess/PrincessStringFormulaManager.java
```diff
@@ -0,0 +1,312 @@
...
+        // Try to parse the escape sequence to replace it with its 16bit Unicode character
+        // If parsing fails just keep it in the String
+        String nextToken = literalOrSkip(value);
+
+        // Collect the prefix and the (possibly) translated escape sequence
+        builder.append(prefix).append(nextToken);
+      }
+    }
+    return builder.toString();
+  }
```
kfriedberger, line 125: Just a comment, not blocking the merge:
Does escaping and unescaping with Strings align with all operations like String comparison and String inclusions? This feature seems to be quite complex.

kfriedberger (approved):
This PR looks good to me.
There are some bugs or missing features in Princess.
We can improve this in the future.

@daniel-raffler Feel free to merge this PR.

src/org/sosy_lab/java_smt/solvers/princess/PrincessStringFormulaManager.java
daniel-raffler: Hmm, that's a good question..

I wrote this method since many/most of the tests in `StringFormulaManagerTest` use (SMTLIB) escaped Unicode, which Princess does not seem  to support. It should be fine for that purpose, but I'm not sure how well escaping/unescaping will work in general. This shouldn't be a problem for this pull request as we're currently only unescaping input Strings, but don't try to escape the output.

However, we might still want to have a closer look at this. To be honest I'm not even quite sure what the format of `value` in `StringFormulaManager.makeString(String value)` is supposed to be. Several of the tests use SMTLIB escape sequences here, but the argument is really just a regular Java String. Do we want allow Unicode characters in it or not? When I try `makeString("\\u{39E}")` different solver return different results: CVC5 and Princess recognize the escape sequence and treat it as `"Ξ"`, whereas CVC4 and Z3 do not and simply return `\u{39E}` again. On the other hand `makeString("Ξ")` will throw and exception for CVC4/5 (because of the invalid character), however it passes for Princess, and Z3 will read the String as `\u{ce}\u{9e}`, which is the UTF-8 encoding of `"Ξ"` and doesn't seem quite right either.

Maybe it would be best to figure this out as part of another issue and then come up with a common definition? In the mean time I may also ask the Ostrich developers to add support for escape Unicode characters as they are part of the SMTLIB standard.

kfriedberger: I opened a new issue for that: https://github.com/sosy-lab/java-smt/issues/412

Feel free to add more details there.

Issue #380, opened 2024-06-24
support for Mac with M series chips


baoluomeng:
Curious about the timeline to support Mac with Apple M1,2,3?

kfriedberger:
We currently do not have Apple hardware for development and testing. Cross-compilation for Apple might not work. That operating system is special. 💰 

If an SMT solver provides native libraries on its own, such as Z3 maybe, we can include it. 
Otherwise, you can use Java-based SMT solvers like SMTInterpol on any  platform.

baoluomeng:
The latest Z3 has native support for arm64 Mac: https://github.com/Z3Prover/z3/releases/tag/z3-4.13.0. I downloaded the dylib files and plugged them into JavaSMT on my arm64 Mac. It works fine so far. But I would prefer a centralized approach like JavaSMT so that I don't have to tell my teammates how to set it up. Thanks.

ThomasHaas:
See this long-standing [discussion](https://github.com/sosy-lab/java-smt/issues/226). The issue seems to be somewhat orthogonal to MacOS as it applies to all ARM machines.

hernanponcedeleon:
@kfriedberger since z3 already provides native libraries for arm, would it be possible to make these available via maven?

kfriedberger:
Hi @hernanponcedeleon .

The topic appeared often enough to make me really think about it. :smile: 
I will try to find a solution within the next week, and report here.

### Step 1: Define a goal.

Users want X64 and ARM64 binaries for Z3. 

### Step 2: Make a plan.

Therefore we need: 
- publication of binaries into Ivy and Maven:
  - find a future-proof artifact pattern for naming (avoid name collision from X64 and ARM64)
  - update publication scripts
- download via Ivy and Maven:
  - test download,
  - test execution on different architectures (I might even have a ARM64 device available, but no Apple device)

### Step 3: Do it.  ... Coming soon.

hernanponcedeleon:
I am pretty sure @ThomasHaas would be willing to help with the testing 😉

ThomasHaas:
Sure, I can test it on my machine.

kfriedberger:
### Step 3: Collect data and knowhow

#### Z3

Z3 is available in x64 and arm64 version for Linux, Windows, and OSX. This is impressive. :+1: 
Linux version went wrong in last few releases, i.e., until 4.13.0, see https://github.com/Z3Prover/z3/issues/7201 :lady_beetle: 
[...]

Issue #384, opened 2024-07-07
Improve CI coverage for more systems like MacOS and Windows


kfriedberger:
Our AppVeyor CI is good and runs fine.
However, it has limited resources and does not provide more architectures, such as ARM64 for both, Windows and MacOS.
We could take a look at alternatives, without costs for open-source projects.

kfriedberger:
As a first step, we improved the support for Linux with additional configurations for arm64 architecture in #442 .

anaslari23:
I have implemented a new GitHub Actions CI workflow to provide the requested coverage for macOS and Windows runners.

Summary of work:

Added 
.github/workflows/ci.yml
 with a matrix for Ubuntu, macOS, and Windows.
Configured testing across Java 11, 17, and 21.
Resolved Windows-specific native library loading issues by ensuring .dll files are correctly placed during the build.
Included automated JUnit report archiving and JVM crash log capturing.
I've opened a Pull Request with these changes here: #577 .

This should centralize the CI process and ensure no regressions on non-Linux platforms moving forward.

kfriedberger:
With the merge of #577, we got a working (initial) [GitHub-CI pipeline](https://github.com/sosy-lab/java-smt/actions/workflows/ci.yml) for all supported platforms:
- Linux, Windows, and macOS ✅ 
- on x64 and arm64 platform ✅  

We currently only run the **smaller unit-tests** on GitHub-CI, due to possible (still unknown) limitations on computational resources, such as minutes of runtime. We need to figure out the exact limits, maybe 2000 min per month, or maybe more for open-source projects. It seems to be sufficient for now.

We also have:
- [GitLab-CI](https://gitlab.com/sosy-lab/software/java-smt/-/pipelines/), where more detailed and the longer running tests are executed on Linux. This service is used in other projects of SoSy-Lab as well, and its configuration is based on a common configuration file.
- [AppVeyor](https://ci.appveyor.com/project/sosy-lab/java-smt/), where unit tests are executed on x64-Windows only. Maybe we can remove or decommision this service in the future.

Lets close this issue.

PR #395, opened 2024-09-17
#380 support multiple architectures for solver binaries z3

kfriedberger:
This branch / pull request exists for tracking some steps towards supporting multiple architectures. I will use Z3 as example, because it was requested several times. Some explanation can be found in #380 .

I will not yet commit binaries to any public archive (Ivy or Maven) until we can make sure that the new approach is backwards-compatible (e.g. default architecture x86 can be used as before) and we have a maintainable code base (including scripts and documentation for publication and testing).


kfriedberger:
This PR is currently blocked by https://github.com/sosy-lab/java-common-lib/pull/43

kfriedberger:
The current state seems to work as expected. I tested:
- publishing solver Z3 to Ivy and Maven, including x64 and arm64 for several operating systems (/)
- downloading from Ivy is backwards-compatible for Z3 (e.g., just increment the version of Z3). This works without changing the ivy-settings. We still nedd to find a way for publishing JavaSMT itself with a reasonable configuration.
- downloading from Maven requires more smaller changes. For example, as shown in this change: https://github.com/sosy-lab/java-smt/pull/395/commits/958dbc39b9498469dfe88b6a468b0bb188cabe03

Z3 is already published in Maven and Ivy, and works well with JavaSMT 5.0.1.
Feel free to test and give feedback.

As I do not have an ARM-based Apple machine available, maybe @ThomasHaas and @hernanponcedeleon can test whether it works as expected. Thanks.

kfriedberger:
If there is positive feedback, I will also publish Z3 in its latest version [4.13.2](https://github.com/Z3Prover/z3/releases) to Iva and Maven.

ThomasHaas:
It seems to work on ARM-based Macs now. Good job :)

kfriedberger:
Z3 was published to Ant/Ivy and Maven in versions 4.13.2 and 4.13.3.

There are some smaller issues that are currently blocking this PR from merging:
- JavaSMT uses outdated CI images for Linux, e.g., Z3 has removed support for Ubuntu 20.* and requires newwer GLIBC version. We will update the CI in another branch, before merging this PR.
- Z3 has a smaller regression when parsing formulas (due to a new feature that allows to compare formulas of incompatible types like Int/Bool): 
  https://github.com/Z3Prover/z3/issues/7419
  (our fix from 8e2f3b0b81c83f0bce649b06315536f0ac8c3caf might be changed/reverted in the future, if Z3 provides a bugfix)
- Z3 DLL on Windows x64 crashes:
  https://github.com/Z3Prover/z3/issues/7420

kfriedberger:
This branch is currently blocked by or waiting for:
- Z3 for Windows is broken on a default WIndows installation due to unexpected changes in msvcp140.dll. We need to wait for a new release of Z3 where this is fixed.
- MathSAT5 was compiled for newer GLIBC (Ubuntu 22.04) and does currently not work on Ubuntu 20.04 or older. This behaviour is similar to Z3, where support for Ubuntu 20.04 was dropped some weeks ago. However, we might want to re-compile the bindings with an older GLIBC.
- Gitlab-CI for ARM64 might be useful here, too.

kfriedberger:
I would like to have a preliminary review for these changes.

The following parts are work in progress and could be solved later:
- CI for ARM, for Linux, Windows, and also MacOS in general
- support more solvers on more architectures

doc/Developers-How-to-Release-into-Ivy.md
```diff
@@ -46,18 +46,35 @@ there are scripts for publishing available at the root of the [Ivy Repository](h
...
 ```
-ant publish-z3 -Dz3.path=$Z3_DIR/bin -Dz3.version=$Z3_VERSION
+ant publish-z3 -Dz3.path=$Z3_DIR -Dz3.version=$Z3_VERSION
 ```
-Finally follow the instructions shown in the message at the end.
+Example:
+```
+ant publish-z3 -Dz3.path=/workspace/solvers/z3/z3-z3-4.13.3 -Dz3.version=4.13.3
+```
+Finally, follow the instructions shown in the message at the end.
```
daniel-raffler, line 76: Would these steps work on windows/mac as well? Or is the idea to always use the docker image and then cross compile for all platforms from there? If so, we might want to make this more clear in the documentation.

doc/Developers-How-to-Release-into-Ivy.md
```diff
@@ -185,37 +211,44 @@ but in the normal system environment, where some testing can be applied by the d
...
-      -Dgmp.path=$WD/gmp-6.1.2 \
-      -Dmathsat-windows.path=$WD/mathsat-5.6.7-win64-msvc \
-      -Dmpir-windows.path=$WD/mpir-2.7.2-win \
-      -Djdk-windows.path=$WD/jdk-11 \
-      -Dmathsat.version=5.6.7
+We publish MathSAT for Linux (x64 and arm64) and Windows (x64) systems at once.
+The build process can fully be done on a Linux system, 
+and requires several dependencies, such as gcc, x86_64-w64-mingw32-gcc, and aarch64-linux-gnu-gcc.
+We prefer to use the Docker container based on Ubuntu 22.04 (better use Ubuntu 18.04 for older GLIBC!)
+for compiling the dependencies and assembling the libraries.
```
daniel-raffler, line 218: Maybe this paragraph from MathSAT could simple be moved further up?

daniel-raffler (commented):
I've tried building the Z3 and Bitwuzla backends and everything seems to be working fine. Unfortunately I don't have any ARM system do  much more testing.

build/build-publish-solvers/solver-bitwuzla.xml
```diff
@@ -13,8 +13,9 @@ SPDX-License-Identifier: Apache-2.0
 <!-- vim: set tabstop=8 shiftwidth=4 expandtab sts=4 filetype=ant fdm=marker: -->
 <project name="publish-solvers-bitwuzla" basedir="." xmlns:ivy="antlib:org.apache.ivy.ant">
 
-    <property name="bitwuzla.buildForLinux" value="true"/>
-    <property name="bitwuzla.buildForWindows" value="true"/>
+    <property name="bitwuzla.buildForLinuxX64" value="true"/>
+    <property name="bitwuzla.buildForLinuxArm64" value="true"/>
+    <property name="bitwuzla.buildForWindowsX64" value="true"/>
```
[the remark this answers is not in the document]
kfriedberger: For testing single steps, having a property might be beneficial. For publication, one might always want all versions included. I would not remove the properties.

doc/Developers-How-to-Release-into-Ivy.md
kfriedberger: Z3 is not compiled during these steps, but only packaged. We generate sources and JavaDoc for Java bindings and pack them, too. I wuld assume that these steps are independent of any platform or operating system and also work on Windows or MacOs.

doc/Developers-How-to-Release-into-Ivy.md
kfriedberger: Until now, we did not enforce any ordering of solvers in the documentation. One could apply alphabetical sorting as extra cleanup, possibly after this PR is merged.

build.xml
```diff
@@ -79,12 +79,14 @@ SPDX-License-Identifier: Apache-2.0
     <!-- Main targets -->
 
     <target name="clean" description="Clean">
-        <property name="bitwuzla.path" value="${user.dir}/lib/native/source/libbitwuzla"/>
+        <property name="jarFiles" value="${class.dir}/** ${ivy.module}-*.jar ivy-*.xml *.jar"/>
+        <property name="libraryFiles" value="*.so *.dll *.dylib x64/*.so x64/*.dll x64/*.dylib arm64/*.so arm64/*.dll arm64/*.dylib"/>
```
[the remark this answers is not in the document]
kfriedberger: I am unsure whether folders can be used in a 'fileset' attribute.

Currently, the folder structure is actually not optimal. We have the `root` directory for all architecture-independent or only x64-provided libraries and jars, and `x64` or `arm64` for specific libraries. It could be nicer to have an extra `dist` directory for everything we want to publish, with optional subdirectories for architecture.
Then we could simply remove `dist` completely.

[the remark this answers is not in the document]
kfriedberger: With commit cd2529a68dafccc58a69f21d26f34ec04d96d349, we introduced a `dist` directory for solvers. This allows us to remove the directories `arm64` and `x64` and simplify the cleanup task.

doc/Developers-How-to-Release-into-Ivy.md
daniel-raffler: Thanks for updating the documentation!

kfriedberger:
@daniel-raffler and @baierd : could you take a look and provide feedback? I would like to merge this PR in near time.

Beneigh Java-based solvers SMTInterpol and Princess, we now have 4 native solvers that are working on ARM64 platform: 
- with Linux on ARM64: Z3, MathSAT5, Bitwuzla, and OpenSMT.
- with Windows and MacOS on ARM64: Z3

The current changes to Ivy are backwards-compatible and allow to use and update existing JavaSMT integrations without further changes (such as in CPAchecker). If the user wants to use binary solvers for ARM64, he can modify his Ivy dependency to load the corresponding files. We already provide the corresponding files in Maven, so the user specify them via the pom-file..

build/build-maven-publish.xml
```diff
@@ -179,9 +203,9 @@ SPDX-License-Identifier: Apache-2.0
         <!-- prepare the pom-file -->
         <generate-solver-pom-file/>
         <!-- then publish the files -->
-        <stage-solver-file filename="libboolector" fileending="so"/>
-        <stage-solver-file filename="libminisat" fileending="so"/>
-        <stage-solver-file filename="libpicosat" fileending="so"/>
+        <stage-solver-file filename="libboolector" classifier="libboolector" fileending="so"/>
```
baierd, line 206: Maybe we should add the `x64` suffix for all existing solvers as well, so that we can discern their architecture more easily, and prepare adding ARM for more solvers. (Although we will most likely not support/get support for older solvers to be fair)

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
```diff
@@ -119,20 +122,38 @@ private void requireSupportedOperatingSystem() {
...
+      for (String dependency : getRequiredLibcxx(library)) {
+        if (e.getMessage().contains("version `" + dependency + "' not found")) {
           return false;
         }
       }
     }
     return true;
   }
 
+  private String[] getRequiredLibcxx(String library) {
```
baierd, line 142: This is very nice!
We should add this information (and requirements in general) to the README.

baierd (changes requested):
Overall very nice. I will test this PR on my Windows machine later today.

I have some small questions/suggestions, but did not find anything mayor.

build/build-maven-publish.xml
```diff
@@ -160,12 +181,15 @@ SPDX-License-Identifier: Apache-2.0
...
         <property name="stage.revision" value="${bitwuzla.revision}"/>
+        <property name="libDir.x64" value="lib/java/runtime-bitwuzla/x64"/>
+        <property name="libDir.arm64" value="lib/java/runtime-bitwuzla/arm64"/>
         <!-- prepare the pom-file -->
         <generate-solver-pom-file/>
         <!-- then publish the files -->
         <stage-solver-file filename="bitwuzla" fileending="jar"/>
-        <stage-solver-file filename="libbitwuzlaj" fileending="so"/>
-        <stage-solver-file filename="libbitwuzlaj" fileending="dll"/>
+        <stage-solver-file filename="libbitwuzlaj" classifier="libbitwuzlaj-x64" filedirectory="${libDir.x64}" fileending="so"/>
```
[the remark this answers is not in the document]
kfriedberger: While we can provide a backward-compatible version in Ivy/Ant, the same procedure was not implemented for Maven.

The difference is:
- Ivy needs to set a (global or project-dependent) artifact pattern, where an optional architecture flag is specified. See [build/build-ivy.xml L69](https://github.com/sosy-lab/java-smt/blob/13b3f1267994f33139ea906353eb3e08c532799c/build/build-ivy.xml#L69). If this is change not applied (also in other projects like CPAchecker), the developer will only access the x64-version of libraries.
- Maven allows to just add the architecture-flag directly: (example pom.xml)[https://github.com/sosy-lab/java-smt/blob/13b3f1267994f33139ea906353eb3e08c532799c/doc/Example-Maven-Project/pom.xml#L115], which makes it much simpler to update.

build/build-publish-solvers/macros.xml
```diff
@@ -31,19 +32,20 @@ SPDX-License-Identifier: Apache-2.0
...
                     pubrevision="@{solverVersion}"
                     resolver="Sosy-Lab-Publish"
-                    artifactspattern="[artifact]-[revision](-[classifier]).[ext]"
+                    artifactspattern="@{distDir}/([arch]/)[artifact]-[revision](-[classifier]).[ext]"
                     status="release"
             />
             <echo>
                 You now want to run
-                svn add repository/${ivy.organisation}/${ivy.module}/*-@{solverVersion}*
+                svn add repository/${ivy.organisation}/${ivy.module}/*-@{solverVersion}* repository/${ivy.organisation}/${ivy.module}/*/*-@{solverVersion}*
```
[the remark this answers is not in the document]
kfriedberger: This line is correct, because we use additional directories `x64` and `arm64` for architecture-dependent files.

Examples:
- https://www.sosy-lab.org/ivy/org.sosy_lab/javasmt-solver-z3/arm64/
- https://www.sosy-lab.org/ivy/org.sosy_lab/javasmt-solver-z3/x64/

When publishing a solver for a specific architecture the first time, the line might even be insufficient, because SVN does not track contents in a new directory, if the directory itself is not tracked. I ignored this special case and just extended the command for the usual case, like updating an existing solver.

build/build-maven-publish.xml
kfriedberger: I though about this topic as well and would do the following:
- If there is a (sufficiently) new version of a solver, where we could provide multiple architectures or need to compile the new  library anyway, then update the Ivy and Maven configurations and scripts.
- Otherwise, keep current state.

I did this with OpenSMT in the last few commits in this PR. It was quite easy to compile it for ARM64 after refactoring the script and playing around with the GMP dependency. :smile:

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
kfriedberger: Good point. I will update the Linux-column in the README with this information.

Overall, we might think about dropping support for anything older than Ubuntu 22.04 (`GLIBC_2.33` and `GLIBCXX_3.4.26`), because the overhead in developement effort is quite high for quite few users.

build/build-maven-publish.xml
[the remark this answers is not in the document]
baierd: Thanks for the explanation!
We should add this to our guide for Maven.
Have you checked Gradle? Does it work like Maven?

build/build-publish-solvers/macros.xml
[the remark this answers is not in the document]
baierd: Ah i get it. Thanks!

build/build-maven-publish.xml
baierd: I agree with your assessment!
It's nice that apparently multiple solvers work with ARM64 rather easily.

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
baierd: Thank you!

I agree with that. Maybe we should release one more version with minimal changes, e.g. 5.1, and add all of the coming changes in a way that users can prepare.

baierd:
Something that might be worth a try is adding a static `info()` method that tries to read your OS and gives you information about available solvers, dependencies and requirements (e.g. copy the Windows binaries for a Dev build), or possibly even checks those requirements for you.

kfriedberger:
I would not expect such a info-feature from a library, but only in the main application.

build/build-maven-publish.xml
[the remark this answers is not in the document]
kfriedberger: Gradle is quite similar to Maven. I expect it to work.
I have not yet updated the Gradle files, but will do it soon.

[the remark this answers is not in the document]
kfriedberger: Both, Maven and Gradle will depend on the latest release of JavaSMT to be included.
I updated the Maven examples with all newwest solvers, however, kept Gradle as-is for now. We need to update both after the merge and the corresponding release.

kfriedberger:
For executing and testing the ARM-based binaries, you can apply Podman/Docker with Qemu.
A corresponding Dockerfile would look like this:

```
FROM arm64v8/ubuntu:22.04

ENV DEBIAN_FRONTEND=noninteractive
ENV TZ=UTC
ENV LANG=C.UTF-8

RUN apt-get update && apt-get install -y --no-install-recommends \
  wget curl \
  ant git \
  openjdk-17-jre-headless \
  openjdk-17-jdk-headless \
  libgomp1
```

Build with:

```
podman build --platform=linux/arm64 - < ubuntu2204-arm64.Dockerfile
```
The output will contain the ID in the last line.

Run with:
```
podman run -it \
    --mount type=bind,source=${HOME}/workspace,target=/workspace \
    --workdir /workspace/java-smt \
    ID
```

Podman/Docker with qemu is emulating ARM64 on x64 about 3-10 times slower than normal execution. I am thinking about an ARM-based CI, however Podman/Docker with Qemu might be too slow for that.

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
kfriedberger: Readme is updated now.

baierd (approved):
Looks good. Thanks for incorporating my suggestions! 

Do we want to document the Podman/Docker info in some file or issue?

kfriedberger:
The Podman/Docker file might be interesting for additional CI jobs. We will surely come across that topic in https://github.com/sosy-lab/java-smt/issues/384.

Issue #371, opened 2024-04-11
Bitwuzla is unusualy slow for FP query based on IEEE conversion


baierd:
Bitwuzla is very slow for the tests in `FloatingPointFormulaManagerTest` based on the `getListOfFloats()` method. We should extract examples and inform the devs of Bitwuzla about this.

baierd:
Whats the current state of this issue and what can we do? @daniel-raffler could you investigate please?

PR #396, opened 2024-09-19
Bitwuzla: Fix performance issues in FloatingPointFormulaManagerTest

daniel-raffler:
Hello,
this MR aims to fix the performance issues described in #371. The problem was not about the conversion operation itself, but about the large number of ProverEnvironments that would be created, one for each test input. We solve this by reusing the same ProverEnvironment for all inputs of the test. This also improves the performance of Z3 for those same tests.

In addition to that the MR includes a number of minor fixes:
- Bitwuzla does not support float-to-bitvector conversions and we are using a work-around that introduces new side-conditions for the casts. We now make sure that side-conditions are only pushed onto the stack when needed. The old version would simply include everything in the query, even for variables that don't occur in the assertions, which can cause significant overhead
- We fixed the names that are used for the variables in these side-conditions to make sure that there are no clashes
- `FunctionDeclarationKind` gets a new Symbol `FP_REM` which was missing
- We also added support for `BV_SHR` and `BV_ROL`, `BV_ROLI`, `BV_ROR`, `BV_RORI` to the Bitwuzla `FormulaCreator`

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
@@ -212,6 +212,16 @@ public class BitwuzlaFloatingPointManager
 
     Sort bvSort = termManager.mk_bv_sort(sizeExp + sizeSig);
 
+    // The following code creates a new variable that is returned as result.
+    // Additionally, we track constraints about the equality of the new variable and the FP number,
+    // which is added onto the prover stack whenever the new variable is used as assertion.
+
+    // TODO This internal implementation is a technical dept and should be removed.
+    //   The additional constraints are not transparent in all cases, e.g., when visiting a
+    //   formula, creating a model, or transferring the assertions onto another prover stack.
+    //   A better way would be a direct implementation of this in Bitwuzla, without interfering
+    //   with JavaSMT.
+
     // Note that NaN is handled as a special case in this method. This is not strictly necessary,
     // but if we just use "fpTerm = to_fp(bvVar)" the NaN will be given a random payload (and
     // sign). Since NaN payloads are not preserved here anyway we might as well pick a canonical
@@ -231,7 +241,7 @@ public class BitwuzlaFloatingPointManager
                 termManager.mk_term(Kind.FP_TO_FP_FROM_BV, bvVar, sizeExp, sizeSig),
                 pNumber));
 
-    bitwuzlaCreator.addVariableCast(newVariable, equal);
+    bitwuzlaCreator.addConstraintForVariable(newVariable, equal);
     return bvVar;
   }
 
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
@@ -201,10 +201,6 @@ public class BitwuzlaFloatingPointManager
         pTargetType.getMantissaSize() + 1);
   }
 
-  private String newVariable() {
-    return "__CAST_FROM_BV_" + counter++;
-  }
-
   @Override
   protected Term toIeeeBitvectorImpl(Term pNumber) {
     int sizeExp = pNumber.sort().fp_exp_size();
@@ -225,11 +221,13 @@ public class BitwuzlaFloatingPointManager
     // Note that NaN is handled as a special case in this method. This is not strictly necessary,
     // but if we just use "fpTerm = to_fp(bvVar)" the NaN will be given a random payload (and
     // sign). Since NaN payloads are not preserved here anyway we might as well pick a canonical
-    // representation.
-    Term bvNaN =
-        termManager.mk_bv_value(bvSort, "0" + "1".repeat(sizeExp + 1) + "0".repeat(sizeSig - 2));
+    // representation, e.g., which is "0 11111111 10000000000000000000000" for single precision.
+    String nanRepr = "0" + "1".repeat(sizeExp + 1) + "0".repeat(sizeSig - 2);
+    Term bvNaN = termManager.mk_bv_value(bvSort, nanRepr);
 
-    String newVariable = newVariable();
+    // TODO creating our own utility variables might eb unexpected from the user.
+    //   We might need to exclude such variables in models and formula traversal.
+    String newVariable = "__JAVASMT__CAST_FROM_BV_" + counter++;
     Term bvVar = termManager.mk_const(bvSort, newVariable);
     Term equal =
         termManager.mk_term(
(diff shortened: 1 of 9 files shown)


src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
```diff
@@ -305,6 +313,10 @@ private FunctionDeclarationKind getDeclarationKind(Term term) {
       return FunctionDeclarationKind.FP_CASTTO_UBV;
     } else if (kind.equals(Kind.BV_XOR)) {
       return FunctionDeclarationKind.BV_XOR;
+    } else if (ImmutableList.of(Kind.BV_ROL, Kind.BV_ROLI, Kind.BV_ROR, Kind.BV_RORI)
+        .contains(kind)) {
+      // Rotations are not yet supported by FunctionDeclarationKind
+      return FunctionDeclarationKind.OTHER;
```
kfriedberger, line 319: Please do not just add such a mapping here, but report this as a bug for JavaSMT. The FunctionDeclarationKinds are missing on our interface.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
```diff
@@ -564,11 +576,44 @@ public Object convertValue(Term term) {
...
 
-  public void addVariableCast(Term equal) {
-    variableCasts.add(equal);
+  public void addVariableCast(String newVariable, Term equal) {
+    variableCasts.put(newVariable, equal);
   }
 
-  public Iterable<Term> getVariableCasts() {
-    return variableCasts;
+  private Map<String, Set<String>> calculateTransitions(Map<String, Term> pVariableCasts) {
```
kfriedberger, line 583: What does this code? This needs to be commented and explained.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
```diff
@@ -564,11 +576,44 @@ public Object convertValue(Term term) {
...
+
+  private Set<String> takeAStep(Map<String, Set<String>> pTransitions, Set<String> pVariables) {
+    ImmutableSet.Builder<String> builder = ImmutableSet.builder();
+    for (String var : pVariables) {
+      builder.addAll(pTransitions.get(var));
+    }
+    return builder.build();
+  }
+
+  public Iterable<Term> getVariableCasts(Iterable<Term> pTerms) {
```
kfriedberger, line 607: What does this code? This needs to be commented and explained.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -726,6 +726,8 @@ private FunctionDeclarationKind getDeclarationKind(long f) {
         return FunctionDeclarationKind.FP_DIV;
       case Z3_OP_FPA_MUL:
         return FunctionDeclarationKind.FP_MUL;
+      case Z3_OP_FPA_REM:
+        return FunctionDeclarationKind.FP_REM;
```
kfriedberger, line 730: Good fix.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaCreator.java
```diff
@@ -600,6 +600,7 @@ private Term normalize(Term operator) {
           .put(Kind.FLOATINGPOINT_ADD, FunctionDeclarationKind.FP_ADD)
           .put(Kind.FLOATINGPOINT_SUB, FunctionDeclarationKind.FP_SUB)
           .put(Kind.FLOATINGPOINT_MULT, FunctionDeclarationKind.FP_MUL)
+          .put(Kind.FLOATINGPOINT_REM, FunctionDeclarationKind.FP_REM)
```
kfriedberger, line 603: Good fix.

src/org/sosy_lab/java_smt/test/FloatingPointFormulaManagerTest.java
```diff
@@ -773,24 +774,44 @@ public void ieeeFpConversion() throws SolverException, InterruptedException {
         .isTautological();
   }
 
+  /**
+   * Map the function over the input list and prove the returned assertions.
+   *
+   * @param f A function that takes values from the list and returns assertions
+   * @param args A list of arguments to the function
+   */
+  private <T> void proveForAll(Function<T, BooleanFormula> f, List<T> args)
```
kfriedberger, line 783: just styling and personal preference: switch parameters. Using the list of args upfront makes it sound like `prove for all <numbers> that <function> is satisfiable`.

kfriedberger (commented):
Looks good overall. I annotated some minor things.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
kfriedberger: Looks like a fixedpoint iteration that collects variables from terms, and from related terms.

daniel-raffler:
Thanks for the review! I've opened a new issue about the rotations (https://github.com/sosy-lab/java-smt/issues/397). Once this has been resolved the workaround can be removed here. I've also added some more documentations about `getVariableCasts`() and merged the calculation into a single function.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
daniel-raffler: > Looks like a fixedpoint iteration that collects variables from terms, and from related terms, and from their related terms,..

Yes, that's the idea. Whenever a formula is pushed onto the assertion stack we need to check if it uses any variables that come from fp-to-bv casts and then include side-conditions for those casts. This needs to be done recursively until all casts are covered.

I think there are a few other solvers (like CVC5) that don't allow fp-to-bv casting?  It would be possible to include this code in AbstractBitformulaManager so that it than can be used by those solvers as well.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
daniel-raffler: I've merged master with the new changes and removed this workaround

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
```diff
@@ -198,22 +201,36 @@ protected Term fromIeeeBitvectorImpl(Term pNumber, FloatingPointType pTargetType
         pTargetType.getMantissaSize() + 1);
   }
 
+  private String newVariable() {
+    return "__CAST_FROM_BV_" + counter++;
```
kfriedberger, line 205: We should name this utility variable `__javasmt_cast_from_bv_<ctr>` to have some internal naming.

kfriedberger: done

Issue #372, opened 2024-04-12
Problems with common SMTLIB2 Strings for OpenSMT2 and Bitwuzla


baierd:
Both OpenSMT2 and Bitwuzla need the removal of certain parts of SMTLIB2 Strings before we can parse them. Examples are comments and `set-logic`.
This is currently done lazily and incomplete. For example escape characters are not taken into account.
We should ask for better support by the developers and/or improve our filtering (possibly by using the branch for the SMTLIB2 parser).

baierd:
Progress for OpenSMT2 + tests in branch [improve opensmt2 parsing](https://github.com/sosy-lab/java-smt/tree/improve_opensmt2_parsing)

baierd:
Whats the current state of this issue and what can we do? @daniel-raffler could you investigate please?

PR #401, opened 2024-09-28
Strip SMT-LIB2 queries before parsing

daniel-raffler:
Hello,
this MR adds some basic preprocessing for SMT-LIB2 queries before they are sent off to the solver for parsing. Most of the functionality has already been included as part of the OpenSMT/Bitwuzla backends, but it has now been improved and moved directly to `AbstractFormulaManger` where it can be used by all solvers. This should avoid further redundancies in the code and will help us solves the issues described in #372.

The idea behind the MR is to restrict ourselves to a simplified version of the SMT-LIB2 format that is sufficient to parse (and print) SMT-LIB2 queries. For this we remove comments and commands like `(set-logic ...)` and `(exit)` as they keep us from reusing the solver context. Operations like `(push)` or `(get-value ..)` are also removed, and we only keep declarations and assertions in our query. These simplified SMT-LIB2 scripts are then passed on to the solver where the actual parsing is done. We also use the same transformation to clean up the output of `dumpFormula` in a solver independent way.


src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
+   * <p>We remove all comments and put each command on its own line. Declarations and asserts are
+   * kept and everything else is removed. For <code>(set-logic ..)</code> we make sure that it's at
+   * the top of the file before removing it, and for <code>(exit)</code> we make sure that it can
+   * only occur as the last command.
+   */
+  private String sanitize(String formulaStr) {
+    List<String> tokens = tokenize(formulaStr);
+
+    StringBuilder builder = new StringBuilder();
+    int pos = 1; // index of the current token
```
kfriedberger, line 448: Be careful with indices. Developers like to start counting from zero... :smile:  Looks good.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
+  }
+
+  /**
+   * Check if the token is <code>(exit ...)</code>.
+   *
+   * <p>Use {@link #tokenize(String)} to turn an SMT-LIB2 script into a string of input tokens.
+   */
+  protected static boolean isExitToken(String token) {
+    return token.matches("\\(\\s*exit.*");
+  }
```
kfriedberger, line 434: Is the visibility required? Maybe make it "private"?

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
+
+  /**
+   * Check if the token is <code>(set-logic ..)</code>.
+   *
+   * <p>Use {@link #tokenize(String)} to turn an SMT-LIB2 script into a string of input tokens.
+   */
+  protected static boolean isSetLogicToken(String token) {
+    return token.matches("\\(\\s*set-logic.*");
+  }
+
```
kfriedberger, line 426: Is the visibility required? Maybe make it "private"?

kfriedberger (changes requested):
Overall, this PR is quite well-written and very useful. I like this approach.

I added some comments.
Main point: Could you add a few more tests for the parser?
Thanks.

daniel-raffler:
> Overall, this PR is quite well-written and very useful. I like this approach.
> 
> I added some comments. Main point: Could you add a few more tests for the parser? Thanks.

Hi Karlheinz,
thanks for the feedback!
I'll be quite busy these next few days, but I can add some more test later this week.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
+  /**
+   * Check if the token is a function definition.
+   *
+   * <p>Use {@link #tokenize(String)} to turn an SMT-LIB2 script into a string of input tokens.
+   */
+  protected static boolean isDefinitionToken(String token) {
+    return token.matches("\\(\\s*define-fun.*");
+  }
+
```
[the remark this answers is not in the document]
daniel-raffler: `isDefinitionToken` is actually used in `BitwuzlaFormulaManger` [here](https://github.com/sosy-lab/java-smt/blob/b706b7b6f6ab18e47e968d5d5b9d7cd7b4a2fe5a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaManager.java#L69). But I can make the other ones private.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
+  /**
+   * Check if the token is a function or variable declaration.
+   *
+   * <p>Use {@link #tokenize(String)} to turn an SMT-LIB2 script into a string of input tokens.
+   */
+  protected static boolean isDeclarationToken(String token) {
+    return token.matches("\\(\\s*(declare-const|declare-fun).*");
+  }
+
```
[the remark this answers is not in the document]
daniel-raffler: `isDefinitionToken` is actually used in BitwuzlaFormulaManger [here](https://github.com/sosy-lab/java-smt/blob/b706b7b6f6ab18e47e968d5d5b9d7cd7b4a2fe5a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaManager.java#L69). But I can make the other ones private.

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
```diff
@@ -340,8 +341,8 @@ public Appender dumpFormula(IFormula formula, final PrincessFormulaCreator creat
     final IExpression lettedFormula = tuple._1();
     final Map<IExpression, IExpression> abbrevMap = asJava(tuple._2());
 
-    return new Appenders.AbstractAppender() {
-
+    StringBuilder builder = new StringBuilder();
+    new Appenders.AbstractAppender() {
```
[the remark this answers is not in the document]
daniel-raffler: This Appender is pretty large and used several helper functions that would need to be moved to the main class. It also captures a local variable `abbrevMap` that would have to passed around as an argument. It probably could be rewritten, but I think the code might be simpler as it is.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
daniel-raffler: Fixed 😉

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -261,11 +265,225 @@ public EnumerationFormulaManager getEnumerationFormulaManager() {
...
   public Appender dumpFormula(BooleanFormula t) {
-    return dumpFormula(formulaCreator.extractInfo(t));
+    return new Appenders.AbstractAppender() {
+      @Override
+      public void appendTo(Appendable out) throws IOException {
+        String raw = dumpFormulaImpl(formulaCreator.extractInfo(t));
+        out.append(sanitize(raw));
+      }
+    };
   }
```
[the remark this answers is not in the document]
kfriedberger: Well, not possible. :-)

src/org/sosy_lab/java_smt/solvers/princess/PrincessEnvironment.java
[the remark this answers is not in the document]
kfriedberger: ok, then keep it as it is now.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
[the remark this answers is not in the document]
daniel-raffler: Yes, I think we need an actual `AbstractAppender` object for the `toString` overload. When I changed the code it created an `Appender` instance, but `toString` would just return the Object name and not the printed String.

daniel-raffler:
@kfriedberger 
I've added a new test class [TokenizerTest](https://github.com/sosy-lab/java-smt/blob/improve_opensmt2_parsing/src/org/sosy_lab/java_smt/test/TokenizerTest.java). Could you have another look and tell me if there is still something missing?

src/org/sosy_lab/java_smt/test/SanitizerTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Test
+  public void validLogicTest() throws SolverException, InterruptedException {
+    BooleanFormula expected = mgr.parse("(declare-const v Int)(assert (= v 3))");
+    BooleanFormula broken =
+        mgr.parse("(set-logic QF_ALL)" + "(declare-const v Int)" + "(assert (= v 3))");
+    assertThatFormula(expected).isEquivalentTo(broken);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
kfriedberger, line 33: Please use `assertThrows(IllegalArgException.class, () -> mgr.parse(...))` if possible. This simplifies the test.

Same for following tests.

src/org/sosy_lab/java_smt/test/TokenizerTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.common.truth.Truth;
+import org.junit.Test;
+import org.sosy_lab.java_smt.basicimpl.Tokenizer;
+
+public class TokenizerTest {
+  @Test
+  public void validBrackets() {
```
kfriedberger, line 21: Writing that many test function is nice and show that you have thought bout several cases.
It would be simpler and better readable to combine several cases into one function.
For example:

```
public void testParserForValidInput() {
    assertThat(Tokenizer.tokenize("(assert (= 3 (+ 2 1)))")).containsExactly("(assert (= 3 (+ 2 1)))");
    assertThat(Tokenizer.tokenize("(assert (= 3;)\n(- 4 1)))")).containsExactly("(assert (= 3;)\n(- 4 1)))");
    ...
```

and 

```
public void testParserForInalidInput() {
    assertThrows(IllegalArgumentException.class, () -> Tokenizer.tokenize("(assert (= 3 (+ 2 1))))"));
    assertThrows(...);
    ...
```

kfriedberger (approved):
Overall: Well done.

Tests can be written in a cleaner way, see comments.

daniel-raffler:
Hello Karlheinz,
thanks for the review! I've summed up some of the test like you suggested.

Do you have any comments on [SanitizerTest](https://github.com/sosy-lab/java-smt/blob/improve_opensmt2_parsing/src/org/sosy_lab/java_smt/test/SanitizerTest.java)? Especially with the changes in fd420893e8f1fdc3a1dc8dd4fd5149846b927a23? The problem is that if we simply remove commands such as `(pop)` or `(reset)` it may change the end result of the SMT script. On the other hand, handling those commands in the sanitizer seems like overkill as the scripts normally *should* not contain them. I've now changed the code to just throw an exception if such a command is found. However, this might be too restrictive.

What's your opinion on that?

src/org/sosy_lab/java_smt/basicimpl/Tokenizer.java
```diff
@@ -0,0 +1,202 @@
...
+   *   <li>push
+   *   <li>pop
+   *   <li>reset-assertions
+   *   <li>reset
+   * </ul>
+   *
+   * <p>Use {@link #tokenize(String)} to turn an SMT-LIB2 script into a string of input tokens.
+   */
+  public static boolean isUnsupportedToken(String token) {
+    return matchesOneOf(token, "push", "pop", "reset-assertions", "reset");
```
kfriedberger, line 200: Instead of using a list of unsupported commands, it might be simpler to list the supported ones.
--> return !matchesOneOf(token, "assert", "exit", ...)

kfriedberger (approved):
Well done. Some smaller comments remain, then you can merge.

kfriedberger:
> Do you have any comments on [SanitizerTest]... > What's your opinion on that?

Good idea. This limits the input provided by a user and might avoid errors.

src/org/sosy_lab/java_smt/basicimpl/Tokenizer.java
daniel-raffler: There are some commands that are neither `supported`, nor `unsupported`. An example is `(check-sat)` which Bitwuzla will append to any SMTLIB formula it dumps. These commands should be skipped by the sanitizer as they are not needed. Throwing an exception would cause issues with some solvers.

I've renamed `isUnsupportedToken` to `isForbiddenToken` to make the difference more clear.

daniel-raffler:
> Well done. Some smaller comments remain, then you can merge.

Thanks for having another look! I think I've fixed all the issues you've mentioned. Unless you have any more comments I'll merge the branch tomorrow.

kfriedberger (approved):

daniel-raffler:
@kfriedberger 
I've merged the branch, but unfortunately the gitlab CI is now failing ([here](https://gitlab.com/sosy-lab/software/java-smt/-/pipelines/1487130293)). What's the best way to clean this up? Could you still revert the merge?

Most of the failed checks are about minor style issues, but there is a failing tests for yices2. This seems to be a genuine bug where the yices backend returns an invalid SMTLIB script when dumping formulas. The other failing checks should be fixed by the commits I made to [improve_opensmt2_parsing](https://github.com/sosy-lab/java-smt/tree/improve_opensmt2_parsing) after 4c2e56e2d311fffadfb6f3f6b40dab4fee5ba092 (where I tried the merge). However, this still needs testing. Is there a way to run the gitlab CI before merging the branch into master?

kfriedberger:
Sorry, this was partially my fault.
I did not check which CIs were executed and assumed that an JavaSMT-internal branch would be executed normally, including GItlabCI.

I prefer to never rollback a MR and just apply the fixes to the main/master branch. I noticed that this already happened by you and Daniel. Thanks.

PR #407, opened 2024-10-29
Add InterruptedExceptions to Model Generation 

baierd:
Adds `InterruptedException`s to model generation and catch Z3s solver specific exception when model generation is interrupted and throw a `InterruptedException` instead.

This adds the exception in quite a lot of places as we have the evaluation connected the model generation.

Also, we can't add the exception to the `Iterator` class and i threw a `RunTimeException` instead. We probably want to throw a more precise exception here.

Also, we should check if this is possible in other solvers as well.


PhilippWendler:
Note that adding a checked exception to a public method is an API-breaking change.

For the `Iterator`, throwing any exception would be a bug. We don't want the application to crash if Z3 happens to be interrupted at the wrong time (which we often cannot control), no matter whether the crash is due to a `Z3Exception` or a `RuntimeException` or something else. The best alternative is likely to set the thread's interrupt status and continue normally.

kfriedberger:
I would prefer a solution as in #408 where the InterruptException is handled as RuntimeException. This keeps the signatures cleaner and is compatible with Java-defined interfaces like Iterator.

baierd:
I completely agree that this solution is ugly and breaking.
I would also prefer something different.

However, i am unsure if we want to throw them as unchecked exceptions for the reasons Philipp pointed out.

I am also unsure if users would expect the interrupt flag of the thread to be set even if we document it, as we have to return something in the methods. And if we return a empty model users might just never notice.

PhilippWendler:
Yes, setting the interrupted flag is really only good if the code can continue as it would without the interrupt. Dummy data should never be returned in place of real data.

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
```diff
@@ -375,9 +378,14 @@ public void close() {
...
-  protected Long evalImpl(Long formula) {
+  protected Long evalImpl(Long formula) throws InterruptedException {
     LongPtr resultPtr = new LongPtr();
-    boolean satisfiableModel = Native.modelEval(z3context, model, formula, false, resultPtr);
+    boolean satisfiableModel;
+    try {
+      satisfiableModel = Native.modelEval(z3context, model, formula, false, resultPtr);
+    } catch (Z3Exception e) {
+      throw new InterruptedException("Z3 model evaluation was interrupted.");
+    }
```
kfriedberger, line 388: This code is wrong. Z3Exceptions can have other causes as well. Please use proper handling from `Z3FormulaCreator.handleZ3Exception(e)`.

src/org/sosy_lab/java_smt/solvers/z3/Z3OptimizationProver.java
```diff
@@ -197,8 +197,12 @@ private Optional<Rational> round(int handle, Rational epsilon, RoundingFunction
...
 
   @Override
-  protected long getZ3Model() {
-    return Native.optimizeGetModel(z3context, z3optSolver);
+  protected long getZ3Model() throws InterruptedException {
+    try {
+      return Native.optimizeGetModel(z3context, z3optSolver);
+    } catch (Z3Exception e) {
+      throw new InterruptedException("Z3 model generation interrupted.");
+    }
```
kfriedberger, line 205: This code is wrong. Z3Exceptions can have other causes as well. Please use proper handling from Z3FormulaCreator.handleZ3Exception(e).

src/org/sosy_lab/java_smt/solvers/z3/Z3TheoremProver.java
```diff
@@ -137,8 +137,12 @@ protected long getUnsatCore0() {
   }
 
   @Override
-  protected long getZ3Model() {
-    return Native.solverGetModel(z3context, z3solver);
+  protected long getZ3Model() throws InterruptedException {
+    try {
+      return Native.solverGetModel(z3context, z3solver);
+    } catch (Z3Exception e) {
+      throw new InterruptedException("Z3 model generation interrupted.");
```
kfriedberger, line 144: This code is wrong. Z3Exceptions can have other causes as well. Please use proper handling from Z3FormulaCreator.handleZ3Exception(e).

src/org/sosy_lab/java_smt/api/Model.java
```diff
@@ -52,11 +52,15 @@ public interface Model extends Evaluator, Iterable<ValueAssignment>, AutoCloseab
    */
   @Override
   default Iterator<ValueAssignment> iterator() {
-    return asList().iterator();
+    try {
+      return asList().iterator();
+    } catch (InterruptedException pE) {
+      throw new RuntimeException(pE);
```
kfriedberger, line 58: It would be nice to hide the runtime eexception here and just forward the InterruptedException, e.g., without changing the stacktrace and interrupt the exception trace.

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
```diff
@@ -206,7 +208,8 @@ private Collection<ValueAssignment> getConstantArrayAssignment(
    * @return a list of assignments {@code a[1]=0; a[2]=0; a[5]=0}.
    */
   private Collection<ValueAssignment> getArrayAssignments(
-      long arraySymbol, long arrayFormula, long value, List<Object> upperIndices) {
+      long arraySymbol, long arrayFormula, long value, List<Object> upperIndices)
+      throws InterruptedException {
     long evalDecl = Native.getAsArrayFuncDecl(z3context, value);
     Native.incRef(z3context, evalDecl);
     long interp = Native.modelGetFuncInterp(z3context, model, evalDecl);
```
kfriedberger, line 215: Based on the annotations from https://github.com/Z3Prover/z3/blob/master/src/api/java/Model.java , it seems that methods like "modelGetFuncInterp" can also throw a Z3Exception. We might want to check more places for InterruptedException.

kfriedberger (changes requested):
The API change is plausible and might be useful for the user.

Please fix the handling of exceptions.

src/org/sosy_lab/java_smt/solvers/z3/Z3OptimizationProver.java
kfriedberger: fixed

src/org/sosy_lab/java_smt/solvers/z3/Z3TheoremProver.java
kfriedberger: fixed

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
kfriedberger: fixed

src/org/sosy_lab/java_smt/api/Model.java
kfriedberger: ignore this

kfriedberger (approved):
There are several API changes from SolverException and InterruptedException.

As Z3 can throw such exception at a lot of places, this topic grows and we might end up with having exceptions in all API methods.

I approve this change. We need a minor version increment, maybe even a new major version for the next release.

@baierd and @PhilippWendler any more review comments before we can merge this PR?

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
baierd: Thanks for noticing! You already changed it, thanks!

baierd:
Looks good to me. The only thing that i would like to have a better handling for is the iterator and its exception, but i guess we have to live with that.

PhilippWendler:
> @baierd and @PhilippWendler any more review comments before we can merge this PR?

It seems like this was merged already, do you still want me to look at it?

kfriedberger:
@PhilippWendler Helpful comments are always welcome!

src/org/sosy_lab/java_smt/api/Model.java
```diff
@@ -52,11 +52,15 @@ public interface Model extends Evaluator, Iterable<ValueAssignment>, AutoCloseab
    */
   @Override
   default Iterator<ValueAssignment> iterator() {
-    return asList().iterator();
+    try {
+      return asList().iterator();
+    } catch (InterruptedException | SolverException e) {
+      throw new RuntimeException(e);
+    }
```
PhilippWendler, line 59: I think this should be improved.

First, it is absolutely clear that this needs proper documentation that tells users what to expect and that they should not use it if if they care about handling these exceptions. This should be added to the JavaDoc of the method as well as of the class and maybe even to `getModel()`, because users are unlikely to read the JavaDoc of an `iterator()` method that can be called implicitly in a for-each loop.

Second, using a `RuntimeException` is very generic and makes it look like a bug if it is thrown, even if it is not. Users who want to catch and handle these exceptions need to catch `RuntimeException` and risk catching too many exceptions. The API would be much nicer if a specific exception type was used that signals that some expected checked exception is contained that was not possible to be thrown directly. Maybe there even exists such an exception type already somewhere.

Third, for catching `InterruptedException` I think that the code should re-add the interrupted status of the thread.

In general it would also probably be a good idea to look at how other code with the same problem is designed and handles checked exceptions. AFAIR there are for example cases in the Java API where streams are used for I/O and checked exceptions are a problem, and there are probably several other such cases in the Java API and maybe Guava. If we look at their solutions, we do have to invest less time into finding our own good solution.

Apart from the API change here, I just noticed that if I did not overlook anything, `Model.iterator()` seems to be never overwritten in any of the solver-specific classes. Why is that?

I was under the impression as a JavaSMT user that I have two ways for iterating over model values: Either I retrieve all model values into a list with `asList()` or `getModelAssignments()` and have to pay for the initial creation and memory consumption, but can then cheaply iterate over all values repeatedly. Or I could iterate using `.iterator()` and retrieve values lazily from the solver, without O(n) memory consumption and upfront cost. The JavaDocs of `getModelAssignments()` and `iterator()` even still indicate that this performance difference may exist.

I wonder why is this no longer possible? Could this be re-added?

If the answer to the last question is "no", then the question remains why JavaSMT even offers two ways to retrieve model values that are in all circumstances the same. Every caller could easily use `getModel().asList().iterator()` or `getModelAssignments().iterator()` instead of `getModel().iterator`, this is not too much burden. In this case the `iterator()` method and the inheritance from `Iterable` could simply be marked as `@Deprecated` with the exception problem and a reference to `asList()` as explanation.

But in general I would prefer if a proper implementation of lazy model iteration is re-added.

kfriedberger:
It does not look like Java has though about that topic in general. There exist several partial solutions that each handle exceptions in their own way. Some references for handling Java exceptions in special situations:
- exceptions in Java lambda expressions: https://www.baeldung.com/java-lambda-exceptions
- sneaky throws: https://4comprehension.com/sneakily-throwing-exceptions-in-lambda-expressions-in-java/

Actually, JavaSMT did not use a lazy implementation of `Model::iterator()` for any solver since several years, i.e., we always forward the call of `Model::iterator()` to `Model::asList()` (previously named `toList` or `modelToList`) and iterate over the returned  list of assignments. I found at least one commit in 2016 regarding that topic: https://github.com/sosy-lab/java-smt/commit/645d66d07bb6a3b2bacf99b6da1f994a773ed2ea where this behaviour was already set in stone. We never had a user complaining about the non-lazy implementation, but rather even had the benefit of always providing `ValueAssignments` that were available after closing the model.

I do no longer see the need of providing an `Iterator`-interface for the Model as such, we could remove it and simplify the JavaSMT API.

PR #415, opened 2024-12-11
Fix ivy dependency org name

GrishaVar:
Note the org name here
`https://mvnrepository.com/artifact/org.sosy-lab`

Otherwise perhaps I can add a note to use different org names depending on mvn vs sosy-lab repo usage?

Also bumps version to 5.0.1


kfriedberger (approved):

kfriedberger:
Good catch. Thanks.

PR #418, opened 2024-12-25
#407: Revert API changes from introducing InterruptedException in more methods

kfriedberger:
In #407, we introduced a more detailed specification of exceptions for solvers, including InterruptedException and SolverException, to better handle model evaluation and prover operations.

However, based on user feedback and internal discussion, these changes caused significant disruption. To address this, we are reverting these API changes to restore the previous behavior and reduce the impact on users. We will still handle these exceptions internally, but will now throw them as RuntimeException to avoid changes in the API.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Table;
 import com.google.common.primitives.Longs;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.microsoft.z3.Native;
 import com.microsoft.z3.Z3Exception;
 import com.microsoft.z3.enumerations.Z3_ast_kind;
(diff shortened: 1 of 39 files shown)


daniel-raffler (commented):
I've just tried it with CPAchecker and there are still `SolverException` for `FormulaManager.applyTactics()` and `FormulaManager.simplify()`. Is this intentional or were those two methods missed?

kfriedberger:
This is intentional. The methods for `simplify` and `applyTactics` are more complex and can throw exceptions on interrupt or error.
The remaining API change is small but required.

src/org/sosy_lab/java_smt/solvers/z3/Z3FormulaCreator.java
```diff
@@ -169,6 +171,30 @@ final SolverException handleZ3Exception(Z3Exception e)
     throw new SolverException("Z3 has thrown an exception", e);
   }
 
+  /**
+   * This method handles a Z3Exception, however it only throws a RuntimeException. This method is
+   * used in places where we cannot throw a checked exception in JavaSMT due to API restrictions.
+   *
+   * @param e the Z3Exception to handle
+   * @return nothing, always throw a RuntimeException
+   * @throws RuntimeException always thrown for the given Z3Exception
```
PhilippWendler, line 180: This documentation is wrong, this method *never* throws a `RuntimeException`.

PhilippWendler (commented):
What cases where the actually problematic ones? From the discussion I inferred that model iteration was problematic, but this PR also removes the declared exception from a bunch of other methods, some of which even declare one of the two exceptions already. Shouldn't these declarations be kept?

> We will still handle these exceptions internally, but will now throw them as RuntimeException to avoid changes in the API.

This does not seem to be the case, though? I did not find any code that throws `RuntimeException`, instead the actual exceptions are thrown in an unchecked and undeclared way.

I would also say that this behavior should be clearly documented for users of the API.

Issue #414, opened 2024-12-11
Docker images not working on Ubuntu 24.10


daniel-raffler:
Hello everyone,
the docker images for building solver binaries don't seem to work for me on Ubuntu 24.10. When I run `docker/runUbuntuXX.sh` I'll be logged in as user `developer`, but the `/workspace` volume is mounted as `root`, making it impossible to create any files. Maybe there is a way to update the hack in `ubuntuXX.Dockerfile` ? Or we could just switch to `podman`? Ubuntu 24.10 comes with Docker version `26.1.3` and I'm using the `docker.io` package.

PhilippWendler:
I can only recommend to use Podman. It is not only a technically superior solution, it is also usable without root access, and it immediately gets rid of all the permission problems without having to pass any special parameters at all.

PR #423, opened 2024-12-29
Use Podman for the build container

daniel-raffler:
Hello,
as reported in #414 the docker build images are no longer working with the latest Ubuntu release. There seems to be an issue with the work-around we use to build the solver binaries as a user inside the Docker image. I've tried finding a solution, but had no luck so far. This pull request now aims to replace Docker with Podman entirely.

Podman is a drop in replacement for Docker and supports the same command line options, making larger changes to our build scripts unnecessary. It also runs in user mode by default, meaning that we no longer need the work around that is broken now.

### DIFF
--- a/docker/buildUbuntu1804.sh
+++ b/docker/buildUbuntu1804.sh
@@ -4,7 +4,7 @@
 # an API wrapper for a collection of SMT solvers:
 # https://github.com/sosy-lab/java-smt
 #
-# SPDX-FileCopyrightText: 2021 Dirk Beyer <https://www.sosy-lab.org>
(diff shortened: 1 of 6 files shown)


kfriedberger (approved):
The PR looks good from my side.

We can keep the directory as `docker` or rename it to something like `build-images`. I prefer to keep it "as is".

kfriedberger:
I rebuild the Docker images with Podman and uploaded/updated them into https://gitlab.com/sosy-lab/software/java-smt/container_registry/4100030 . 

Podman seems to be slower that Docker when starting containers.
For a simple build-process I did not notice a longer delay.

I will merge this PR.

daniel-raffler:
Thanks for merging!

Issue #26, opened 2016-01-21
Declaring variable "select" fails with SMTInterpol


PhilippWendler:
With SMTInterpol, the following code fails:

```
bmgr.makeVariable("select");
```

Exception:

```
de.uni_freiburg.informatik.ultimate.logic.SMTLIBException: Function select is already defined.
    at de.uni_freiburg.informatik.ultimate.logic.NoopScript.declareFun(NoopScript.java:120)
    at de.uni_freiburg.informatik.ultimate.smtinterpol.smtlib2.SMTInterpol.declareFun(SMTInterpol.java:1885)
    at org.sosy_lab.solver.smtinterpol.SmtInterpolEnvironment.declareFun(SmtInterpolEnvironment.java:304)
    at org.sosy_lab.solver.smtinterpol.SmtInterpolFormulaCreator.makeVariable(SmtInterpolFormulaCreator.java:84)
    at org.sosy_lab.solver.smtinterpol.SmtInterpolBooleanFormulaManager.makeVariableImpl(SmtInterpolBooleanFormulaManager.java:55)
    at org.sosy_lab.solver.smtinterpol.SmtInterpolBooleanFormulaManager.makeVariableImpl(SmtInterpolBooleanFormulaManager.java:1)
    at org.sosy_lab.solver.basicimpl.AbstractBooleanFormulaManager.makeVariable(AbstractBooleanFormulaManager.java:69)
    at org.sosy_lab.solver.test.SolverBasicTest.testArrayFunctions(SolverBasicTest.java:143)
```

The reason is that `select` is already an existing interpreted function in SMTInterpol.

I think that JavaSMT should transparently escape such reserved names for variables and functions by quoting them. Probably `select` is not the only case where this needs to be done.

kfriedberger:
The same problem appears for "true" and "false", if they are created as integer variables. SMTInterpol cannot create variables with such names. It might also happen for other reserved tokens.

kfriedberger:
SMTlib allows to escape names with `|name|`. Perhaps we should always apply this (transparently) to variables for SMTInterpol. The problem might then be moved to cases where parsing or dumping of formulas is needed.

cheshire:
Let's slow down a bit here, why do we have a variable with such a name in a first place?
I saw the corresponding ticket in CPAchecker, but don't we escape names with pipes in ctoformula package? I'm somewhat opposed to doing opaque escaping.

cheshire:
[...]

PR #424, opened 2025-01-01
Blacklist more SMTLIB keywords as variable/uf names

daniel-raffler:
Hello,
this will update the list of forbidden variable/uf names to include all SMTLIB commands, such as `check-sat` or `assert`, and all predefined function names, like `mod` or `fp.add`. While such symbol names are fine when using the API, they cause issues when printing the formulas as most solvers will fail to add the necessary quotes to the symbols. This will then causing issues in `FormulaManager.translateFrom()` as the generated output is invalid and can't to be read back by the target solver to translate the formula.

By forbidding such symbol names entirely we circumvent the issues. User should always use `FormulaManager.isValidName()` to check if their name is on the list, and substitute it with `FormulaManager.escape()` if necessary.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
@@ -15,8 +15,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableBiMap;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import java.io.IOException;
 import java.util.Arrays;
@@ -58,29 +58,135 @@ import org.sosy_lab.java_smt.utils.SolverUtils;
 public abstract class AbstractFormulaManager<TFormulaInfo, TType, TEnv, TFuncDecl>
     implements FormulaManager {
 
-  /**
-   * Avoid using basic mathematical or logical operators of SMT-LIB2 as names for symbols.
-   *
-   * <p>We do not accept some names as identifiers for variables or UFs, because they easily
-   * misguide the user. Most solvers even would allow such identifiers directly, currently only
-   * SMTInterpol has problems with some of them. For consistency, we disallow those names for all
-   * solvers.
-   */
   @VisibleForTesting
-  public static final ImmutableSet<String> BASIC_OPERATORS =
-      ImmutableSet.of("!", "+", "-", "*", "/", "%", "=", "<", ">", "<=", ">=");
+  public static final ImmutableList<String> RESERVED =
+      ImmutableList.of(
+          // Keywords
+          "_",
+          "!",
+          "as",
+          "let",
+          "exists",
+          "forall",
+          "match",
+          "par",
+
+          // Commands
+          "assert",
+          "check-sat",
+          "check-sat-assuming",
+          "declare-const",
+          "declare-datatype",
+          "declare-datatypes",
+          "declare-fun",
+          "declare-sort",
+          "define-fun",
+          "define-fun-rec",
+          "define-funs-rec",
+          "define-sort",
+          "echo",
+          "exit",
+          "get-assertions",
+          "get-assignment",
+          "get-info",
+          "get-model",
+          "get-option",
+          "get-proof",
+          "get-unsat-assumptions",
+          "get-unsat-core",
+          "get-value",
+          "pop",
+          "push",
+          "reset",
+          "reset-assertions",
+          "set-info",
+          "set-logic",
+          "set-option",
+
+          // Predefined symbols
+          // Arrays
+          "select",
+          "store",
+          "const",
+
+          // Bitvectors
+          "concat",
+          "extract",
+          // + any symbol starting with "bv"
+          "bvneg",
+
+          // Core
+          "true",
+          "false",
+          "not",
+          "=>",
+          "and",
+          "or",
+          "xor",
+          "=",
+          "distinct",
+          "ite",
+
+          // Floats
+          "roundNearestTiesToEven RoundingMode",
+          "roundNearestTiesToAway",
+          "roundTowardPositive",
+          "roundTowardNegative",
+          "roundTowardZero",
+          "RNE",
+          "RNA",
+          "RTP",
+          "RTN",
+          "RTZ",
+          "fp",
+          "+oo",
+          "-oo",
+          "+zero",
+          "-zero",
+          "NaN",
+          "to_fp",
+          "to_fp_unsigned",
+          // + any symbol starting with "fp."
+          "fp.neg",
+
+          // Integers and Reals
+          "-",
+          "+",
+          "*",
+          "div",
+          "mod",
+          "/",
+          "abs",
+          "<=",
+          "<",
+          ">=",
+          ">",
+          "divisible",
+          "to_real",
+          "to_int",
+          "is_int",
+
+          // Strings
+          // + any symbol starting with "str."
+          "str.concat",
+          // + any symbol starting with "re."
+          "re.opt");
 
   /**
-   * Avoid using basic keywords of SMT-LIB2 as names for symbols.
+   * Checks if the symbol name is a reserved keyword in SMTLIB2.
    *
    * <p>We do not accept some names as identifiers for variables or UFs, because they easily
    * misguide the user. Most solvers even would allow such identifiers directly, currently only
    * SMTInterpol has problems with some of them. For consistency, we disallow those names for all
    * solvers.
    */
-  @VisibleForTesting
-  public static final ImmutableSet<String> SMTLIB2_KEYWORDS =
-      ImmutableSet.of("true", "false", "and", "or", "select", "store", "xor", "distinct", "let");
+  public static boolean isReserved(String pVar) {
+    return pVar.startsWith("bv")
+        || pVar.startsWith("fp.*")
+        || pVar.startsWith("str.")
+        || pVar.startsWith("re.")
+        || RESERVED.contains(pVar);
+  }
 
   /**
    * Avoid using escape characters of SMT-LIB2 as part of names for symbols.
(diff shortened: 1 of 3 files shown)


src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -58,29 +58,160 @@
...
+          "divisible",
+          "to_real",
+          "to_int",
+          "is_int",
+
+          // Strings
+          // + any symbol starting with "str."
+          "str.concat",
+          // + any symbol starting with "re."
+          "re.opt");
```
kfriedberger, line 199: String and Regex operations are matched via "startsWith" and not via String equality. Do we need these entries here?

kfriedberger (commented):
This change disallows quite a few names for symbols.
Are we sure that we want to go in that direction?

What is the motivation for this PR? Parsing and writing SMTLIB is complex, and solvers need to escape all unexpeted names anyway. This PR seems to forbid such cases upfront. This might be valid, but it might be unfriendly for existing callers.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
daniel-raffler: The keywords in `RESERVED` are used by `VariableNamesTest` and I've included these two entries as "examples" for the tests. Otherwise, they could be removed. Come to think of it, `VariableNamesTest` is already kind of slow and it might even need to test all the entries in `RESERVED`. Maybe one or two examples for each "kind" of reserved word would be sufficient? Then we could just include those test cases directly in `VariableNamesTest` and remove these two dummy keywords from `RESERVED`.

daniel-raffler:
> This change disallows quite a few names for symbols. Are we sure that we want to go in that direction?
> 
> What is the motivation for this PR? Parsing and writing SMTLIB is complex, and solvers need to escape all unexpeted names anyway. This PR seems to forbid such cases upfront. This might be valid, but it might be unfriendly for existing callers.

This was motivated by an issue in CPAchecker ([Bitwuzla issues when using a distinct solver for interpolation](https://gitlab.com/sosy-lab/software/cpachecker/-/issues/1301)) were we try to use Bitwuzla as the main solver for CPAchecker. Since interpolation is not supported in Bitwuzla, we need to translate formulas from/to MathSAT whenever CPAchecker wants to calculate interpolants. For this we need to first dump the formulas to SMTLIB and then read them back in with the "target" solver. However, this will often fail with a parse error:
```
Exception in thread "main" java.lang.IllegalArgumentException: <string>:48:14: expected symbol after 'declare-fun'
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.BitwuzlaNativeJNI.Parser_parse__SWIG_0(Native Method)
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.Parser.parse(Unknown Source)
	at org.sosy_lab.java_smt.solvers.bitwuzla.BitwuzlaFormulaManager.parse(BitwuzlaFormulaManager.java:131)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.SeparateInterpolatingProverEnvironment.convertToMain(SeparateInterpolatingProverEnvironment.java:116)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.SeparateInterpolatingProverEnvironment.getInterpolant(SeparateInterpolatingProverEnvironment.java:92)
	at org.sosy_lab.cpachecker.util.predicates.smt.InterpolatingProverEnvironmentView.getInterpolant(InterpolatingProverEnvironmentView.java:32)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.strategy.ITPStrategy.getInterpolantFromSublist(ITPStrategy.java:166)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.strategy.SequentialInterpolation.getFwdInterpolants(SequentialInterpolation.java:157)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.strategy.SequentialInterpolation.getInterpolants(SequentialInterpolation.java:110)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.getInterpolants(InterpolationManager.java:783)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager$Interpolator.buildCounterexampleTrace(InterpolationManager.java:996)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.buildCounterexampleTrace0(InterpolationManager.java:449)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.lambda$buildCounterexampleTrace$0(InterpolationManager.java:328)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.callWithTimelimit(InterpolationManager.java:337)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.InterpolationManager.buildCounterexampleTrace(InterpolationManager.java:327)
	at org.sosy_lab.cpachecker.cpa.predicate.PredicateCPARefiner.performInterpolatingRefinement(PredicateCPARefiner.java:400)
	at org.sosy_lab.cpachecker.cpa.predicate.PredicateCPARefiner.checkCounterexample(PredicateCPARefiner.java:390)
	at org.sosy_lab.cpachecker.cpa.predicate.PredicateCPARefiner.performRefinementForPath(PredicateCPARefiner.java:281)
	at org.sosy_lab.cpachecker.cpa.predicate.PredicateStaticRefiner.performRefinementForPath(PredicateStaticRefiner.java:183)
	at org.sosy_lab.cpachecker.cpa.arg.AbstractARGBasedRefiner.performRefinementForPath(AbstractARGBasedRefiner.java:158)
	at org.sosy_lab.cpachecker.cpa.arg.AbstractARGBasedRefiner.performRefinement(AbstractARGBasedRefiner.java:104)
	at org.sosy_lab.cpachecker.core.algorithm.CEGARAlgorithm.refine(CEGARAlgorithm.java:310)
	at org.sosy_lab.cpachecker.core.algorithm.CEGARAlgorithm.run(CEGARAlgorithm.java:256)
	at org.sosy_lab.cpachecker.core.algorithm.ParallelAlgorithm.runParallelAnalysis(ParallelAlgorithm.java:390)
	at org.sosy_lab.cpachecker.core.algorithm.ParallelAlgorithm.lambda$createParallelAnalysis$4(ParallelAlgorithm.java:317)
	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:131)
	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:76)
	at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:82)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
The issue is that many of the programs use the C function `exit()` and CPAchecker will model this with a UF that has the same name. When the formula is then printed we get output that looks like this:
```
(declare-fun exit ((_ BitVec 32) ) (_ BitVec 8))
```
Since `exit` is already a SMTLIB command, this is not valid input and that's what leads to the parse error. From what I can see this mostly seems to happen for `exit` and `true`. The latter is already a reserved keyword in `AbstractFormulaManager` and the easiest solution would be to just add `exit` to the list. However, I'm somewhat concerned that we will eventually run into the same problem for other keywords, and that's why I've tried to include as many of them as possible.

As @PhilippWendler pointed out there is already an old JavaSMT issues [here](https://github.com/sosy-lab/java-smt/issues/26), but it's never been entirely resolved. The proper fix would be to escape such names transparently, but this would require us to write our own parser and printer for SMTLIB formulas, which still seem out of reach. Blacklisting the reserved symbols for now at least makes sure that the user gets an error immediately. Once the parser/printer is ready we can then allow them again.

Note that I added some tests in [supported_symbols_test](https://github.com/sosy-lab/java-smt/tree/supported_symbols_test) to see what symbol names are actually allowed by the solvers. You can run `ParserSymbolsTest.dumpTest()` to see the results. If the test fails for the assertion, it means that the (quoted) symbol can be parsed, but the solver fails to add quotes when printing it. If there is a different exception, it means that the solver won't parse the symbol at all, even when properly quoted.

From the tests it seems that most solvers don't quote reserved words when printing formulas and then fail to read back their own output. It's even worse for Bitwuzla, which fails to parse any reserved word, even when quotes are used. This means that just fixing formula printing won't solve the issue either.

daniel-raffler:
I've added transparent escaping/unescapig with a43cb6dc2528b851afdb0d2d759f3393dff46a3a. This means that all variables are now allowed, and the escaping is done automatically when needed.

We still have to figure  out what to do about all the tests that expect `makeVariable` to fail for certain illegal names. And there are couple of parsing bugs in the solvers that should be reported.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
baierd: I guess that it might be a good idea to split the tests into a full test run that only runs once in a while (e.g. weekly in the CI) and a subset of unit tests that run always, as this problem is also present in some other PRs. If @kfriedberger and @PhilippWendler agree i will create an issue.

baierd:
> I've added transparent escaping/unescapig with [a43cb6d](https://github.com/sosy-lab/java-smt/commit/a43cb6dc2528b851afdb0d2d759f3393dff46a3a). This means that all variables are now allowed, and the escaping is done automatically when needed.
> 
> We still have to figure out what to do about all the tests that expect `makeVariable` to fail for certain illegal names. And there are couple of parsing bugs in the solvers that should be reported.

Well we have to update the tests AND we have to test this new feature.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
kfriedberger: First, we already have split tests for all (slower) and some (fast) tests, that were introduced with the ARM-based CI.

Second, is the main issue here really, that the VariableNamesTests are slow individually, or just in general?
The tests only generate about 1000 variables (maybe a few more :-) ), and the issue is that we need to setup the whole JavaSMT context that often, e.g., we load the solver and initialize the managers several 1000 times, per solver.
Inlining the list of symbols into a simple loop might speedup the VariableNamesTests, with just loosing the granularity of reporting each symbol as individual test.

baierd:
> This change disallows quite a few names for symbols. Are we sure that we want to go in that direction?
> 
> What is the motivation for this PR? Parsing and writing SMTLIB is complex, and solvers need to escape all unexpeted names anyway. This PR seems to forbid such cases upfront. This might be valid, but it might be unfriendly for existing callers.

I see your point. But here is a counterpoint: if you want to use multiple solvers, then all keywords used in any solver should be forbidden for all. Otherwise you can't ensure that they all can parse the output of the others.

The solution might well be that we track forbidden names by solver and switch up the API such that a user can choose between only screening for names forbidden in the current solver or all.

PR #430, opened 2025-01-16
Extend compile.sh for libmathsat5j for macOS dylib

lembergerth:
The produced library can _not_ be distributed yet, because the script does not sign it.

### DIFF
--- a/lib/native/source/get_jni_headers.sh
+++ b/lib/native/source/get_jni_headers.sh
@@ -8,8 +8,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-if [ `uname` = "Darwin" ] ; then
-  echo "-I/usr/local/include -I/sw/include -I/System/Library/Frameworks/JavaVM.framework/Headers"
+if [ "$(uname)" = "Darwin" ] ; then
+  java_home=`readlink -f \`which java\``
+  java_home=`echo $java_home | sed 's#/jre/bin/java##' | sed 's#/bin/java##'`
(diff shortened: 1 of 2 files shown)


kfriedberger (approved):
Looks valid. I can not test this on MacOS.

Would it be possible/useful to
- cross-compile the bindings on Linux,
- publish the unsigned artifact via Ivy, and
- delay the signing process for the user, e.g. manually done after installation?

We might think of requesting a developer account from Apple to sign such libraries. Seems to be free of costs for open-source and educational projects. 

Perhaps we can even get some Gitlab runner / Github action for building and signing our artifacts automatically. That could be a nice student project.

lembergerth:
> Perhaps we can even get some Gitlab runner / Github action for building and signing our artifacts automatically. That could be a nice student project.

I think this would be the first necessary step.
I would like java-smt to support more solvers for macOS out-of-the-box, but manual testing each release would be a nightmare.

lembergerth:
> cross-compile the bindings on Linux,

I don't think this is possible.
Just building the library with a non-Apple clang on macOS already disallowed the signing process (but there was no nice error message, so I am not sure about the concrete issue).
So a cross-compiled library may not be signable.

kfriedberger:
AppVeyor has MacOS as platform available: https://www.appveyor.com/docs/macos-images-software/
This could be enough for JUnit tests. As first step could extend our CI-pipeline for AppVeyor for MacOS, as already written here: https://github.com/sosy-lab/java-smt/issues/384

As Apple is quite special in building and signing software, I am not sure about a efficient and maintainable long term solution.

Issue #412, opened 2024-11-24
Inconsistent handling of Unicode characters in String theory 


kfriedberger:
Different solvers return different results when using Unicode characters in String theory.
This should be analyzed. Maybe we need to fix JavaSMT or report to the solvers' developers.

Details: https://github.com/sosy-lab/java-smt/pull/391#discussion_r1847998043

daniel-raffler:
Thanks for opening the issue!

I've now added some tests:
* When creating String constants in Z3 all Unicode characters must be (SMTLIB) escaped. The solver will also escape all its output.
* Princess will accept any Java String and does not recognize SMTLIB escape sequences for Unicode characters. We added code in our backend to handle the conversion ourselves, but Strings from the model don't get converted back to SMTLIB format when they are printed.
* CVC4 behaves like Z3,
* But CVC5 is closer to Princess, even though it does not allow Unicode characters when creating String constants

It's not too difficult to convert between UTF-16 and the SMTLIB escape format, but we'll have to decide which encoding we want to use internally. Specifically, should the `Strings` in `StringFormula StringFormulaManager.makeString(String value)` and `String Evaluator.evaluate(StringFormula formula)` only allow SMTLIB encoded Unicode characters, or do we want to support all Java  Strings?

daniel-raffler:
I've added some more conversions and all solvers should now behave the same. The current format got `makeString(...)` and `evaluate(StringFormula formula)` is is SMTLIB escaped Strings, but we may still change this now. The argument for keeping it that way is that we want to stay as close to the SMTLIB standard as possible. This may help interoperability, for instance when someone reads in a SMTLIB script and then tries to recreate it with JavaSMT. In that case it might be confusing if escaped Unicode characters are not properly recognized.

On the other hand JavaSMT is written in Java, and the type `String` in `makeString(...)` and `evaluate(StringFormula formla)` suggest that any Java String should be valid input. The conversion from Java String to the SMTLIB format (and back) is easy enough and can be handled by JavaSMT automatically, so there is no reason why we would be bound by the SMTLIB standard on this issue.

Either choice will break the API, although I'd argue that keeping SMTLIB as format is more in line with how the functions used to work so far (we just didn't document it).

[...]

PR #431, opened 2025-01-16
Fix handling of Unicode characters in String theory

daniel-raffler:
Hallo,
this MR patches several minor issues remaining after #422:
* When escaping a String we now also escape `\` to make sure the backslash is preserved and doesn't get captured later when unescaping again
* We now always apply `unescape` first when creating a new String constant. This is needed even when the solver does not support plain Unicode and we'll have to escape the String again later.  This change closes an issue with Z3 and CVC4 where broken escape sequences were not handled correctly.
* Added two more tests and fixed a bug in another

As mentioned in https://github.com/sosy-lab/java-smt/issues/412 I think that `makeString` should not be translating SMTLIB escape sequences at all. However, this is an API breaking change and can still be considered some other time

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
@@ -29,8 +29,8 @@ class CVC4StringFormulaManager extends AbstractStringFormulaManager<Expr, Type, 
 
   @Override
   protected Expr makeStringImpl(String pValue) {
-    // The boolean enables escape characters!
-    return exprManager.mkConst(new CVC4String(escapeUnicodeForSmtlib(pValue), true));
+    String str = escapeUnicodeForSmtlib(unescapeUnicodeForSmtlib(pValue));
+    return exprManager.mkConst(new CVC4String(str, true));
   }
 
   @Override
--- a/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5StringFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5StringFormulaManager.java
@@ -27,7 +27,8 @@ class CVC5StringFormulaManager extends AbstractStringFormulaManager<Term, Sort, 
 
   @Override
   protected Term makeStringImpl(String pValue) {
-    return solver.mkString(escapeUnicodeForSmtlib(pValue), true);
+    String str = escapeUnicodeForSmtlib(unescapeUnicodeForSmtlib(pValue));
+    return solver.mkString(str, true);
   }
 
   @Override
--- a/src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
+++ b/src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
@@ -24,8 +24,11 @@ import org.junit.Test;
 import org.sosy_lab.java_smt.SolverContextFactory.Solvers;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Formula;
+import org.sosy_lab.java_smt.api.Model;
 import org.sosy_lab.java_smt.api.NumeralFormula.IntegerFormula;
+import org.sosy_lab.java_smt.api.ProverEnvironment;
 import org.sosy_lab.java_smt.api.RegexFormula;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 import org.sosy_lab.java_smt.api.SolverException;
 import org.sosy_lab.java_smt.api.StringFormula;
 
@@ -120,6 +123,26 @@ public class StringFormulaManagerTest extends SolverBasedTest0.ParameterizedSolv
   }
 
   // Tests
+
+  @Test
+  public void testInputEscape() throws SolverException, InterruptedException {
+    // Test if SMTLIB Unicode literals are recognized and converted to their Unicode characters.
+    assertEqual(smgr.length(smgr.makeString("Ξ")), imgr.makeNumber(1));
+    assertEqual(smgr.length(smgr.makeString("\\u{39E}")), imgr.makeNumber(1));
+  }
+
+  @Test
+  public void testOutputUnescape() throws SolverException, InterruptedException {
+    // Test if Unicode escape sequences get properly converted back when reading from the model.
+    try (ProverEnvironment prover = context.newProverEnvironment(ProverOptions.GENERATE_MODELS)) {
+      assertThat(!prover.isUnsat()).isTrue();
+      try (Model model = prover.getModel()) {
+        assertThat(model.evaluate(smgr.makeString("\\u{39E}"))).isEqualTo("Ξ");
+        assertThat(model.evaluate(smgr.concat(smgr.makeString("\\u{39E"), smgr.makeString("}"))))
+            .isEqualTo("\\u{39E}");
+      }
+    }
+  }
 
   @Test
   public void testRegexAll() throws SolverException, InterruptedException {
--- a/src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
+++ b/src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
@@ -1640,7 +1640,7 @@ public class StringFormulaManagerTest extends SolverBasedTest0.ParameterizedSolv
         .isNotEqualTo(Solvers.Z3);
 
     for (int i = 0; i < WORDS.size(); i++) {
-      for (int j = 1; j < WORDS.size(); j++) {
+      for (int j = 0; j < WORDS.size(); j++) {
         String word1 = WORDS.get(i);
         String word2 = WORDS.get(j);
         String word3 = "replacement";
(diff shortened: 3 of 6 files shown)


src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
```diff
@@ -29,8 +29,8 @@ class CVC4StringFormulaManager extends AbstractStringFormulaManager<Expr, Type,
 
   @Override
   protected Expr makeStringImpl(String pValue) {
-    // The boolean enables escape characters!
-    return exprManager.mkConst(new CVC4String(escapeUnicodeForSmtlib(pValue), true));
+    String str = escapeUnicodeForSmtlib(unescapeUnicodeForSmtlib(pValue));
```
kfriedberger, line 32: I would assume that `escape` and `unescape` are exactly reversed operations. This line is then a no-op and not required.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5StringFormulaManager.java
```diff
@@ -27,7 +27,8 @@ class CVC5StringFormulaManager extends AbstractStringFormulaManager<Term, Sort,
 
   @Override
   protected Term makeStringImpl(String pValue) {
-    return solver.mkString(escapeUnicodeForSmtlib(pValue), true);
+    String str = escapeUnicodeForSmtlib(unescapeUnicodeForSmtlib(pValue));
```
kfriedberger, line 30: I would assume that `escape` and `unescape` are exactly reversed operations. This line is then a no-op and not required.

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
```diff
@@ -1617,15 +1640,16 @@ public void testConstStringReplaceAll() throws SolverException, InterruptedExcep
...
         String word1 = WORDS.get(i);
         String word2 = WORDS.get(j);
         String word3 = "replacement";
         StringFormula word1F = smgr.makeString(word1);
         StringFormula word2F = smgr.makeString(word2);
         StringFormula word3F = smgr.makeString(word3);
 
-        StringFormula result = smgr.makeString(word3.replaceAll(word2, word1));
+        StringFormula result =
+            smgr.makeString(word2.isEmpty() ? word3 : word3.replaceAll(word2, word1));
```
kfriedberger, line 1652: what does this test do? If word2 is empty, nothing happens?

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
```diff
@@ -121,6 +124,26 @@ private void requireVariableStringLiterals() {
 
   // Tests
 
+  @Test
+  public void testInputEscape() throws SolverException, InterruptedException {
+    // Test if SMTLIB Unicode literals are recognized and converted to their Unicode characters.
+    assertEqual(smgr.length(smgr.makeString("Ξ")), imgr.makeNumber(1));
+    assertEqual(smgr.length(smgr.makeString("\\u{39E}")), imgr.makeNumber(1));
```
kfriedberger, line 131: Critical question: The documentation (https://github.com/sosy-lab/java-smt/blob/master/src/org/sosy_lab/java_smt/api/StringFormulaManager.java#L22-L34) says that escaped strings are recognized. Additionally, the Java-supported Unicode range (0x00000-0x10FFFF) is larger than SMTLIB-supported range (0x00000-0x2FFFF). 
Are we sure, we want to implement in this way, e.g, we already use Java-Strings? Why would a user need to use escaping in Java? Wouldn't it be nicer to ignore escaping and see a string like `\\u{1234}` as 6 individual chars?

kfriedberger (commented):
I am undecided whether our String escaping is useful.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
daniel-raffler: That's true for a single character (or a single escape sequence), but here we might have "mixed" Strings where some characters are escaped and others aren't. Take `"\u{4321}\n"` for instance, where the first character is already written as an escape sequence, but the newline following it isn't. By applying `unescape` we get `"䌡\n"` and then `escape` will give us the properly escaped String `"\u{4321}\u{a}"`.

(Just using `escape` here no longer works since we're now also escaping `\` to prevent it from capturing an escape sequence when unescaping back from the model)

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
daniel-raffler: Yes, this is as by the definition from SMTLIB:
```
; (str.replace_all s t t’) is s if t is the empty string. Otherwise, it
; is the string obtained from s by replacing all occurrences of t in s
; by t’, starting with the first occurrence and proceeding in
; left-to-right order. 
```
With `String.replaceAll` the result will be different for the empty String:
```
"abc.replaceAll("", "_") == "_a_b_c_"
```

kfriedberger: thanks for looking it up.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
kfriedberger: we could handle `\u{4321}\n` as String with 8 chars like `['\','u','{','4','3','2','1','}','\n']` and simply not handle it as escaped unicode char? what happens if a user would want to use it that way, e.g. like `"\u{43" + "21" + "}\n"`?

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
daniel-raffler: I'd prefer to just have Java Strings and only use `escape`/`unescape` when communicating with the SMT solver. One downside is interoperability as some users might expect SMTLIB escape sequences to work. Maybe there is a way to add `escape`/`unescape` to the `StringFormulaManager` so that users can convert their own Strings when needed?

The other problem is that this is an API breaking change and that we would have to fix quite a few of the tests in `StringFormulaManager` that expect SMTLIB escape sequences to work.

kfriedberger: There was no release that JavaSMT officially supports unicode strings. The API change happens mostly on developer stage. Also, changing test-code should not be an issue.

Perhaps we could get some third opinion on this topic?

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
daniel-raffler: > what happens if a user would want to use it that way, e.g. like "\u{43" + "21" + "}\n"?

They would have to escape the String themselves and use `StringFormulaManager.makeString("\u{5c}u{4321}\n")`. However, I agree that this can be quite confusing when assembling several Strings. For instance `StringFormulaManager.length(StringFormulaManager.makeString("\u{a"))` is `4` and `StringFormulaManager.length(StringFormulaManager.makeString("}"))`is `1`, but  `StringFormulaManager.length(StringFormulaManager.makeString("\u{a" + "}"))` will give you `1` again.

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
Avgor46: Hi! As an active user of java-smt lib, I would like to notice that treating all symbols in string as unique chars is more preferable for me. I do not want to consider such difference between Java and SMTLIB strings. In my code strings are generated almost automatically, the behavior like in the code above would be surprising and require manual effort to process. If I really want to treat `\\u{1234}` as a symbol, I just use `escape/unescape` methods. I think this is the right way)

baierd:
@daniel-raffler whats the current state of this PR?

daniel-raffler:
> @daniel-raffler whats the current state of this PR?

This is an API breaking change and I think we still have to decide how to handle it

The issue is that the format for the String argument in `StringFormulaManager.makeString` has never really been well defined. In the SMTLIB standard the Strings are Unicode, but a special encoding has to be used where all non-ASCII characters are written as `\uXXXX` escape sequences. One the other hand the argument for `makeString` is a Java String, so one would expect Unicode characters to be allowed. It's possible to translate between the two representations, but we have to know what "kind" of a String we have

Before [422](https://github.com/sosy-lab/java-smt/pull/422), which wasn't merged until after the last stable release, we didn't really say which characters are allowed in the String. Some of the tests expect SMTLIB escape sequences to be recognized, but this was never officially documented. Currently both Unicode characters and SMTLIB escape sequences are allowed, but this can lead to some confusing situations for developers when a valid escape sequence is created "by accident"

The PR tries to sort this out by committing to Java Strings and removing support for SMTLIB escape sequences. It's still possible to handle these sequences, but developers have to do the conversion themselves by calling `unescapeUnicodeForSmtlib` and `escapeUnicodeForSmtlib` from `AbstractStringFormulaManager`. This shouldn't be too much of a problem as the extra step is only needed when reading SMTLIB from an external source, or trying to store the model in SMTLIB format. Most developers should therefore never have to worry about it

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5StringFormulaManager.java
kfriedberger: fixed

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
kfriedberger: fixed, JavaSMT uses plain unicode chars in its API.

src/org/sosy_lab/java_smt/test/StringFormulaManagerTest.java
kfriedberger: done. The API uses plain Unicode chars, as expected from a Java-based library.

PR #432, opened 2025-01-17
Solver independent interpolation

baierd:
This MR adds solver independent interpolation as an option to JavaSMT.
Currently there are 2 distinct techniques used, model based and uniform interpolants based on quantifier elimination.
The uniform interpolants can be generated in a forward and backwards way.

Tests have been extended to test the native interpolation together with the new independent interpolation.

This is currently a WIP MR and not yet finished.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
@@ -345,12 +345,12 @@ public class IndependentInterpolatingProverEnvironment<F, T>
         BooleanFormula itpBackwardQuantifierEliminated =
             qfmgr.eliminateQuantifiers(itpBackwardQuantified);
         // check if the quantifier has been eliminated properly
-        if (itpBackwardQuantifierEliminated.toString().contains(SYMBOL_QUANTIFIER_FORALL)) {
-          throw new SolverException(
-              "Quantifier-elimination failed. "
-                  + "The resulting interpolant still contains quantifiers.");
+        if (!itpBackwardQuantifierEliminated.toString().contains(SYMBOL_QUANTIFIER_FORALL)) {
+          return itpBackwardQuantifierEliminated;
         }
-        itpBackward = itpBackwardQuantifierEliminated;
+        throw new SolverException(
+            "Quantifier elimination failed. "
+                + "Resulting interpolant still contains quantifiers.");
       } catch (Exception e) {
         throw new UnsupportedOperationException(
             "Solver does not support quantifier-elimination (for this logic).", e);
@@ -384,12 +384,12 @@ public class IndependentInterpolatingProverEnvironment<F, T>
         BooleanFormula itpForwardQuantifierEliminated =
             qfmgr.eliminateQuantifiers(itpForwardQuantified);
         // check if the quantifier has been eliminated properly
-        if (itpForwardQuantifierEliminated.toString().equals(SYMBOL_QUANTIFIER_EXISTS)) {
-          throw new SolverException(
-              "Quantifier-elimination failed. "
-                  + "The resulting interpolant still contains quantifiers.");
+        if (!itpForwardQuantifierEliminated.toString().equals(SYMBOL_QUANTIFIER_EXISTS)) {
+          return itpForwardQuantifierEliminated;
         }
-        itpForward = itpForwardQuantifierEliminated;
+        throw new SolverException(
+            "Quantifier elimination failed. "
+                + "Resulting interpolant still contains quantifiers.");
       } catch (Exception e) {
         throw new UnsupportedOperationException(
             "Solver does not support quantifier-elimination (for this logic).", e);
--- a/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
@@ -17,7 +17,6 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -94,7 +93,6 @@ public class IndependentInterpolatingProverEnvironment<F, T> extends AbstractPro
     return getIndependentInterpolationStrategy(options, pFormulaManager) != null;
   }
 
-  @SuppressFBWarnings(value = "DLS_DEAD_LOCAL_STORE", justification = "unused")
   private static @Nullable ProverOptions getIndependentInterpolationStrategy(
       Set<ProverOptions> options, FormulaManager pFormulaManager) {
     List<ProverOptions> itpStrat = new ArrayList<>(options);
@@ -104,6 +102,7 @@ public class IndependentInterpolatingProverEnvironment<F, T> extends AbstractPro
     }
     Preconditions.checkState(itpStrat.size() == 1);
     try {
+      @SuppressWarnings("unused")
       QuantifiedFormulaManager dummy = pFormulaManager.getQuantifiedFormulaManager();
     } catch (UnsupportedOperationException e) {
       throw new UnsupportedOperationException(
--- a/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
@@ -140,15 +140,14 @@ public class IndependentInterpolatingProverEnvironment<TFormulaInfo, TType>
 
     ImmutableList<Formula> sharedVars = getSharedVars(varsOfA, varsOfB);
 
-    if (varsOfA.isEmpty() || varsOfB.isEmpty() || sharedVars.isEmpty()) {
-      return bmgr.makeTrue();
-    }
-
     if (formulasOfA.isEmpty()) {
       return bmgr.makeTrue();
     }
     if (formulasOfB.isEmpty()) {
       return bmgr.makeFalse();
+    }
+    if (varsOfA.isEmpty() || varsOfB.isEmpty() || sharedVars.isEmpty()) {
+      return bmgr.makeTrue();
     }
 
     Preconditions.checkNotNull(interpolationStrategy);
--- a/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
@@ -93,6 +93,7 @@ public class IndependentInterpolatingProverEnvironment<F, T> extends AbstractPro
     return getIndependentInterpolationStrategy(options, pFormulaManager) != null;
   }
 
+  @SuppressWarnings("unused")
   private static @Nullable ProverOptions getIndependentInterpolationStrategy(
       Set<ProverOptions> options, FormulaManager pFormulaManager) {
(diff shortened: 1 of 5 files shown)


baierd (commented):
Some small hints about the most urgent needed changes @stefaniesinner

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -0,0 +1,519 @@
...
+import org.sosy_lab.java_smt.api.ProverEnvironment;
+import org.sosy_lab.java_smt.api.QuantifiedFormulaManager;
+import org.sosy_lab.java_smt.api.SolverContext;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
+import org.sosy_lab.java_smt.api.SolverException;
+import org.sosy_lab.java_smt.api.UFManager;
+
+// The non-interpolating delegates usually return nothing useful as itp points, so we have to
+// track it here
+@SuppressWarnings("unused")
```
baierd, line 44: This is not needed anymore.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -0,0 +1,519 @@
...
+   *     otherwise.
+   */
+  public static boolean hasIndependentInterpolationStrategy(Set<ProverOptions> options) {
+    return getIndependentInterpolationStrategy(options) != null;
+  }
+
+  private static @Nullable ProverOptions getIndependentInterpolationStrategy(
+      Set<ProverOptions> options) {
+    List<ProverOptions> itpStrat = new ArrayList<>(options);
+    final Set<ProverOptions> allIndependentInterpolationStrategies =
```
baierd, line 96: This looks like a constant.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -0,0 +1,519 @@
...
+        ImmutableSet.of(
+            ProverOptions.GENERATE_MODEL_BASED_INTERPOLANTS,
+            ProverOptions.GENERATE_UNIFORM_BACKWARD_INTERPOLANTS,
+            ProverOptions.GENERATE_UNIFORM_FORWARD_INTERPOLANTS);
+    itpStrat.retainAll(allIndependentInterpolationStrategies);
+    if (itpStrat.isEmpty()) {
+      return null;
+    }
+    Preconditions.checkState(itpStrat.size() == 1);
```
baierd, line 105: Since we only allow exactly 1 solver independent interpolation option, we could fail here more gracefully with an `InvalidConfigurationException` explaining the problem.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -0,0 +1,548 @@
...
+   *
+   * @param itp The given interpolant to check if it satisfies the definition of Craig interpolants.
+   * @param formulasOfA The set of formulas A, combined into one {@link BooleanFormula}.
+   * @param formulasOfB The set of formulas B, combined into one {@link BooleanFormula}.
+   * @param varsOfA A list of all variables in formulas of A.
+   * @param varsOfB A list of all variables in formulas of B.
+   * @return {@code true}, if the given interpolant is a valid Craig interpolant, returns {@code
+   *     false} otherwise.
+   */
+  private boolean satisfiesInterpolationCriteria(
```
baierd, line 460: This method does not check the criteria of an Craig-Interpolant as stated.

The criteria are:

-  the implication φ− ⇒ ψ holds,
- the conjunction ψ ∧ φ+ is unsatisfiable, and
- ψ only contains symbols that occur in both φ− and φ+.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -127,17 +124,41 @@ public BooleanFormula getInterpolant(Collection<TFormulaInfo> pFormulasOfA)
...
+    // handle empty interpolation groups and trivial interpolants
+    if (formulasOfA.isEmpty()) {
+      conjugatedA =  bmgr.makeTrue();
+    }
+    if (formulasOfB.isEmpty()) {
+      conjugatedB = bmgr.makeTrue();
+    }
+
+    if ((bmgr.isTrue(conjugatedA) && bmgr.isTrue(conjugatedB))) {
+      return null;
```
baierd, line 157: You should never return `null` for interpolating. If you can`t determine a correct value you should throw an exception.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -121,10 +122,17 @@ public <T> void emptyInterpolationGroup() throws SolverException, InterruptedExc
...
-      BooleanFormula emptyA = prover.getInterpolant(ImmutableList.of());
-      assertThat(bmgr.isTrue(emptyA)).isTrue();
+      if (interpolationStrategy == null) {
+        BooleanFormula emptyB = prover.getInterpolant(ImmutableList.of(id1, id2));
+        assertThat(bmgr.isFalse(emptyB)).isTrue();
+        BooleanFormula emptyA = prover.getInterpolant(ImmutableList.of());
+        assertThat(bmgr.isTrue(emptyA)).isTrue();
+      } else {
+        BooleanFormula emptyB = prover.getInterpolant(ImmutableList.of(id1, id2));
+        assertThat(emptyB).isEqualTo(bmgr.makeTrue());
```
baierd: We already know that the expected result for an interpolant for A with id1 and id2 is `false`. So it is not possible that an interpolant `true` is valid as well.

In this case especially, the group B is empty. So:
(1) A -> itp holds. 
We know that A AND B is UNSAT. We know B is empty, so its SAT on its own. Therefore A must be UNSAT. So it follows that itp can be whatever here.

(2) itp AND B has to be UNSAT. 
Since we know B is empty, its SAT on its own. So itp has to make it UNSAT. itp also is not allowed to use any variables. So the only boolean that fulfills this is itp = `false`.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -266,17 +281,41 @@ public <T> void binaryBVInterpolation1() throws SolverException, InterruptedExce
   }
 
   private void requireTreeItp() {
+    requireInterpolation();
+    if (interpolationStrategy == null) {
+      assume()
+          .withMessage("Solver does not support tree-interpolation.")
```
baierd, line 286: You could create a `requireTreeInterpolation()` method that simply checks if an exception is thrown, similarly as for sequential interpolants or interpolation in general.
This goes for all occurrences of this in this class.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -305,34 +344,40 @@ public <T> void sequentialInterpolation() throws SolverException, InterruptedExc
...
+      stack.close();
+
+      checkItpSequence(ImmutableList.of(A, B, C, D), itps1);
+      checkItpSequence(ImmutableList.of(D, C, B, A), itps2);
+      checkItpSequence(ImmutableList.of(A, C, B, D), itps3);
+      checkItpSequence(ImmutableList.of(A, A, A, B, C, D, D), itps4);
+      checkItpSequence(ImmutableList.of(A, A, B, C, D, A, D), itps5);
+      checkItpSequence(ImmutableList.of(B, C, D, A, A, A, D), itps6);
+    } else {
+      assertThrows(UnsupportedOperationException.class,
```
baierd, line 369: This would imply that this is intented behavior. But we could implement sequential and/or tree interpolants. So i would simply put an assumption at the beginning of the method that does not run this test for solvers/options that don't support it. (see my comment about the `requireTreeInterpolation()` method) 
This goes for all occurrences of this in this class.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -415,15 +473,21 @@ public <T> void sequentialInterpolationWithOnePartition()
 
     // list of one partition
     List<T> partition = ImmutableList.of(TA, TB);
-    List<BooleanFormula> itps = stack.getSeqInterpolants(ImmutableList.of(partition));
-    assertThat(itps).isEmpty();
+    if (interpolationStrategy == null) {
+      List<BooleanFormula> itps = stack.getSeqInterpolants(ImmutableList.of(partition));
+      assertThat(itps).isEmpty();
+    } else {
+      assertThrows(UnsupportedOperationException.class,
```
baierd, line 479: Should never be reached if `requireIndependentSeqItp()` is correctly implemented.
This goes for all occurrences of this in this class.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -127,16 +127,34 @@ public BooleanFormula getInterpolant(Collection<TFormulaInfo> pFormulasOfA)
     BooleanFormula conjugatedA = bmgr.and(formulasOfA);
     BooleanFormula conjugatedB = bmgr.and(formulasOfB);
 
-    // handle empty interpolation groups and trivial interpolants
+    ProverEnvironment prover = getDistinctProver();
+    prover.push(conjugatedA);
```
baierd, line 131: Why are you checking SAT here?
We already know that the current state of `A AND B`  has to be UNSAT.
Either there are trivial interpolants that can't be covered by the implemented procedures, or the procedures will produce an interpolant.
And trivial interpolants can be found without ever invoking a SAT check.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
```diff
@@ -127,16 +127,34 @@ public BooleanFormula getInterpolant(Collection<TFormulaInfo> pFormulasOfA)
...
       conjugatedA =  bmgr.makeTrue();
     }
     if (formulasOfB.isEmpty()) {
       conjugatedB = bmgr.makeTrue();
     }
 
+    // handle trivial interpolants
     if ((bmgr.isTrue(conjugatedA) && bmgr.isTrue(conjugatedB))) {
-      return null;
+      throw new IllegalArgumentException("There exists no interpolant.");
```
baierd, line 157: While this is correct, we only allow the users to call `getInterpolant()`  if the last SAT check was UNSAT and the provers stack is unchanged. So this is not needed.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -282,31 +284,31 @@ public <T> void binaryBVInterpolation1() throws SolverException, InterruptedExce
...
+        .isAnyOf(ProverOptions.GENERATE_MODEL_BASED_INTERPOLANTS,
+            ProverOptions.GENERATE_UNIFORM_FORWARD_INTERPOLANTS,
+            ProverOptions.GENERATE_UNIFORM_BACKWARD_INTERPOLANTS);
   }
 
-  private void requireIndependentSeqItp() {
+  private void requireSequentialInterpolation() {
     requireInterpolation();
+    assume()
+        .withMessage("Solver does not support tree-interpolation.")
```
baierd, line 303: Either the text is wrong, or you are confusing tree and sequential interpolants.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingProverEnvironment.java
baierd: I just noticed that you change `conjugatedA` and `conjugatedB` above, making this check wrong based on the wrong assignments above.

baierd:
Due to the unnecessary size of the history in this branch/PR, we close this PR in favor of a new PR/branch.

PR #440, opened 2025-02-25
FP theory: improve documentation and tests.

kfriedberger:


src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -345,36 +353,118 @@ FloatingPointFormula multiply(
...
+   * method is the same as the method <code>equal</code> for other theories.
    */
   BooleanFormula assignment(FloatingPointFormula number1, FloatingPointFormula number2);
 
   /**
    * Create a term for comparing the equality of two floating-point terms, according to standard
-   * floating-point semantics (i.e., NaN != NaN). Be careful to not use this method when you really
-   * need {@link #assignment(FloatingPointFormula, FloatingPointFormula)}.
+   * floating-point semantics (i.e., <code>NaN != NaN</code> and <code>0.0 == -0.0</code>). Be
+   * careful to not use this method when you really need {@link #assignment(FloatingPointFormula,
```
baierd, line 365: Maybe an example that shows the difference of this method and `assignment()` would be nice.

baierd (changes requested):
We also could add the information about the sign bit (is it part of mantissa or exponent), as well as the FP standard (IEEE 754) to the constant and variable creating methods and/or the type.

baierd:
I made the FP->BV support check adaptive. The `assume()` has an ugly check in it, but i didn't find something nicer. Maybe @kfriedberger you have a preference?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
kfriedberger: The example is in the documentation:
- assignment: `NaN ==/!= NaN` and `0.0 != -0.0`
- FP-equality: `NaN != NaN` and `0.0 == -0.0`

baierd:
There is no problem with the non-adaptive solution. I am just lazy and don't want to change it in the future if possible ;D

kfriedberger:
Adaptive tests are quite dangerous. You might not detect changes, if a solver suddenly support or no longer support a feature.

baierd:
True. Shall i change it back?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: Perfect!

baierd (approved):

kfriedberger:
Lets keep it as is.

PR #442, opened 2025-03-02
Extend Docker images and CI

kfriedberger:
With the support of more architectures we should also extend our test-coverage for those architectures.
This PR provides several smaller steps towards this:
- we add a new ant-based test-step that is smaller, faster, and checks only general solver features, like loading/closing a context or prover, and creating formula managers,
- we provide the initial technical foundation for aarm64-based tests, i.e., a docker image that is pushed into Gitlab registry and is used for tests.

Screenshot of the pipeline:
- left column: building Docker image for x64 and arm64 (will only run when scheduled, i.e., once per week)
- right column: unit-tests on x64 (full tests) and arm64 (quick tests)
image

This PR violates the rule to reuse SoSy-Lab common project files for building and testing applications and libraries. However, this is due to the lack of support for multiple architectures from the common project template. We opened an issue there: https://gitlab.com/sosy-lab/software/java-project-template/-/issues/7 and aim for a merge of scripts into the common project template.


build/gitlab-ci.Dockerfile.jdk-11
```diff
@@ -14,6 +14,22 @@
...
+  curl \
+  git \
+  jq \
+  openjdk-11-jre-headless \
+  openjdk-11-jdk-headless \
+  wget \
+  && rm -rf /var/lib/apt/lists/*
+
+ENV LANG C.UTF-8
+
```
daniel-raffler, line 33: Are the changes here just temporary due to the `java-project-template` issue you mentioned? Otherwise, I think that something like `FROM --platform=?? registry.gitlab.com/sosy-lab/software/java-project-template/test:jdk-11` could be used to pull the image for a specific architecture.

build/build-junit.xml
```diff
@@ -69,6 +69,41 @@ SPDX-License-Identifier: Apache-2.0
...
+            <include name="**/solvers/mathsat5/Mathsat5OptimizationNativeApiTest.*"/>
+            <include name="**/solvers/opensmt/OpenSmtNativeAPITest.*"/>
+            <include name="**/solvers/yices2/Yices2NativeApiTest.*"/>
+            <include name="**/test/FloatingPointNumberTest.*"/>
+            <include name="**/test/FormulaManagerTest.*"/>
+            <include name="**/test/NumeralFormulaManagerTest.*"/>
+            <include name="**/test/ProverEnvironmentTest.*"/>
+            <include name="**/test/RationalFormulaManagerTest.*"/>
+            <include name="**/test/SolverContextFactoryTest.*"/>
+            <include name="**/test/SolverContextTest.*"/>
```
daniel-raffler, line 95: I think `FloatingPointNumberTest` is not architecture dependent and could probably be removed. We might also not need the native tests for `CVC4` and `Boolector` anymore as both are deprecated and have successors now.

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
```diff
@@ -82,46 +86,51 @@ private void requireNativeLibrary() {
...
+  }
+
+  private void requireNoSupportedOperatingSystem() {
+    assume()
+        .withMessage(
+            "Solver %s is not yet supported in the operating system %s with architecture %s",
+            solverToUse(), OS, ARCH)
+        .that(isSupportedOperatingSystemAndArchitecture())
+        .isFalse();
+  }
```
daniel-raffler, line 104: Could we rename this to something like `requirePlatformNotSupported`? And `requirePlatformSupported` for the other case? The current name seems a bit confusing to me.

daniel-raffler (approved):
We might want to wait for feedback on the [java-project-template](https://gitlab.com/sosy-lab/software/java-project-template/-/issues/7) issue. Otherwise this looks good to me!

kfriedberger:
> We might want to wait for feedback on the [java-project-template](https://gitlab.com/sosy-lab/software/java-project-template/-/issues/7) issue. Otherwise this looks good to me!

I am unsure who will work on issues for `java-project-template` and when. JavaSMT is short for tests on arm64 now.

build/build-junit.xml
kfriedberger: The list of tests was chosen, because they are faster than about 3s each. I did not select by feature, but by performance.

src/org/sosy_lab/java_smt/test/SolverContextFactoryTest.java
kfriedberger: done

daniel-raffler: Thanks!

PR #445, opened 2025-03-03
Extend mathsat wrapper with more parse functions

baierd:
This PR extends the MathSAT5 wrapper with the function `msat_to_smtlib2_ext()` and fixes an error for function `msat_to_smtlib2_term()`, making it usable.

This also adds the JNI, JavaDoc and native tests.

@kfriedberger the normal SMTLib2 dump function currently used in the `FormulaManager` is unable to dump quantified formulas.
(Note: MathSAT5 can create quantified formulas, but not solve them. However, they can be dumped with the 2 methods above.)
Since we don't support them yet, we don't need to change anything. However, once we support quantification (see #297 ), we should discuss if we want to switch to one of the methods above for dumping (or add an option to switch). See !444.


baierd:
Note: i tested the 2 functions locally with a local compilation of the new wrapper. We still need to publish MathSAT5 with the changes publicly.

baierd:
The branch was built on top of a past commit on purpose, so that we can merge this into #297, as well as master.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractNativeApiTest.java
```diff
@@ -119,4 +125,48 @@ public void modularCongruence()
...
+
+    long type = msat_get_integer_type(env);
+    long xFun = msat_declare_function(env, "x", type);
+    long x = msat_make_constant(env, xFun);
+    long one = msat_make_int_number(env, 1);
+    // x = 1, x unbound
+    long body = msat_make_equal(env, x, one);
+    // Make bound x and substitute
+    long boundX = msat_make_variable(env, "x", type);
+    long substBody =   msat_apply_substitution(env, body, 1, new long[]{x}, new long[]{boundX});
```
kfriedberger, line 143: whitespace in line? apply formatter.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractNativeApiTest.java
```diff
@@ -119,4 +125,48 @@ public void modularCongruence()
     assertThat(msat_check_sat(env)).isFalse(); // 4 != 45 mod 42
     msat_pop_backtrack_point(env);
   }
+
+  // msat_to_smtlib2() can not export quantified formulas, use msat_to_smtlib2_ext() instead
+  @Test
+  public void quantifierToSmtlib2() {
+    String expectedSMTLib2Formula = "Bool (forall ((x Int)) (.def_14 x))";
```
kfriedberger, line 132: is this valid SMTLIB? The "Bool" at start looks strange.

kfriedberger (commented):
overall approved, but:
- fix formatting.
- MathSAT5 needs to be published before merging this PR.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractNativeApiTest.java
baierd: I check against parts of the output. The Bool is the return type of the formula here.

I added the full output as a comment + reasoning why i don't simply check against the full output.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractNativeApiTest.java
baierd: My bad. Done.

PR #447, opened 2025-03-06
Add Proper Yices2 Quantifier Support

baierd:
This PR adds proper support for quantified formulas in Yices2.
Proper substitution of unbound with bound variables was added.
The `QuantifiedFormulaManager` is now available publicly.
Also, the quantifier tests were refactored, so that the automatic setup does not block most tests.


baierd:
Note: this PR was created on 5.0.1, so that we can merge it into older feature branches.

baierd:
Note: i could not find a canonical way of discerning the quantifier types from a quantified term. I currently use the string representation, which is not good. Maybe any one of you @kfriedberger or @daniel-raffler wants to try his luck?

Note: you can get the type, but thats always `Bool`. You can also get the constructor (function def), but that is `14`, which means `YICES_FORALL_TERM`, but bundles both types. So you can think of it as `YICES_QUANTIFIED_TERM` instead.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
```diff
@@ -246,6 +259,60 @@ public FormulaType<?> getFormulaType(Integer pFormula) {
...
+          return termFromName;
+        }
+      } else {
+        throw new IllegalArgumentException(
+            String.format(
+                "Can't create variable with name '%s' and type '%s' "
+                    + "as it would omit a variable with type '%s'",
+                name, yices_type_to_string(type), yices_type_to_string(termFromNameType)));
+      }
+    }
```
kfriedberger, line 301: it looks like some ELSE-cases are missing. Is this intentional?

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
```diff
@@ -258,6 +325,22 @@ public <R> R visit(FormulaVisitor<R> pVisitor, Formula pFormula, Integer pF) {
...
+      case YICES_FORALL_TERM:
+        int children = yices_term_num_children(pF);
+        ImmutableList.Builder<Formula> boundVars = ImmutableList.builder();
+        for (int i = 0; i < children - 1; i++) {
+          int boundVar = yices_term_child(pF, i);
+          boundVars.add(encapsulateWithTypeOf(boundVar));
+        }
+        BooleanFormula body = encapsulateBoolean(yices_term_child(pF, children - 1));
+        Quantifier quant = Quantifier.EXISTS;
+        if (yices_term_to_string(pF).startsWith("(forall")) {
```
kfriedberger, line 339: Yices is open source. We could find out, how Yices computes the string and check for the flag that tells the difference from FORALL to EXISTS:

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApiTest.java
```diff
@@ -201,6 +204,8 @@ public void rationalError() {
     System.out.println(rat); // "use" variable
   }
 
+  // TODO: what is this test supposed to be doing? And remove print.
+  @Ignore
```
kfriedberger, line 208: Most of those tests are simply tests for loading and running Yices. They serves as examples and were helpful for debugging initial issues. If the tests do no longer bring benefit, please remove them, and do not just disable them.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApiTest.java
```diff
@@ -580,4 +593,27 @@ public void bvMul() {
...
+            String.format(
+                "Can't create variable with name '%s' and type '%s' "
+                    + "as it would omit a variable with type '%s'",
+                name, yices_type_to_string(type), yices_type_to_string(termFromNameType)));
+      }
+    }
+    int var = yices_new_uninterpreted_term(type);
+    yices_set_term_name(var, name);
+    return var;
+  }
```
kfriedberger, line 618: This IF-THEN-ELSE logic seems to be used twice. Can we refactor it into a proper utility method?

kfriedberger (approved):
overall a valid PR. some minor comments.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
daniel-raffler: Yices does not seem to have a separate term for `exists`:
```
term_t mk_exists(term_manager_t *manager, uint32_t n, const term_t var[], term_t body) {
  if (body == true_term) return body;
  if (body == false_term) return body;

  // (not (forall ... (not body))
  return opposite_term(forall_term(manager->terms, n, var, opposite_term(body)));
}
```
To see if we have a `forall` or an `exists` we'd have to check the polarity of the term. Yices does this internally with a simple bitmask:
```
static inline uint32_t polarity_of(term_t x) {
  return ((uint32_t) x) & 1;
}
```
I couldn't find any public API for this, but we could probably just keep track of the polarity in the visitor.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
baierd: This is intentional. I added documentation for it.

Essentially: Names work like a stack in Yices2. If we associate a term with a name, we get that term if we ask yices_get_term_by_name(). However, if we create bound variables, we associate them with the same name as the free variable (so that it has the same name). This pushes the name stack, and we get the bound var when asking yices_get_term_by_name(). That's why i added a cache for free vars. So this way we can re-use bound variables, and don't clash with free vars.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
baierd: No actually that is the solution!
`exists` is negated and we can check that.

This opens another questions however: do we traverse and/or rebuild terms correctly with this?
Do we have a test for `exists` that manipulates the formula? If not, we need to add one!

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApiTest.java
baierd: I reworked all tests with printouts into assertions and re-enabled all. I also found some issues and fixed them, added documentation and 2 more native tests.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApiTest.java
baierd: Only parts of the structure is reused. This version is for tests and for free variables only. The other one is for bound variables and has additional logic. (I reworked the creation of free vars and its not similar to this one anymore)

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
baierd: (I fear that the visitor might rebuild a formula `exists x x=y` into `not(forall x x=y)`, which is not equal.

baierd: Exactly as i expected, this happens. I added a test that should work for all solvers that checks this for both types of quantifiers.

baierd: OK so Yices2 does simply transform the formula on-the-fly while traversing it.
Example: If we traverse the formula `EXISTS x . OR(x, y)` we first get a negation and the child of that is `FORALL x . NOT(NOT x, NOT y)`, which is correct of course, but it still transforms the formula unexpectetly.

Do we want to accept this behavior @kfriedberger ?
It does lead to unexpected behavior when trying to substitute quantified formulas.
But also, it would be hard for us to circumvent this.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApi.java
```diff
@@ -610,7 +610,15 @@ private Yices2NativeApi() {}
   public static native int[] yices_bvsum_component(int t, int i, int bitsize);
 
   // TODO can return up to UINT32_MAX ?
-  /** Returns an array in the form [term,power]. */
+  //  Daniel: we cast the uint return of exp to signed int (jint), this is obviously wrong!
```
kfriedberger, line 613: Can someone give details and precise example for this comment?

kfriedberger:
@baierd 
I updated this branch to include changes from current master branch,
then I fixed some handling of quantifiers in Yices2 and Z3, such that:
- Yices2 handles and visits `NOT(FORALL(...))` as `EXISTS`.
- Yices2 and Z3 handle quantifiers and replace bound variables when visiting a quantified body formula.

From my side, this could be merged to master branch.
Are there any more changes required?

PR #448, opened 2025-03-06
Remove Boolector Quantifiers

baierd:
Boolectors quantifier support never worked in JavaSMT and never will without changing our public API.
The reason is that it needs bound variables, but Boolector does not allow formula manipulation.
Boolector is also in maintenance mode and does not receive any updates.

This PR removes Boolectors quantifiers.


baierd:
This PR depends on #447 (due to test changes).
This PR is also based on 5.0.1, as it is meant to be merged into feature branches as well.

baierd:
Closed in favor of #449

PR #449, opened 2025-03-07
Remove boolector quantifiers standalone

baierd:
Boolectors quantifier support never worked in JavaSMT and never will without changing our public API.
The reason is that it needs bound variables, but Boolector does not allow formula manipulation.
Boolector is also in maintenance mode and does not receive any updates.

This PR removes Boolectors quantifiers.
(Supersedes #448)


kfriedberger (approved):
lgtm. nice cleanup of an unused/unusable feature.

Issue #392, opened 2024-09-13
Unwanted output in tests (and probably outside of it, too)


PhilippWendler:
When running tests, such as in [this case](https://gitlab.com/sosy-lab/software/java-smt/-/jobs/7816950900), some of the tests seem to print warnings like the following:
- `Unexpected logic for quantifier elimination ALL`
- `WARNING: optimization with quantified constraints is not supported`

I am not sure how to interpret this, does this indicate a bug?

Anyway, given that these strings do not appear in the source of JavaSMT, I assume that this outputs will also appear in certain situations during regular usage of JavaSMT (outside of tests). However, JavaSMT and the solvers are libraries, they should never write to stdout or stderr unless explicitly told so by the calling application. Any output is likely to mess up the desired output of the application. Libraries should use logging, or, if the message indicate bugs, communicate this via exceptions.

kfriedberger:
Related issue: https://github.com/uuverifiers/ostrich/issues/95

PR #451, opened 2025-03-08
Princess: redirect 'System.err' to avoid cluttering the output.

kfriedberger:
This is a first step for https://github.com/sosy-lab/java-smt/issues/392 .
The situation is also reported to Ostrich in https://github.com/uuverifiers/ostrich/issues/95 .

Questions to the reviewers:
- Do we want to redirect StdErr for the whole application whenever Ostrich is used?
- Or should we wait for the fix from Ostrich itself?

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
@@ -20,6 +20,10 @@ import ap.parser.IFunction;
 import ap.parser.ITerm;
 import com.google.common.base.Preconditions;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -29,6 +33,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Supplier;
 import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.common.UniqueIdGenerator;
 import org.sosy_lab.common.collect.PathCopyingPersistentTreeMap;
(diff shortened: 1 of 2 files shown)


src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
```diff
@@ -278,4 +284,34 @@ public String toString() {
...
+      System.setErr(nullStream);
+      return fn.get();
+    } catch (IOException ioException) {
+      throw new AssertionError(ioException);
+    } finally {
+      System.setErr(originalErrStream);
+    }
+  }
+
+  private static final class NullOutputStream extends OutputStream {
```
PhilippWendler, line 311: There is `ByteStreams.nullOutputStream()` for this.

PhilippWendler (commented):
There are two (potential) problems with this approach:

### Thread-safety

If two threads attempt to use Princess at the same time and the order of events is "thread A starts operation, thread B starts operation, thread A ends operation, thread B ends operation", thread B will set `System.err` back to the dummy stream that A created, and `System.err` will from now on continue to have the wrong value. This means that suddenly all error output from the whole program vanishes. I imagine that for a developer of a program that uses JavaSMT+Princess, this would be a horrible situation to be in while debugging, if non-deterministically the error output just is not working and the program crashes silently.

Adding a mutex to `callWithoutSystemErrStream` would make it only safe with regards to concurrent calls to Princess, but not safe with regards to other code changing `System.err`. And it would prevent actually concurrent usage of Princess.

## Performance

`System.out` and `System.err` are to some degree a design mistake from Java's past, and even have to have a special hard-coded exception in the Java memory model and special treatment by the compiler/JVM (cf. JLS 17.5.4). Most programs do not change these values at all, or only rarely (like once at the beginning). Taken together I would expect the JVM to be optimized for the common case of these fields not changing, and it could be that there is an unexpectedly high (i.e., higher that just a usual non-synchronized field assignment) performance cost related to changing them. Given that we change them twice on every operation, I would say we should evaluate the performance cost before adding this code.

---

Taking together, I am skeptical whether the advantages are worth the problems. A better solution would be if Princess simply stops using `System.err`. Is there a chance they would do this?

kfriedberger:
On second thought, this code should be part of JavaSMT. Lets wait for a fix in Ostrich: https://github.com/uuverifiers/ostrich/issues/95

PR #453, opened 2025-03-17
Fix library symlinks

daniel-raffler:
Hello,
I've noticed that there is a broken symlink in `lib/native/arm64-linux`. The README file for `lib/native/x86-64-windows` is also outdated, and we're missing instructions to link the binaries for windows on arm64. This PR should take care of all these minor issues.

### DIFF
--- a/lib/native/arm64-linux/libopensmtjava.so
+++ b/lib/native/arm64-linux/libopensmtjava.so
@@ -1,1 +1,0 @@
-../../java/runtime-opensmt/arm64/libopensmtjava.so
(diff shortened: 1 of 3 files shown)


kfriedberger (approved):
Thanks for this cleanup and update of documentation.

The escaped backslashes do no longer allow a simple copy&paste of the command by a user. Could you rewrite those lines with proper markup for `code` (single backtick around command) to avoid this escaping?

The rest looks good. Approved.

kfriedberger (approved):
lgtm. can be merged.

Issue #439, opened 2025-02-22
Z3: Regression in String theory with update to 4.14.0


kfriedberger:
See ca2f0e8aefdada71264f2ff301c4f2a744ec0f1c for details: A unit test for String theory and String replacement runs into timeout, Linux and Windows.

The unit test was actually finetuned to do not run into timeout.
We can either optimize the test or ask Z3 for a fix.

![Image](https://github.com/user-attachments/assets/7c0bfd07-504b-401d-b1b3-f3de4a920654)

baierd:
The problem is in the first SAT check beginning in line `1441`. The other checks run smoothly. I will extract the problem into Z3 and report it.

baierd:
SMTLib2 terminates immediately for the Z3 4.14.0 binary. We need a API based program.
```
(set-logic ALL)
(define-fun original () String "original")
(define-fun prefix () String "prefix")
(define-fun replacement () String "replacement")
(define-fun replaced () String "replaced")
(assert (or (not (and
      (str.prefixof prefix original) 
      (=
        (str.len prefix) (str.indexof original (str.substr original (str.len prefix) (str.len original))
                            0))
                    (>= (str.len original) (str.len prefix))
                    (>= (str.len prefix) 0)
                    (>= 
                        (str.len 
                            (str.substr original (str.len prefix) (str.len original)))
                        0)
                    (=
                        replaced
                        (str.replace 
                            original
                            (str.substr original (str.len prefix) (str.len original))
                            replacement)))) (= replacement (str.substr replaced (str.len prefix) (str.len replaced)))))
(check-sat)
(get-model)
(exit)
```

baierd:
The SMTLib2 was incorrect. I fixed it. But it still terminates immediately.

baierd:
Using the Z3 API (by either parsing the output from the tests formula or building the formula by the API) also terminates immediatly.

Program (either use JavaSMT to load Z3 or edit the path in the commented out section):
```java
import com.microsoft.z3.BoolExpr;
import com.microsoft.z3.CharSort;
import com.microsoft.z3.Context;
import com.microsoft.z3.Expr;
import com.microsoft.z3.IntSort;
import com.microsoft.z3.Model;
import com.microsoft.z3.Native;
import com.microsoft.z3.SeqSort;
import com.microsoft.z3.Solver;
import com.microsoft.z3.Sort;
import com.microsoft.z3.Status;
import java.util.HashMap;
[...]

PR #455, opened 2025-03-17
Reorder test formula in testStringVariableReplaceSubstring() to fix Z3 regression

baierd:
A Z3 regression in a String test is resolved by reordering the formulas subexpressions (without changing them).

I added a TODO so that we can take a look at this with a new Z3 version. I also suggest to not close the issue until then.

See #439 for more info.


kfriedberger (approved):
lgtm. can be merged.

PR #456, opened 2025-03-18
Allow empty String ranges

daniel-raffler:
Hello,
I've noticed that there is a precondition the `char` override for `StringFormulaManager.range`:
```
* @return formula denoting the range regular expression over two chars.
* @see #range(StringFormula, StringFormula)
*/
default RegexFormula range(char start, char end) {
  Preconditions.checkArgument(
      start <= end,
      "Range from start '%s' (%s) to end '%s' (%s) is empty.",
      start,
      (int) start,
      end,
      (int) end);
  return range(makeString(String.valueOf(start)), makeString(String.valueOf(end)));
}
```
However, the JavaDoc for the (general) `range` method does not mention this restriction:
```
/**
 * @return formula denoting the range regular expression over two sequences of length 1.
 */
RegexFormula range(StringFormula start, StringFormula end);
```
And, according to SMTLIB, empty ranges should be fine:
```
; RE range
; (re.range s₁ s₂) is the set of all *singleton* strings s such that
; (str.<= s₁ s s₂) provided s₁ and s₂ are singleton. Otherwise, it 
; is the empty language.
(re.range String String RegLan)
```
I've tried it out with different solvers, and only CVC4 seems to cause issues when the precondition in the `default` method is simply removed. CVC5 and Princess still assume that the bounds are single character Strings and will throw an exception otherwise. However, they can handle the case when the lower bound is larger than the upper bound and simply return the empty language. Z3 supports both cases, and will also return the empty language when one of the bounds is not a single character String.

In this PR I removed the precondition from the `default` override to make sure it matches the behavior of the general method. I also added some checks to prevent crashes on CVC5 and Princess if either of the bounds is not a single character Strings, and fixed two bugs for Princess. Note that there are still some issues on CVC4, but this will be hard to fix as the solver is no longer maintained.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
@@ -145,6 +145,7 @@ class CVC4StringFormulaManager extends AbstractStringFormulaManager<Expr, Type, 
 
   @Override
   protected Expr range(Expr start, Expr end) {
+    // FIXME CVC4 will crash when start > end, even if we add an ITE with preconditions
     return exprManager.mkExpr(Kind.REGEXP_RANGE, start, end);
   }
 
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
@@ -143,10 +143,33 @@ class CVC4StringFormulaManager extends AbstractStringFormulaManager<Expr, Type, 
     return exprManager.mkExpr(Kind.REGEXP_SIGMA, new vectorExpr());
   }
 
+  /**
+   * Check if the String only contains printable US ASCII characters.
+   *
+   * <p>We use this function to check if the String contains any characters that would have to be
+   * escaped in SMTLIB.
+   */
+  private boolean isAsciiString(String str) {
+    return str.codePoints().allMatch(c -> c >= 0x20 && c <= 0x7E);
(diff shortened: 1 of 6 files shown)


src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
```diff
@@ -145,6 +145,7 @@ protected Expr allCharImpl() {
 
   @Override
   protected Expr range(Expr start, Expr end) {
+    // FIXME CVC4 will crash when start > end, even if we add an ITE with preconditions
```
kfriedberger, line 148: CVC4 uses eager evaluation and simplifies constants upfront. This might explain the crash for the range `[c1,c2]` if `c1>c2`, even if this is nested in an expression like `ITE(c1<c2, [c1,c2], [])`. We could try to check for constants upfront, like `if (isConst(c1) && isConst(c2) && c1.value>c2.value) { return [] } else { return ITE(c1<c2, [c1,c2], []) }`.
If that does not help, ignore CVC4. We should mark CVC4 as `deprecated` sooner or later.

kfriedberger (approved):
It is good to match our API against the SMTLIB standard.
Looks good for me.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4StringFormulaManager.java
daniel-raffler: Thanks for the tip! I've added the checks for CVC4 in the last commit.

kfriedberger (approved):
lgtm.

PR #458, opened 2025-03-19
Develop a common proof format and export proofs

gcarpio21:
## Objective  
This PR aims to expose proofs produced by supported solvers through the API, primarily via a `getProof()` method.  

## Key Changes  
- Introduced `ProofNode`, `ProofDag`, and `ProofRule` as the main components for handling proofs.  
- Implemented proof retrieval for the **Z3 solver**.  
- Explored a standardized format for proofs across different solvers.  

## Proof Format Conversion  
To unify proof representation, this PR investigates converting solver-specific proofs into a **common format**, with **RESOLUTE** (used by SMTInterpol) as a candidate.  
- The proposed approach models proof rules in **CNF** and transforms them into **resolution steps**.  
- The `Z3ToResoluteProofConverter` provides an example of this transformation.  

## Task List  
- [x] Implement `getProof()` for Z3.  
- [x] Introduce `ProofNode`, `ProofDag`, and `ProofRule`.  
- [ ] Implement `getProof()` for other solvers.  
  - [x] Z3  
  - [ ] MathSAT5  
  - [ ] CVC5  
  - [ ] SMTInterpol  
  - [ ] Princess
  - [ ] Opensmt

- [ ] Test `getProof` more robustly.

## Next Steps
- More flexibility while handling proofs could be provided through the `ProofDag` interface.


src/org/sosy_lab/java_smt/solvers/z3/Z3NonRecursiveProofProcessor.java
```diff
@@ -0,0 +1,211 @@
...
+      case Z3_OP_PR_TH_LEMMA:
+        return Z3ProofRule.TH_LEMMA;
+      case Z3_OP_PR_HYPER_RESOLVE:
+        return Z3ProofRule.HYPER_RESOLVE;
+      default:
+        return Z3ProofRule.OPERATION;
+    }
+  }
+
+  private Formula generateFormula(long proof) {
```
gcarpio21, line 192: I wanted to check here if the `Formula` matches some already created `Formula` to return that instead of creating a new one by encapsulating. Any idea if this is possible? @baierd

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDag.java
```diff
@@ -0,0 +1,57 @@
...
+  private final Map<Integer, ProofNode> nodes = new HashMap<>();
+  private int nodeIdCounter = 0;
+
+  @Override
+  public void addNode(ProofNode node) {
+    nodes.put(nodeIdCounter++, node);
+  }
+
+  @Override
+  public ProofNode getNode(int nodeId) {
```
baierd, line 35: Is the argument really `nodeId` or rather `index`?

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolProofsTest.java
```diff
@@ -0,0 +1,247 @@
...
+    Configuration config = Configuration.defaultConfiguration();
+    LogManager logger = BasicLogManager.create(config);
+    ShutdownManager shutdown = ShutdownManager.create();
+
+    // Create new context with SMTInterpol
+    context = SmtInterpolSolverContext.create(
+        config,
+        logger,
+        shutdown.getNotifier(),
+        null, // no logfile
```
baierd, line 59: Why not use the `SolverContextFactory`?

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolProofsTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+      Clause proof = smtinterpol.retrieveProof();
+      ProofNode proofNode = proof.getProof();
+      Term term = smtinterpol.getProof();
+
+      assertThat(proof).isNotNull();
+
+
+      //String proofStr = proof.toString();
+      System.out.println(invokeGetProofMode(smtinterpol).toString());
```
baierd, line 144: Please use assertions instead of printouts.

src/org/sosy_lab/java_smt/solvers/z3/Z3ProofsTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+  @Before
+  public void setUpSolver() throws Exception {
+    Configuration config =
+        Configuration.builder().setOption("solver.z3.requireProofs", "true").build();
+    LogManager logger = BasicLogManager.create(config);
+    ShutdownManager shutdown = ShutdownManager.create();
+
+    // Create new context with SMTInterpol
+    context = Z3SolverContext.create(
```
baierd, line 51: Why not use the `SolverContextFactory`?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ProofsTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private Solver solver;
+
+  @Before
+  public void init() throws CVC5ApiException {
+    solver = createEnvironment();
+  }
+
+  private static Solver createEnvironment() throws CVC5ApiException {
+    Solver newSolver = new Solver();
```
baierd, line 41: This tests nothing currently. Also, native proof tests are good, but should be part of the CVC5 native test-class.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5NativeApiTest.java
```diff
@@ -79,6 +85,7 @@ public void createEnvironment() {
     long cfg = msat_create_config();
 
     msat_set_option_checked(cfg, "model_generation", "true");
+    msat_set_option_checked(cfg, "proof_generation", "true");
```
baierd, line 88: Why enable proofs for all tests and then build a new, dedicated solver environment for the proof test?

src/org/sosy_lab/java_smt/solvers/z3/Z3ToResoluteProofConverter.java
```diff
@@ -0,0 +1,616 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Converts a Z3 proof to a RESOLUTE proof.
+ */
+
+@SuppressWarnings({"unchecked", "rawtypes", "unused", "static-access","ModifiedButNotUsed"})
+public class Z3ToResoluteProofConverter {
```
baierd, line 39: This class tackles a hard problem and should be documented with the used ideas and possible sources behind it.

baierd (changes requested):
Some general things that i noticed (some may be mentioned in change requests, but apply generally):

- Tests should be generalized and moved into the test package if they are not using native API. You can still have tests for a single solver in the test package by using the `assume` API and only allowing a single one. The name of the solver should be mentioned in the test name then.

- Please add documentation to the new classes and interfaces. Especially mention which ones are used internally and which ones are public/usable by users and the purpose/background of the classes/interfaces.

- You can have classes within classes, which is especially helpful to structure smaller components, e.g. nodes of a graph, within the context you are using them in. That makes the code much easier to read and structures it better.

- Please avoid pushing print-outs or commented out code. If you are working on a feature thats not yet working, thats fine. Document it and don't use the component. Please add some documentation to the commented out code. Print-outs should never be pushed. Try using the logger or exceptions. In tests please use assertion instead of printouts. (It is OK to expect the current static output of a solver. If the test fails in the future, we devs are notified about changes that we might need to consider.)

- Native API tests that just test that proofs work are good and can be added for all solvers supporting proofs.

src/org/sosy_lab/java_smt/solvers/z3/Z3NonRecursiveProofProcessor.java
baierd: Don't bother. The solver optimizes this as far as possible.
(Some solvers don't cache Variables or UFs and we have to to it ourselves, but formulas are well optimized)

baierd:
@leventeBajczi could you spare a look at this PR and maybe give us some feedback on the current proof interface?

src/org/sosy_lab/java_smt/ResProofRule.java
```diff
@@ -0,0 +1,149 @@
...
+package org.sosy_lab.java_smt;
+
+import org.sosy_lab.java_smt.api.proofs.ProofRule;
+
+/**
+ * A proof rule in the proof DAG of the internal proof format of JavaSMT.
+ *
+ * @author Gabriel Carpio
+ */
+public class ResProofRule {
```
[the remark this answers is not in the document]
gcarpio21: I added some documentation and will wait for further information on the format form Mathsat5 as it seems to be very similar.

src/org/sosy_lab/java_smt/ResolutionProofDag.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.sosy_lab.java_smt;
+
+import org.sosy_lab.java_smt.basicimpl.AbstractProofDag;
+
+@SuppressWarnings("all")
+public class ResolutionProofDag extends AbstractProofDag {
+
+  /*
+  public ResolutionProofDAG() {
```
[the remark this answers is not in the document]
gcarpio21: Made some adjustments to this.

src/org/sosy_lab/java_smt/solvers/z3/Z3ToResoluteProofConverter.java
gcarpio21: I added further documentation

src/org/sosy_lab/java_smt/api/proofs/ProofFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.sosy_lab.java_smt.api.proofs;
+
+import org.sosy_lab.java_smt.ResolutionProofNode;
+import org.sosy_lab.java_smt.SourceProofNode;
+import org.sosy_lab.java_smt.api.Formula;
+import org.sosy_lab.java_smt.ResProofRule.ResAxiom;
+
+public class ProofFactory {
+  public static ProofNode createSourceNode(ResAxiom rule, Formula formula) {
```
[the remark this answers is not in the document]
gcarpio21: I'm currently trying to utilize this class to eliminate classes like `Mathsat5ProofProcessor` and `Z3ToResoluteProofConverter`. However, I'm facing an issue: ideally, the user shouldn't have direct access to these methods and should instead retrieve proofs via `getProof` from a `ProverEnvironment` instance. The challenge is that making the class non-public complicates its use within the solvers' packages.

gcarpio21:
> > @leventeBajczi could you spare a look at this PR and maybe give us some feedback on the current proof interface?
> 
> Definitely!
> 
> So far, I really like the relatively clean API for accessing the proofs, and I'm especially happy to see that the Z3 proofs are transformed to a 'simple' proof tree/DAG -- however, I'm uncertain how complete that transformation is based on the code, I think some information about that is missing (maybe as a comment?) to state how well Z3 proofs can be transformed into the common format.
> 
> Other than that, great work! I'm looking forward to integrating this with Theta. I'm really surprised at the number of solvers supporting proof generation, I only knew about MathSAT and Z3 beforehand. They will definitely be handy for us.
> 
> Have you maybe done some performance impact tests with the solvers when proof generation is enabled? I'd be interested to see how much performance is lost when it's active.

Right now, the Z3ToResoluteProofConverter class is still a work in progress. The goal is to transform Z3 proofs into a format similar to SMTInterpol’s (RESOLUTE), following the same proof rules.

Currently, the main functionality is to make Z3 proofs accessible through JavaSMT by simplifying Z3’s proof structures and implementing the ProofNode interface, which should capture all relevant proof information.

I haven’t run any performance tests yet. From what I know, enabling proof generation in a solver itself is usually quite costly. Hopefully, retrieving proofs through JavaSMT won’t introduce significant additional overhead.

I’ve also added more documentation in the code to clarify some of these points.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -149,6 +150,19 @@ default ImmutableMap<String, String> getStatistics() {
...
 
+  /**
+   * Get proof of unsatisfiability of the conjuction of the current satck of all formulas. Should
+   * only be called after {@link #isUnsat()} returned <code>true</code>.
+   */
+  default ProofNode getProof() {
+    try {
+      throw new UnsupportedOperationException("Proof generation isn't enabled.");
+    } catch (UnsupportedOperationException e) {
+      System.out.println(e.getMessage());
```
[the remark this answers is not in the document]
gcarpio21: To check the state I will move this method to `AbstractProver`. There, there also a method `checkGenerateProofs` that mimics the behavior of the other check methods, I understand this should throw an Exception already. Aside from this there is still the issue that for example for Z3 proof generation must be set as an `Option` in the configuration so this the solver should retrieve this information from the context in this case.

src/org/sosy_lab/java_smt/ResProofRule.java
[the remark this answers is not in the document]
baierd: As discussed, we will add this as far as possible and wait for more details from the MathSAT5 devs.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -243,6 +248,22 @@ public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
     throw new UnsupportedOperationException();
   }
 
+  @Override
+  public ProofNode getProof() {
+
+    Proof[] proofs = solver.getProof();
+    if (proofs == null || proofs.length == 0) {
+      throw new IllegalStateException("No proof available");
```
baierd, line 256: Why not return an empty proof?
Also, can this happen? E.g. trivial SAT calls?
Tests would be good for this.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -243,6 +248,22 @@ public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
...
+    }
+
+    CVC5ProofProcessor pp = new CVC5ProofProcessor(creator);
+    try {
+      return pp.fromCVC5Proof(proofs[0]);
+    } catch (CVC5ApiException pE) {
+      throw new RuntimeException(pE);
```
baierd, line 263: Re-throwing of exceptions should only happen for very good reasons, as they are much harder to work around than checked exceptions. A `ProverException` seems sensible here.

baierd (changes requested):
Good work in cleaning up the classes!

I added a MathSAT5 native test for the proof generation option and it seems that there is no problem with it. Do other solvers need to option when creating the context?

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDag.java
```diff
@@ -54,4 +60,57 @@ public Collection<ProofNode> getNodes() {
   public void accept(ProofVisitor visitor) {
     visitor.visitDAG(this);
   }
-}
\ No newline at end of file
+
+  public abstract static class AbstractProofNode implements ProofNode {
+    private final List<ProofNode> children;
+    private ProofRule rule;
+    protected Formula formula;
```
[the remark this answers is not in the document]
gcarpio21: I believe rule might be final. Formula I just changed recently because, in the case of MathSAT5, the formula for the res-chain nodes gets set after the conversion of the MathSAT5 proof into a JavaSMT proof

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
gcarpio21: This can happen. CVC5 throws the following exception when using the API: io.github.cvc5.CVC5ApiRecoverableException: cannot get proof unless in unsat mode

gcarpio21:
> Good work in cleaning up the classes!
> 
> I added a MathSAT5 native test for the proof generation option and it seems that there is no problem with it. Do other solvers need to option when creating the context?

Z3 needs the proof generation set when creating context

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5TheoremProver.java
```diff
@@ -43,4 +60,296 @@ protected Void addConstraintImpl(BooleanFormula constraint) throws InterruptedEx
...
+    msat_destroy_proof_manager(pm);
+
+    return pn;
+
+    // return getProof0();
+  }
+
+  // update all RES_CHAIN nodes in the proof DAG by computing resolution
+  // formulas and return the updated root node with formulas attached.
+  private void clausifyResChain(ProofNode root, BooleanFormulaManager bfmgr) {
```
gcarpio21, line 84: @baierd I added this method here to perform a resolution chain and be able to provide a 'Formula' for each node in the proof. I would like your opinion on this, since the terms are not really provided by MathSAT5 but calculated here.  I know the visitor can be implemented better than is right now.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
baierd: Then this is completely fine! Thanks!

src/org/sosy_lab/java_smt/solvers/z3/Z3ToResolutionProofConverter.java
```diff
@@ -400,143 +400,143 @@ ProofNode handleTransitivityStar(Z3ProofDag.Z3ProofNode node) {
...
   }
 
-  ProofNode handleIffOeq(Z3ProofDag.Z3ProofNode node) {
+  ProofNode handleIffOeq(Z3ProofDAG.Z3ProofNode node) {
     return null;
   }
 
-  ProofNode handleNnfPos(Z3ProofDag.Z3ProofNode node) {
+  ProofNode handleNnfPos(Z3ProofDAG.Z3ProofNode node) {
     return null;
```
baierd, line 512: It is generally not advised to return `null`. 
Just throw a `UnsupportedOperationException` then you will know which parts you still need to implement or inherit/override.

baierd (changes requested):
Good job on cleaning some of the classes up! There are still some open change-requests that need some work.

baierd:
> > Good work in cleaning up the classes!
> > I added a MathSAT5 native test for the proof generation option and it seems that there is no problem with it. Do other solvers need to option when creating the context?
> 
> Z3 needs the proof generation set when creating context

You can try to simply enable it per default currently in this branch. We can measure the performance impact and decide if we want this or not later.

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDag.java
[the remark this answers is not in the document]
baierd: Make sure to update your variable modifiers in the end to ensure the minimal visibility and mutability.

src/org/sosy_lab/java_smt/api/proofs/visitors/ProofTraversalVisitor.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.sosy_lab.java_smt.api.proofs.visitors;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.sosy_lab.java_smt.api.proofs.ProofDag;
+import org.sosy_lab.java_smt.api.proofs.ProofNode;
+
+public class ProofTraversalVisitor implements ProofVisitor {
```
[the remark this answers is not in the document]
gcarpio21: This wasn't fleshed out at the time and as of right now it could be useful for the resolution for the solvers that do not provide formulas at every step of the proof

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDAG.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public Formula getFormula() {
+      return formula;
+    }
+
+    @Override
+    public List<ProofNode> getChildren() {
+      return Collections.unmodifiableList(children);
```
baierd, line 86: Guavas immutable data-structures, e.g. `ImmutableList`, are preferable to Javas own implementation of immutable data-structures. You can use `ImmutableList.of()` for single/multiple items, or the builder `ImmutableList.builder().addAll().build()` for collections.

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDAG.java
```diff
@@ -0,0 +1,117 @@
...
+    return nodes.values();
+  }
+
+  @Override
+  public void accept(ProofVisitor visitor) {
+    visitor.visitDAG(this);
+  }
+
+  public abstract static class AbstractProofNode implements ProofNode {
+    private final List<ProofNode> children;
```
baierd, line 66: Always explicitly storing all children is redundant and most of the time not needed.
If the children are requested frequently we can cache them, but in general they should be computed lazily and every time the method is called. (which avoids updating problems)

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDAG.java
```diff
@@ -0,0 +1,117 @@
...
+    private ProofRule rule;
+    @Nullable protected Formula formula;
```
baierd, line 68: `Optional` is in general a better solution compared to `null`.

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDAG.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final UniqueIdGenerator idGenerator = new UniqueIdGenerator();
+    private final int id;
```
baierd, line 70: See my comment about the DAG implementation.

src/org/sosy_lab/java_smt/basicimpl/AbstractProofDAG.java
```diff
@@ -0,0 +1,117 @@
...
+    public boolean isLeaf() {
+      return children.isEmpty();
+    }
+
+    @Override
+    public int getId() {
+      return id;
+    }
+
+    void setRule(ProofRule rule) {
```
baierd, line 109: Is this truly needed? Or can this class be final, so that the rule/formula is only set when constructing the node?

src/org/sosy_lab/java_smt/delegate/logging/LoggingBasicProverEnvironment.java
```diff
@@ -120,6 +121,13 @@ public ImmutableMap<String, String> getStatistics() {
     return wrapped.getStatistics();
   }
 
+  @Override
+  public ProofNode getProof() throws SolverException, InterruptedException {
+    ProofNode p = wrapped.getProof();
+    logger.log(Level.FINE, "proof", p);
```
baierd, line 127: This would print the entire proof into the logger. You only need to log special events or errors, so you can remove this.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -243,6 +249,22 @@ public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
...
+  public ProofNode getProof() {
+
+    Proof[] proofs = solver.getProof();
+    if (proofs == null || proofs.length == 0) {
+      throw new IllegalStateException("No proof available");
+    }
+
+    // CVC5ProofProcessor pp = new CVC5ProofProcessor(creator);
+    try {
+      return CVC5ProofNode.fromCVC5Proof(proofs[0], creator);
```
baierd, line 262: Separation of concerns would dictate that `fromCVC5Proof` is not part of the `CVC5ProofNode`, but `CVC5Proof`. Also, the name could be more accurate, e.g. `generateCVC5Proof()` or even better a method like `generateProof()` that is available for all solvers. You can take a look at how this works in the Model for inspiration.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -243,6 +249,22 @@ public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
...
+    } catch (CVC5ApiException e) {
+      throw new RuntimeException(e);
```
baierd, line 264: Don't just throw `RuntimeException`s. This should be the `CVC5ApiException` wrapped in a `SolverException`, which is thrown in a checked way.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ProofDAG.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.Map;
+import org.sosy_lab.java_smt.api.Formula;
+import org.sosy_lab.java_smt.api.proofs.ProofFrame;
+import org.sosy_lab.java_smt.api.proofs.ProofNode;
+import org.sosy_lab.java_smt.api.proofs.ProofRule;
+import org.sosy_lab.java_smt.basicimpl.AbstractProofDAG;
+
+public class CVC5ProofDAG extends AbstractProofDAG {
```
baierd, line 26: The better name would be `CVC5Proof` ;D

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ProofDAG.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public static class CVC5ProofNode extends AbstractProofNode {
+
+    public CVC5ProofNode(ProofRule pProofRule, Formula formula) {
+
+      super(pProofRule, formula);
+    }
+
+    static CVC5ProofNode fromCVC5Proof(Proof pProof, CVC5FormulaCreator formulaCreator)
```
baierd, line 40: Should be part of the encasing class.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ProofDAG.java
```diff
@@ -0,0 +1,122 @@
...
+
+      while (!stack.isEmpty()) {
+        CVC5Frame frame = stack.peek();
+
+        if (!frame.isVisited()) {
+
+          frame.setNumArgs(frame.getProof().getChildren().length);
+          frame.setAsVisited(true);
+
+          for (int i = frame.getNumArgs() - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
baierd: But why not build the DAG in the same order as the original?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5ProofDAG.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+      }
+      return computed.get(pProof);
+    }
+
+    private static Proof changeRoot(Proof root) {
+      return root.getChildren()[0];
+    }
+
+    String asString(int indentLevel) {
```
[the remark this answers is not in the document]
baierd: Then i suggest to name it something like `proofAsString()`, making it clear to everyone that this prints the entire (sub)-proof.

baierd (changes requested):
Changes go into the right direction! This is now much easier to understand and read!

My change-requests from CVC5 apply to your proof structure in general.
In detail:

The DAG should only consist of nodes and edges. This can be modeled using the map you already use, but you don't need the ID's. 
Simply node -> node.

The interface and general object returned when requesting a proof should be named `Proof`, with implementations for each solver.
This `Proof` should hold your DAG, which consists of nodes. The proof class should also build the proof (DAG) from the native solver return.
Similarly to the `Model`.

src/org/sosy_lab/java_smt/ResProofRule.java
```diff
@@ -10,22 +10,38 @@
...
-   */
+  private static final Map<String, ResAxiom> RULE_MAP = new HashMap<>();
+
+  private ResProofRule() {
+    // prevent instantiation
+  }
+
+  static {
+    for (ResAxiom rule : ResAxiom.values()) {
+      RULE_MAP.put(rule.getName().toLowerCase(Locale.ROOT), rule);
```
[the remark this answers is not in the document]
gcarpio21: Something was complaining about the toLowerCase() without arguments in the CI. I will change the parameter to ENGLISH

src/org/sosy_lab/java_smt/SolverContextFactory.java
```diff
@@ -282,7 +282,8 @@ private SolverContext generateContext0(Solvers solverToCreate)
       case Z3:
         return Z3SolverContext.create(
             logger,
-            config,
+            Configuration.builder().copyFrom(config)
```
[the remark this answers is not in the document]
gcarpio21: I created the option a while ago. I just looked again to make sure everything works properly and made some small changes, it should be enabled by default now.

src/org/sosy_lab/java_smt/api/proofs/ProofFrame.java
```diff
@@ -0,0 +1,57 @@
...
+package org.sosy_lab.java_smt.api.proofs;
+
+/**
+ * Stores proof related information and gets stored in a stack when processing proofs. Each frame
+ * contains a proof object and the number of arguments it has.
+ *
+ * @param <T> The type of the proof object.
+ */
+public abstract class ProofFrame<T> {
+  T proof;
```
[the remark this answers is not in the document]
gcarpio21: Proof is the only variable that only gets set once

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
gcarpio21: I am having trouble because this method is called in `CVC5AbstractProver` which is where the`FormulaCreator` gets passed to it so the transformation can be done. Going around this causes some issues, e.g., if I make `generateProof()` an abstract method in `AbstractProofNode` or `AbstractProofDAG` I and use it to pass the  I get a warning that it is getting exposed outside its declaration scope, I assume that is because  needs to be protected . I guess the method could go in `CVC5AbstractProver`.

PR #459, opened 2025-03-19
Add support for integer modulo in MathSAT

daniel-raffler:
Hello,
MathSAT is missing native support for integer modulo. In this PR we add a work-around based on the definition `remainder = dividend - floor(dividend/divisor)*divisor` to support the operation.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5IntegerFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5IntegerFormulaManager.java
@@ -70,6 +70,11 @@ class Mathsat5IntegerFormulaManager
   }
 
   @Override
+  protected Long modulo(Long pNumber1, Long pNumber2) {
+    return subtract(pNumber1, multiply(divide(pNumber1, pNumber2), pNumber2));
+  }
+
+  @Override
   protected Long modularCongruence(Long pNumber1, Long pNumber2, BigInteger pModulo) {
     return modularCongruence0(pNumber1, pNumber2, pModulo.toString());
   }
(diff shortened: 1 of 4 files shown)


src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5IntegerFormulaManager.java
```diff
@@ -69,6 +77,21 @@ public Long divide(Long pNumber1, Long pNumber2) {
...
+    // any value for the remainder. We solve this by first checking the divisor and returning an
+    // UF symbol "modZeroUF(dividend)" if it is zero. Otherwise, the formula is used to calculate
+    // the remainder.
+    return msat_make_term_ite(
+        mathsatEnv,
+        msat_make_equal(mathsatEnv, pNumber2, msat_make_int_number(mathsatEnv, 0)),
+        getFormulaCreator().callFunctionImpl(modZeroUF, ImmutableList.of(pNumber1)),
+        subtract(pNumber1, multiply(divide(pNumber1, pNumber2), pNumber2)));
+  }
+
```
kfriedberger, line 94: Be aware that UFs return the same value for the same parameters.
If SMTLIB allows "any return value", then the return value could be random and return a different result at second call.
From technical point, returning the same result on second call would be fully ok for me. From SMTLIB point, please check what the standard and other solvers return for `x/0 == x/0`. Is this always satisfied?

kfriedberger: Valid point. Lets approve it.

kfriedberger (approved):
lgtm.
Improved functionality and tests for special cases.

PR #460, opened 2025-03-19
Extend Mathsat5 Native Wrapper with API for Proofs

baierd:
This PR extends the Mathsat5 wrapper with API for proofs and adds a small native test for them.

Unfortunately this came up after the merge of the most recent changes. Sorry about that.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5NativeApi.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5NativeApi.java
@@ -1095,7 +1095,6 @@ final class Mathsat5NativeApi {
 
   public static native String msat_proof_get_name(long proof);
 
-  // Non-zero if term, 0 for no term
   public static native boolean msat_proof_is_term(long proof);
 
   // Term representation of the current proof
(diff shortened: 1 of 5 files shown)


src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5NativeApi.java
```diff
@@ -1064,4 +1064,46 @@ private static native int msat_all_sat(
...
+  public static native boolean msat_proof_is_term(long proof);
+
+  // Term representation of the current proof
+  public static native long msat_proof_get_term(long proof);
+
+  // Can be used to check for the equality of proofs
+  public static native int msat_proof_id(long proof);
+
+  // Cleans up the proof manager and the associated proof
+  public static native void msat_destroy_proof_manager(long proofMgr);
```
kfriedberger, line 1108: Request for information, not for action: Are all those methods also available in OptiMathSAT-API or were they added to MathSAT after year 2022?

kfriedberger (approved):
lgtm.

If I should publish a new version of native libraries to Ivy and Maven repository, you might need to wait a few days.

baierd:
I found a small problem and fixed it.

A new version would be nice once you have the time. @gcarpio21 is currently working on adding a common proof format and MathSAT5 is a good addition.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5NativeApi.java
baierd: The API is from 2012, so it should work.

kfriedberger (approved):
This PR is an internal extension of the supported MathSAT5 API.
With future changes, we might use those new functions to access proofs of the solver.

PR #464, opened 2025-03-21
Fix empty sums in Z3

daniel-raffler:
Hello,
Z3 crashes when `NumeralFormulaManager.sum()` is called with no arguments. This PR adds some checks to handle this special-case.


kfriedberger (approved):
lgtm.

PR #465, opened 2025-03-21
Add special-case handling for distinct() with fewer than 2 arguments

daniel-raffler:
Hello,
several solvers crash when `NumeralFormulaManager.distinct()` is used with fewer than 2 arguments. This PR adds support for such special-cases to all solvers.


kfriedberger (approved):
lgtm.

PR #466, opened 2025-03-22
Add explicit cast in mixed integer-real term

daniel-raffler:
Hello,
several solvers fail when integer terms and rational terms are used together in `RationalFormulaManager`. This PR adds explicit casting operations for solvers that don't convert implicitly.


kfriedberger:
It is not clear why this change is required. Please add useful tests that show your contribution, i.e., tests need to fail upfront and violate the official interfaces or documentation, and work with your changes.

daniel-raffler:
> It is not clear why this change is required. Please add useful tests that show your contribution, i.e., tests need to fail upfront and show a violation the official interfaces or documentation, and pass with your changes.

Did you already try running `MixedArithmeticsTest` with version 53b4026820c6789aef47a11d8f2de41434cd781c? Some of the assertions might be debatable, but I'm also getting 24 crashes due to type errors, which shouldn't be possible. The JavaSMT API guarantees that something like `RationalFormulaManager.add(a,b)` will always return a `RationalFormula`. Yet several of the solvers return Integer terms and crash either while trying to wrap the result, or later when the type of the formula does not match in the next operation.

If you want to avoid explicit casts we could still report this to the solver developers and ask them for help?

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
```diff
@@ -107,30 +108,82 @@ protected IExpression makeVariableImpl(String varName) {
...
+  protected IFormula equal(IExpression number1, IExpression number2) {
+    return super.equal(toReal((ITerm) number1), toReal((ITerm) number2));
+  }
+
+  @Override
+  protected IExpression distinctImpl(List<IExpression> operands) {
+    List<IExpression> castOperands =
+        FluentIterable.from(operands)
+                .anyMatch(t -> Sort.sortOf((ITerm) t).equals(PrincessEnvironment.FRACTION_SORT))
+            ? FluentIterable.from(operands).transform(t -> (IExpression) toReal((ITerm) t)).toList()
```
kfriedberger, line 165: Why is this casting limited to fraction-sort? Why not cast all input?

daniel-raffler: I think the check is part of the condition? We only need to cast to `Real`if at least one of the arguments is a `Real`. Otherwise we might as well compare them as integers

kfriedberger:
I'm surprised that the issues in this PR have not been reported earlier, particularly the potentially breaking casts in multiple solver bindings and the invalid formula comparisons (LESS_THAN, LESS_EQUAL, GREATER_THAN, GREATER_EQUAL) in Princess for Real/Rational theory.

The PR looks completely valid from my side.

kfriedberger:
@daniel-raffler Thanks for reporting this issue and provoding a solid basis fand test or the fix. Well done.

PR #470, opened 2025-04-01
Use official binaries for CVC5/Add parsing support

daniel-raffler:
Hello,
this PR switches the build to using the official CVC5 binaries release on [github](https://github.com/cvc5/cvc5/releases/tag/latest). This will allow us to support macOS on both x64 and arm64 platforms. I've also added support for SMTLIB parsing in CVC5, but this could still be split off into a separate PR if needed. Before merging we should wait for the next CVC5 release as the daily builds that this PR currently uses will not be permanently available. The last stable release, however, does not have some bug fixes that are need.


baierd:
Depending on the size of changes we can include the parsing here or make a new PR for it. Generally, if its just small additions, its fine. If its more difficult, we should split it.

daniel-raffler:
> Depending on the size of changes we can include the parsing here or make a new PR for it. Generally, if its just small additions, its fine. If its more difficult, we should split it.

Ok, I will split up then. The changes for the parser are somewhat larger and I don't know when the next version of CVC5 will be released. It's probably better to do this separately.

PR #473, opened 2025-04-05
Disable the flaky tests for CVC5

daniel-raffler:
Hello,
the flaky tests seem to have gotten worse after the recent update to CVC5. I would suggest disabling `TimoutTest` and `SolverConcurrencyTest` entirely for CVC5 to temporarily fix the issue. The concurrency issues with CVC5 are well known and have already been reported [here](https://github.com/cvc5/cvc5/issues/10256) and [here](https://github.com/cvc5/cvc5/discussions/10265). They seem to be working on a solution, but it will take more time. In the meantime we should have a test-suite that is always green, and once the concurrency problems in CVC5 are resolved the test can easily be turned on again.


kfriedberger (approved):
lgtm.

PR #475, opened 2025-04-05
CVC5: Use official binaries

daniel-raffler:
Hello,
this PR switches the build to using the official CVC5 binaries release on [github](https://github.com/cvc5/cvc5/releases/tag/latest). This will allow us to support macOS on both x64 and arm64 platforms. Before merging we should wait for the next CVC5 release as the daily builds that this PR currently uses will not be permanently available. The last stable release, however, does not have some bug fixes that are need.

### DIFF
--- a/build/build-maven-publish.xml
+++ b/build/build-maven-publish.xml
@@ -238,6 +238,8 @@ SPDX-License-Identifier: Apache-2.0
         <stage-solver-file filename="libcvc5jni" classifier="libcvc5jni-x64" filedirectory="${libDir.x64}" fileending="so"/>
         <stage-solver-file filename="libcvc5jni" classifier="libcvc5jni-arm64" filedirectory="${libDir.arm64}" fileending="so"/>
         <stage-solver-file filename="libcvc5jni" classifier="libcvc5jni-x64" filedirectory="${libDir.x64}" fileending="dll"/>
+        <stage-solver-file filename="libcvc5jni" classifier="libcvc5jni-x64" filedirectory="${libDir.x64}" fileending="dylib"/>
+        <stage-solver-file filename="libcvc5jni" classifier="libcvc5jni-arm64" filedirectory="${libDir.arm64}" fileending="dylib"/>
     </target>
 
     <!--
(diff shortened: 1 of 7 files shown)


kfriedberger:
Using daily builds is completely valid, we just should not update them on daily basis. 😄
Each version is archived in Ivy@SoSy-Lab and its version is a git reference. #️⃣ 

The only disadvantage is that JavaSMT developers give part of their authority over and knowledge about library compilation towards the original developers of CVC5, which is completely fine. We already to the same with other pre-packaged solvers, such as Z3, SMTInterpol, and Princess.

The new build scripts are quite simple to maintain, do not depend on any build setup (like Docker), and we get macOS support for free ✅ .

I would suggest to merge this PR early, maybe even now if there is no serios blocker. Some further improvements for CVC5 can be applied later in other PRs.

kfriedberger (approved):

kfriedberger:
@daniel-raffler 
Using an identical branch name twice is considered "bad style". 😃 
It forces other developers that potenitally have local commits on the old branch to rebase and reorganize. Additionally, as SoSy-Lab, each GitHub repository is synched with GitLab (one-way only), and branch updates are only recognized, as long as the name is unique and no force-push happened, i.e., if the history of all branches is clean. In such cases we need to manually wipe the outdated and synched branch from GitLab to get GitLab-CI working again for a branch.

daniel-raffler:
> @daniel-raffler Using an identical branch name twice is considered "bad style". 😃 It forces other developers that potentially have local commits on the old branch to rebase and reorganize. Additionally, at SoSy-Lab, each GitHub repository is synched with GitLab (one-way only), and branch updates are only recognized, as long as the name is unique and no force-push happened, i.e., if the history of all branches is clean. In such cases we need to manually wipe the outdated and synched branch from GitLab to get GitLab-CI working again for a branch.

Sorry, about that. I'll be more careful next time and will create a new branch.

> I would suggest to merge this PR early, maybe even now if there is no serious blocker.

I think the only thing that is left to do is change the download location if we ever want to switch from dailies to regular releases. Then again we might just want to stick with the dailies to get more frequent updates.

PR #477, opened 2025-04-11
Improve optimization API and add fallback implementation

### LINKED ISSUE #476: Optimization API and Resource Management Improvements Needed
# Inconsistent Optimization Support Across Solvers

## Description
The optimization API in JavaSMT has inconsistent implementation across different solvers, leading to unpredictable behavior when applications switch between solver backends. Currently, only Z3 and MathSAT5 (via OptiMathSAT) properly support optimization capabilities, while other solvers simply throw exceptions without providing alternative functionality.

### Evidence in Implementation Code:

Z3 provides full implementation:
```java
// In Z3SolverContext.java
public OptimizationProverEnvironment newOptimizationProverEnvironment0(
    Set<ProverOptions> options) {
  Z3SolverContext.ExtraOptions extraOptions = new Z3SolverContext.ExtraOptions(config);
  return new Z3OptimizationProver(
      creator,
      logger,
      (Z3FormulaManager) getFormulaManager(),
      options,
      ImmutableMap.<String, Object>builder()
          .put(OPT_ENGINE_CONFIG_KEY, extraOptions.optimizationEngine)
          .put(OPT_PRIORITY_CONFIG_KEY, extraOptions.objectivePrioritizationMode)
          .build(),
      logfile,
      shutdownNotifier);
}
```

While other solvers throw exceptions:
```java
// In YicesSolverContext.java
protected OptimizationProverEnvironment newOptimizationProverEnvironment0(
    Set<ProverOptions> options) {
  throw new UnsupportedOperationException("Yices does not support optimization");
}

// In SMTInterpolSolverContext.java
public OptimizationProverEnvironment newOptimizationProverEnvironment0(
    Set<ProverOptions> options) {
  throw new UnsupportedOperationException("SMTInterpol does not support optimization");
}

// In PrincessSolverContext.java
public OptimizationProverEnvironment newOptimizationProverEnvironment0(
    Set<ProverOptions> pOptions) {
  throw new UnsupportedOperationException("Princess does not support optimization");
}
```

### Test Code Shows Special Handling:

Tests must skip optimization tests for most solvers:
```java
// In OptimizationTest.java
@Before
public void skipUnsupportedSolvers() {
  requireOptimization();
}

// In SolverBasedTest0.java
protected final void requireOptimization() {
  try {
    context.newOptimizationProverEnvironment().close();
  } catch (UnsupportedOperationException e) {
    assume()
        .withMessage("Solver %s does not support optimization", solverToUse())
        .that(e)
        .isNull();
  }
}
```

Even between supported solvers, behavior differs:
```java
// From OptimizationTest.java - MathSAT5 specific workaround
if (solverToUse() == Solvers.MATHSAT5) {
  // see https://github.com/sosy-lab/java-smt/issues/233
  assume()
      .withMessage("OptiMathSAT 1.7.2 has a bug with switching objectives")
      .that(context.getVersion())
      .doesNotContain("MathSAT5 version 1.7.2");
  assume()
      .withMessage("OptiMathSAT 1.7.3 has a bug with switching objectives")
      .that(context.getVersion())
      .doesNotContain("MathSAT5 version 1.7.3");
}

// Different precision handling between solvers
// OptiMathSAT5 has at least an epsilon of 1/1000000. It does not allow larger values.
assume()
    .withMessage("Solver %s does not support large epsilons", solverToUse())
    .that(solver)
    .isNotEqualTo(Solvers.MATHSAT5);
```

## Impact
- Applications using optimization features cannot easily switch between solvers
- Code must handle UnsupportedOperationExceptions and provide fallbacks
- Optimization functionality is effectively limited to Z3 in many cases
- Testing is incomplete due to varying solver capabilities

shivammm21:
# Optimization API and Resource Management Improvements

## Description
This PR addresses the inconsistent optimization support across SMT solvers by implementing a robust fallback mechanism and standardizing the optimization API. The changes provide a consistent interface for optimization features across all supported solvers.

## Changes
- Added `AbstractOptimizationProver` base class for standardization
- Implemented `FallbackOptimizationProver` for solvers without native support
- Added `SolverVersionChecker` for version compatibility checks
- Created comprehensive test suite
- Added detailed documentation

## Key Features
- Binary search approach for finding maximum/minimum values
- Support for integer optimization
- Proper error handling and logging
- Efficient resource management
- Version compatibility checks for MathSAT5

Fixes #476


src/org/sosy_lab/java_smt/SolverContextFactory.java
```diff
@@ -366,4 +366,42 @@ public static SolverContext createSolverContext(Solvers solver)
...
   }
+
+  /**
+   * Create a new optimization prover environment.
+   * @param context The solver context
+   * @param options The prover options
+   * @return An optimization prover environment
+   * @throws SolverOptimizationException if the solver doesn't support optimization
+   */
+  public static OptimizationProverEnvironment createOptimizationProver(
```
kfriedberger, line 377: The `SolverContextFactory` is intended to construct a `SolverContext`. This method violates this constraint, as it constructs a `ProverEnvironment`.

src/org/sosy_lab/java_smt/basicimpl/SolverVersionChecker.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Get a descriptive message about optimization support for a solver.
+     * @param solver The solver to check
+     * @return A message describing optimization support
+     */
+    public static String getOptimizationSupportMessage(Solvers solver) {
+        if (solver == Solvers.Z3) {
+            return "Z3 provides full optimization support";
+        } else if (solver == Solvers.MATHSAT5) {
+            return "MathSAT5 provides partial optimization support (versions > 1.7.3)";
```
kfriedberger, line 44: This message is incorrect. MathSAT5 (original version) does not support optimization, only OptiMathSAT does.

src/org/sosy_lab/java_smt/basicimpl/SolverVersionChecker.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+    
+    /**
+     * Check if optimization is supported for the given solver.
+     * @param solver The solver to check
+     * @return true if optimization is supported
+     */
+    public static boolean isOptimizationSupported(Solvers solver) {
+        return solver == Solvers.Z3 || solver == Solvers.MATHSAT5;
```
kfriedberger, line 32: This check is redundant. The same "feature" was already implemented in the solver-specific code when constructing a `OptimizationProverEnvironment`. The prover throws an `UnsupportedOperationException` if unsupported, or constructs a `OptimizationProverEnvironment` if supported.

src/org/sosy_lab/java_smt/basicimpl/SolverVersionChecker.java
```diff
@@ -0,0 +1,49 @@
...
+public final class SolverVersionChecker {
+    private SolverVersionChecker() {}
+    
+    /**
+     * Check if the given MathSAT5 version has known optimization bugs.
+     * @param version The solver version string
+     * @throws SolverOptimizationException if the version has known issues
+     */
+    public static void checkMathSATVersion(String version) {
+        if (version.startsWith("1.7.2") || version.startsWith("1.7.3")) {
```
kfriedberger, line 18: We avoid hard-coded versions, because this blocks users from simply replacing solver binaries and test on their own. We try to provide as much from the solver as possible and useful for the user, and do not create artificial boundaries.

src/org/sosy_lab/java_smt/basicimpl/FallbackOptimizationProver.java
```diff
@@ -0,0 +1,203 @@
...
+                    } else {
+                        // Solution exists, try a higher value
+                        lowerBound = mid;
+                    }
+                } finally {
+                    baseProver.pop();
+                }
+            }
+            
+            return Optional.of(upperBound);
```
kfriedberger, line 115: Binary search is a valid idea and it should work in most applications. As a first prototype, this might work for several use cases. The performance overhead is not that large, e.g., several SAT checks are required (which is ok) and the number of them is logarithmic in the bounding window size (which is good).

There are some open points:
- We need to think about upper/lower bounds on unbounded theories like Integer or Rational.
- Maybe keep the prover content if the next contraint it stricter than the previous. This might improve reuse and perfomance on larger queries.

kfriedberger (changes requested):
Please split your implementation on "API consistency" from "default fallback algorithm".

For API consistency, I do not yet seen any benefit. It looks to become even worse.

For the fallback algorithm, the initial approach looks quite simple and could work for most use-cases on Integer theory.

There are some further comments above.

shivammm21:
Thank you for the review. I will work on the suggested changes and aim to resolve everything within a week.

PR #478, opened 2025-04-24
Update Maintainers and Add Contributors

baierd:
Updates maintainers and adds contributors.
We could add many more contributors, we should ask them if they want to!


PhilippWendler (commented):
For the record: I was not aware that I am listed as maintainer of JavaSMT and of course I am not a maintainer of JavaSMT, this should indeed be fixed.

There are also several Maven files that need to be updated.

kfriedberger (approved):

kfriedberger:
We cannot change existing Maven releases, but the pom-template was cleaned for future releases.
See 8ddbe962a83ab5a85a01cbc435acc1e8290a4157 for details.

PR #479, opened 2025-04-25
Update OpenSMT to 2.9.0 and include support for proofs in the JNI bindings

daniel-raffler:
Hello everyone,
this PR will update OpenSMT to version 2.9.0 while also adding the method `MainSolver.printResolutionProofSMT2` to the JNI bindings in preparation for https://github.com/sosy-lab/java-smt/pull/458. There are also some minor simplifications to the build process and the binaries for GMP on x86-linux and arm64-linux are now included in the build container and no longer have to be provided by the maintainer when updating the binaries.


kfriedberger (dismissed):
Overall, good work.

The commits are clearly structured and the code changes are good.

Personally, I would prefer to not include dependencies like GMP or JDK in the Docker image as they cannot be easily updated or exchanged, but require a rebuild of the image. In this PR, there still is the comand line option to set them, which is nice and a valid solution.

I will test the overall build chain the next days.

daniel-raffler:
Thanks for the review! 👍

Issue #100, opened 2017-01-12
Can't use shutdownmanager more than once


Zomono:
Hi, I'm very happy with this project and it saved me a lot of work. Thank you so much!

In my current project I need to do a lot of satisfiability checks. Because for some formulas the Z3 Prover runs infinitely long, I am using the ShutdownManager to stop the Z3 after a given timeout. But this works only for the first time. I can't use the Shutdownmanager a second time for the same context.

What I have done:
- I create one context like it is done in your example
- For every sat-check I create a new ProverEnvironment and use the ShutdownManager from a different thread if the Prover runs to long.

The problem is, that the ShutdownNotifier ignores additional calls to _requestShutdown_.
I also tried to close the context and create a new on for each sat-check, but this results immediately to a native InterruptException and has a bad effect on the performance.

So what should I do to stop every sat-check (if needed)?

kfriedberger:
Hi Zerlono,
the `ShutdownManager` has an internal state and can only be used once, afterwards it will always terminate when used. Maybe you can use your own implementation of `ShutdownManager` wrapping an instance of a real `ShutdownManager`, which you can `reset` (or re-instantiate) after calling the shutdown method (and waiting for the solvers termination). Then the internal state would be reset to `running` and should only terminate when `requestShutdown` is called.

PhilippWendler:
`ShutdownManager` is not resettable on purpose because it could not be guaranteed that libraries like the SMT solver leave everything in a clean state when being interrupted. This problem would not be solved by implementing an own resettable replacement.

I would expect, however, that separate contexts are not affected by previous contexts' shutdown managers, and if this happens it should be treated as a bug and be solved. Can you provide some code that reproduces this problem?

Zerlono:
The `ShutdownManger` has a private constructor anyway and can't be extended. 
Is a new context really needed for every call? This may has a performance impact due to the amount of  sat-check calls done within my program.

Because I don't do any thing special compared to the given examples, one of the example is maybe enough to reproduce it. 
Here is the code used within my project:

```
package uniks.sct.solver.subsolver.quantifier;

import lombok.extern.slf4j.Slf4j;
import org.sosy_lab.common.ShutdownManager;
import org.sosy_lab.common.configuration.Configuration;
import org.sosy_lab.common.configuration.InvalidConfigurationException;
import org.sosy_lab.common.log.LogManager;
import org.sosy_lab.java_smt.SolverContextFactory;
[...]

PR #489, opened 2025-05-30
Add Prover Based ShutdownManager

baierd:
Most solvers allow either re-usable shutdown or we use shared environments that are created per prover, making them isolated enough to be shut down without stopping the context.

This PR adds a controlled way of shutting down only a single prover with the manager returned by our API IFF it is supported by the solver. These prover shutdown managers are all children of the context given shutdown manager.

### DIFF
--- a/src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
@@ -15,6 +15,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
+import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 
 /**
@@ -202,4 +204,18 @@ public interface BasicProverEnvironment<T> extends AutoCloseable {
   default boolean registerUserPropagator(UserPropagator propagator) {
     return false;
   }
+
+  /**
+   * Returns the {@link ShutdownManager} registered for this {@link ProverEnvironment}. It is
+   * guaranteed to be a child of the {@link ShutdownNotifier} given to the creating {@link
+   * SolverContext}, resulting in shutdown when the {@link SolverContext}s notifier is shutting
+   * down. The notifier returned here can be shut down independently of the creating contexts
+   * notifier.
+   *
+   * @return a {@link ShutdownManager} who is the child of the {@link ShutdownNotifier} used in the
+   *     creating {@link SolverContext}, that can be used to shut down only this {@link
+   *     ProverEnvironment}.
+   * @throws UnsupportedOperationException if the solver does not support prover specific shutdown.
+   */
+  ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException;
 }
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
@@ -22,6 +22,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
+import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Evaluator;
@@ -47,7 +49,11 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
 
   private static final String TEMPLATE = "Please set the prover option %s.";
 
-  protected AbstractProver(Set<ProverOptions> pOptions) {
+  // Do we even need this?
+  private final ShutdownNotifier contextShutdownNotifier;
+  protected final ShutdownManager proverShutdownManager;
+
+  protected AbstractProver(ShutdownNotifier pShutdownNotifier, Set<ProverOptions> pOptions) {
     generateModels = pOptions.contains(ProverOptions.GENERATE_MODELS);
     generateAllSat = pOptions.contains(ProverOptions.GENERATE_ALL_SAT);
     generateUnsatCores = pOptions.contains(ProverOptions.GENERATE_UNSAT_CORE);
@@ -56,6 +62,9 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
     enableSL = pOptions.contains(ProverOptions.ENABLE_SEPARATION_LOGIC);
 
     assertedFormulas.add(LinkedHashMultimap.create());
+
+    contextShutdownNotifier = pShutdownNotifier;
+    proverShutdownManager = ShutdownManager.createWithParent(contextShutdownNotifier);
   }
 
   protected final void checkGenerateModels() {
@@ -157,5 +166,18 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
     assertedFormulas.clear();
     closeAllEvaluators();
     closed = true;
+  }
+
+  @Override
+  public ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException {
+    return getShutdownManagerForProverImpl();
+  }
+
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    // Override this with the prover specific ShutdownManagers notifier for supporting solvers.
+    // The solver should then use the prover specific ShutdownManagers notifier for stopping
+    // instead of the contexts' notifier!
+    throw new UnsupportedOperationException(
+        "The chosen solver does not support isolated prover " + "shutdown");
   }
 }
--- a/src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Model;
@@ -127,5 +128,10 @@ public class BasicProverWithAssumptionsWrapper<T, P extends BasicProverEnvironme
       throws InterruptedException, SolverException {
     clearAssumptions();
     return delegate.allSat(pCallback, pImportant);
+  }
+
+  @Override
+  public ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException {
+    return delegate.getShutdownManagerForProver();
   }
 }
--- a/src/org/sosy_lab/java_smt/delegate/debugging/DebuggingBasicProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/delegate/debugging/DebuggingBasicProverEnvironment.java
@@ -14,6 +14,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Model;
@@ -27,6 +28,11 @@ class DebuggingBasicProverEnvironment<T> implements BasicProverEnvironment<T> {
       BasicProverEnvironment<T> pDelegate, DebuggingAssertions pDebugging) {
     delegate = checkNotNull(pDelegate);
     debugging = pDebugging;
+  }
+
+  @Override
+  public ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException {
+    return delegate.getShutdownManagerForProver();
   }
 
   @Override
--- a/src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedBasicProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedBasicProverEnvironment.java
@@ -15,6 +15,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Model;
@@ -29,6 +30,11 @@ class SynchronizedBasicProverEnvironment<T> implements BasicProverEnvironment<T>
   SynchronizedBasicProverEnvironment(BasicProverEnvironment<T> pDelegate, SolverContext pSync) {
     delegate = checkNotNull(pDelegate);
     sync = checkNotNull(pSync);
+  }
+
+  @Override
+  public ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException {
+    return delegate.getShutdownManagerForProver();
   }
 
   @Override
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Model;
@@ -38,7 +39,10 @@ class BitwuzlaTheoremProver extends AbstractProverWithAllSat<Void> implements Pr
       new Terminator() {
         @Override
         public boolean terminate() {
-          return shutdownNotifier.shouldShutdown(); // shutdownNotifer is defined in the superclass
+          return proverShutdownManager
+              .getNotifier()
+              .shouldShutdown(); // shutdownNotifer is defined in the
+          // superclass
         }
       };
   private final Bitwuzla env;
@@ -119,7 +123,8 @@ class BitwuzlaTheoremProver extends AbstractProverWithAllSat<Void> implements Pr
       return false;
     } else if (resultValue == Result.UNSAT) {
       return true;
-    } else if (resultValue == Result.UNKNOWN && shutdownNotifier.shouldShutdown()) {
+    } else if (resultValue == Result.UNKNOWN
+        && proverShutdownManager.getNotifier().shouldShutdown()) {
       throw new InterruptedException();
     } else {
       throw new SolverException("Bitwuzla returned UNKNOWN.");
@@ -244,6 +249,11 @@ class BitwuzlaTheoremProver extends AbstractProverWithAllSat<Void> implements Pr
             this,
             creator,
             Collections2.transform(getAssertedFormulas(), creator::extractInfo)));
+  }
+
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
   }
 
   public boolean isClosed() {
--- a/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
@@ -49,7 +49,7 @@ abstract class BoolectorAbstractProver<T> extends AbstractProverWithAllSat<T> {
     this.manager = manager;
     this.creator = creator;
     this.btor = btor;
-    terminationCallback = shutdownNotifier::shouldShutdown;
+    terminationCallback = proverShutdownManager.getNotifier()::shouldShutdown;
     terminationCallbackHelper = addTerminationCallback();
 
     isAnyStackAlive = pIsAnyStackAlive;
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
@@ -202,11 +203,12 @@ class CVC4TheoremProver extends AbstractProverWithAllSat<Void>
     }
 
     Result result;
-    try (ShutdownHook hook = new ShutdownHook(shutdownNotifier, smtEngine::interrupt)) {
-      shutdownNotifier.shutdownIfNecessary();
+    try (ShutdownHook hook =
+        new ShutdownHook(proverShutdownManager.getNotifier(), smtEngine::interrupt)) {
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
       result = smtEngine.checkSat();
     }
-    shutdownNotifier.shutdownIfNecessary();
+    proverShutdownManager.getNotifier().shutdownIfNecessary();
     return convertSatResult(result);
   }
 
@@ -259,5 +261,10 @@ class CVC4TheoremProver extends AbstractProverWithAllSat<Void>
       exprManager.delete();
     }
     super.close();
+  }
+
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
   }
 }
--- a/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
@@ -203,7 +203,7 @@ abstract class CVC5AbstractProver<T> extends AbstractProverWithAllSat<T> {
 
     /* Shutdown currently not possible in CVC5. */
     Result result = solver.checkSat();
-    shutdownNotifier.shutdownIfNecessary();
+    proverShutdownManager.getNotifier().shutdownIfNecessary();
     return convertSatResult(result);
   }
 
@@ -251,4 +251,10 @@ abstract class CVC5AbstractProver<T> extends AbstractProverWithAllSat<T> {
     }
     super.close();
   }
+
+  /* TODO: revisit once CVC5 supports interruption
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
+  }*/
 }
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Evaluator;
@@ -51,27 +52,24 @@ import org.sosy_lab.java_smt.api.SolverException;
 import org.sosy_lab.java_smt.basicimpl.AbstractProver;
 import org.sosy_lab.java_smt.basicimpl.CachingModel;
 import org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.AllSatModelCallback;
+import org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5NativeApi.TerminationCallback;
 
 /** Common base class for {@link Mathsat5TheoremProver} and {@link Mathsat5InterpolatingProver}. */
 abstract class Mathsat5AbstractProver<T2> extends AbstractProver<T2> {
 
-  protected final Mathsat5SolverContext context;
   protected final long curEnv;
   private final long curConfig;
   protected final Mathsat5FormulaCreator creator;
-  private final ShutdownNotifier shutdownNotifier;
 
   protected Mathsat5AbstractProver(
       Mathsat5SolverContext pContext,
       Set<ProverOptions> pOptions,
       Mathsat5FormulaCreator pCreator,
       ShutdownNotifier pShutdownNotifier) {
-    super(pOptions);
-    context = pContext;
+    super(pShutdownNotifier, pOptions);
     creator = pCreator;
     curConfig = buildConfig(pOptions);
-    curEnv = context.createEnvironment(curConfig);
-    shutdownNotifier = pShutdownNotifier;
+    curEnv = pContext.createEnvironment(curConfig);
   }
 
   private long buildConfig(Set<ProverOptions> opts) {
@@ -100,12 +98,24 @@ abstract class Mathsat5AbstractProver<T2> extends AbstractProver<T2> {
   public boolean isUnsat() throws InterruptedException, SolverException {
     Preconditions.checkState(!closed);
 
-    final long hook = msat_set_termination_callback(curEnv, context.getTerminationTest());
+    final long hook = msat_set_termination_callback(curEnv, getTerminationTest());
     try {
       return !msat_check_sat(curEnv);
     } finally {
       msat_free_termination_callback(hook);
     }
+  }
+
+  /**
+   * Get a termination callback for the current prover (who's parent is the contexts callback). The
+   * callback can be registered upfront, i.e., before calling a possibly expensive computation in
+   * the solver to allow a proper shutdown.
+   */
+  private TerminationCallback getTerminationTest() {
+    return () -> {
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
+      return false;
+    };
   }
 
   @Override
@@ -114,7 +124,7 @@ abstract class Mathsat5AbstractProver<T2> extends AbstractProver<T2> {
     Preconditions.checkState(!closed);
     checkForLiterals(pAssumptions);
 
-    final long hook = msat_set_termination_callback(curEnv, context.getTerminationTest());
+    final long hook = msat_set_termination_callback(curEnv, getTerminationTest());
     try {
       return !msat_check_sat_with_assumptions(curEnv, getMsatTerm(pAssumptions));
     } finally {
@@ -273,10 +283,15 @@ abstract class Mathsat5AbstractProver<T2> extends AbstractProver<T2> {
 
     @Override
     public void callback(long[] model) throws InterruptedException {
-      shutdownNotifier.shutdownIfNecessary();
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
       clientCallback.apply(
           Collections.unmodifiableList(
               Lists.transform(Longs.asList(model), creator::encapsulateBoolean)));
     }
+  }
+
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
   }
 }
--- a/src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.ShutdownNotifier;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Evaluator;
@@ -230,8 +231,9 @@ public abstract class OpenSmtAbstractProver<T> extends AbstractProverWithAllSat<
     changedSinceLastSatQuery = false;
 
     sstat result;
-    try (ShutdownHook listener = new ShutdownHook(shutdownNotifier, osmtSolver::stop)) {
-      shutdownNotifier.shutdownIfNecessary();
+    try (ShutdownHook listener =
+        new ShutdownHook(proverShutdownManager.getNotifier(), osmtSolver::stop)) {
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
       try {
         result = osmtSolver.check();
       } catch (Exception e) {
@@ -250,7 +252,7 @@ public abstract class OpenSmtAbstractProver<T> extends AbstractProverWithAllSat<
           throw new SolverException("OpenSMT crashed while checking satisfiability.", e);
         }
       }
-      shutdownNotifier.shutdownIfNecessary();
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
     }
 
     if (result.equals(sstat.Error())) {
@@ -288,5 +290,10 @@ public abstract class OpenSmtAbstractProver<T> extends AbstractProverWithAllSat<
       osmtSolver.delete();
     }
     super.close();
+  }
+
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
   }
 }
--- a/src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolAbstractProver.java
@@ -52,7 +52,6 @@ abstract class SmtInterpolAbstractProver<T> extends AbstractProver<T> {
   protected final FormulaCreator<Term, Sort, Script, FunctionSymbol> creator;
   protected final SmtInterpolFormulaManager mgr;
   protected final Deque<PersistentMap<String, BooleanFormula>> annotatedTerms = new ArrayDeque<>();
-  protected final ShutdownNotifier shutdownNotifier;
 
   private static final String PREFIX = "term_"; // for termnames
   private static final UniqueIdGenerator termIdGenerator =
@@ -63,11 +62,10 @@ abstract class SmtInterpolAbstractProver<T> extends AbstractProver<T> {
       Script pEnv,
       Set<ProverOptions> options,
       ShutdownNotifier pShutdownNotifier) {
-    super(options);
+    super(pShutdownNotifier, options);
     mgr = pMgr;
     creator = pMgr.getFormulaCreator();
     env = pEnv;
-    shutdownNotifier = pShutdownNotifier;
     annotatedTerms.add(PathCopyingPersistentTreeMap.of());
   }
 
@@ -109,7 +107,7 @@ abstract class SmtInterpolAbstractProver<T> extends AbstractProver<T> {
     // by using a shutdown listener. However, SmtInterpol resets the
     // mStopEngine flag in DPLLEngine before starting to solve,
     // so we check here, too.
-    shutdownNotifier.shutdownIfNecessary();
+    proverShutdownManager.getNotifier().shutdownIfNecessary();
 
     LBool result = env.checkSat();
     switch (result) {
@@ -127,7 +125,9 @@ abstract class SmtInterpolAbstractProver<T> extends AbstractProver<T> {
             // SMTInterpol catches OOM, but we want to have it thrown.
             throw new OutOfMemoryError("Out of memory during SMTInterpol operation");
           case CANCELLED:
-            shutdownNotifier.shutdownIfNecessary(); // expected if we requested termination
+            proverShutdownManager
+                .getNotifier()
+                .shutdownIfNecessary(); // expected if we requested termination
             throw new SMTLIBException("checkSat returned UNKNOWN with unexpected reason " + reason);
           default:
             throw new SMTLIBException("checkSat returned UNKNOWN with unexpected reason " + reason);
@@ -244,10 +244,17 @@ abstract class SmtInterpolAbstractProver<T> extends AbstractProver<T> {
     // by using a shutdown listener. However, SmtInterpol resets the
     // mStopEngine flag in DPLLEngine before starting to solve,
     // so we check here, too.
-    shutdownNotifier.shutdownIfNecessary();
+    proverShutdownManager.getNotifier().shutdownIfNecessary();
     for (Term[] model : env.checkAllsat(importantTerms)) {
       callback.apply(Collections3.transformedImmutableListCopy(model, creator::encapsulateBoolean));
     }
     return callback.getResult();
   }
+
+  /* TODO: the shutdownNotifier is bound to the Script. But this is done in the context. It seems
+       like it might be re-usable after a shutdown.
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
+  }*/
 }
--- a/src/org/sosy_lab/java_smt/test/TimeoutTest.java
+++ b/src/org/sosy_lab/java_smt/test/TimeoutTest.java
@@ -8,6 +8,7 @@
 
 package org.sosy_lab.java_smt.test;
 
+import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.TruthJUnit.assume;
 import static org.junit.Assert.assertThrows;
 
@@ -24,6 +25,7 @@ import org.junit.runners.Parameterized.Parameters;
 import org.sosy_lab.java_smt.SolverContextFactory.Solvers;
 import org.sosy_lab.java_smt.api.BasicProverEnvironment;
 import org.sosy_lab.java_smt.api.BooleanFormula;
+import org.sosy_lab.java_smt.api.SolverException;
 import org.sosy_lab.java_smt.api.Tactic;
 import org.sosy_lab.java_smt.solvers.opensmt.Logics;
 
@@ -90,43 +92,97 @@ public class TimeoutTest extends SolverBasedTest0 {
   @Test(timeout = TIMEOUT_MILLISECONDS)
   public void testProverTimeoutInt() throws InterruptedException {
     requireIntegers();
-    testBasicProverTimeoutInt(() -> context.newProverEnvironment());
+    testBasicContextTimeoutInt(() -> context.newProverEnvironment());
   }
 
   @Test(timeout = TIMEOUT_MILLISECONDS)
   public void testProverTimeoutBv() throws InterruptedException {
     requireBitvectors();
+    testBasicContextTimeoutBv(() -> context.newProverEnvironment());
+  }
+
+  // Test shutdown of prover specific shutdown manager. The context should still be usable!
+  @Test(timeout = TIMEOUT_MILLISECONDS)
+  public void testProverInterruptWithSubsequentNewProverUsageBv()
+      throws InterruptedException, SolverException {
+    requireBitvectors();
+    requireIsolatedProverShutdown();
+
     testBasicProverTimeoutBv(() -> context.newProverEnvironment());
+    assertThat(shutdownManager.getNotifier().shouldShutdown()).isFalse();
+
+    HardBitvectorFormulaGenerator gen = new HardBitvectorFormulaGenerator(bvmgr, bmgr);
+    try (BasicProverEnvironment<?> pe = context.newProverEnvironment()) {
+      pe.push(gen.generate(8));
+      assertThat(pe.isUnsat()).isTrue();
+    }
+  }
+
+  // Test shutdown of prover specific shutdown manager. The context should still be usable!
+  @Test(timeout = TIMEOUT_MILLISECONDS)
+  public void testProverInterruptWithSubsequentNewProverUsageInt()
+      throws InterruptedException, SolverException {
+    requireIntegers();
+    requireIsolatedProverShutdown();
+
+    testBasicProverTimeoutInt(() -> context.newProverEnvironment());
+    assertThat(shutdownManager.getNotifier().shouldShutdown()).isFalse();
+
+    HardIntegerFormulaGenerator gen = new HardIntegerFormulaGenerator(imgr, bmgr);
+    try (BasicProverEnvironment<?> pe = context.newProverEnvironment()) {
+      pe.push(gen.generate(8));
+      assertThat(pe.isUnsat()).isTrue();
+    }
   }
 
   @Test(timeout = TIMEOUT_MILLISECONDS)
   public void testInterpolationProverTimeout() throws InterruptedException {
     requireInterpolation();
     requireIntegers();
-    testBasicProverTimeoutInt(() -> context.newProverEnvironmentWithInterpolation());
+    testBasicContextTimeoutInt(() -> context.newProverEnvironmentWithInterpolation());
   }
 
   @Test(timeout = TIMEOUT_MILLISECONDS)
   public void testOptimizationProverTimeout() throws InterruptedException {
     requireOptimization();
     requireIntegers();
-    testBasicProverTimeoutInt(() -> context.newOptimizationProverEnvironment());
+    testBasicContextTimeoutInt(() -> context.newOptimizationProverEnvironment());
   }
 
+  /** Shuts down the shutdown manager of the context. */
+  private void testBasicContextTimeoutInt(Supplier<BasicProverEnvironment<?>> proverConstructor)
+      throws InterruptedException {
+    HardIntegerFormulaGenerator gen = new HardIntegerFormulaGenerator(imgr, bmgr);
+    testBasicContextBasedTimeout(proverConstructor, gen.generate(200));
+  }
+
+  /** Shuts down the shutdown manager of the context. */
+  private void testBasicContextTimeoutBv(Supplier<BasicProverEnvironment<?>> proverConstructor)
+      throws InterruptedException {
+    HardBitvectorFormulaGenerator gen = new HardBitvectorFormulaGenerator(bvmgr, bmgr);
+    testBasicContextBasedTimeout(proverConstructor, gen.generate(200));
+  }
+
+  /**
+   * Shuts down the shutdown manager of the prover. Throws an exception for non-supporting solvers.
+   */
   private void testBasicProverTimeoutInt(Supplier<BasicProverEnvironment<?>> proverConstructor)
       throws InterruptedException {
     HardIntegerFormulaGenerator gen = new HardIntegerFormulaGenerator(imgr, bmgr);
-    testBasicProverTimeout(proverConstructor, gen.generate(200));
+    testBasicProverBasedTimeout(proverConstructor, gen.generate(200));
   }
 
+  /**
+   * Shuts down the shutdown manager of the prover. Throws an exception for non-supporting solvers.
+   */
   private void testBasicProverTimeoutBv(Supplier<BasicProverEnvironment<?>> proverConstructor)
       throws InterruptedException {
     HardBitvectorFormulaGenerator gen = new HardBitvectorFormulaGenerator(bvmgr, bmgr);
-    testBasicProverTimeout(proverConstructor, gen.generate(200));
+    testBasicProverBasedTimeout(proverConstructor, gen.generate(200));
   }
 
   @SuppressWarnings("CheckReturnValue")
-  private void testBasicProverTimeout(
+  private void testBasicContextBasedTimeout(
       Supplier<BasicProverEnvironment<?>> proverConstructor, BooleanFormula instance)
       throws InterruptedException {
     Thread t =
@@ -141,6 +197,27 @@ public class TimeoutTest extends SolverBasedTest0 {
             });
 
     try (BasicProverEnvironment<?> pe = proverConstructor.get()) {
+      pe.push(instance);
+      t.start();
+      assertThrows(InterruptedException.class, pe::isUnsat);
+    }
+  }
+
+  private void testBasicProverBasedTimeout(
+      Supplier<BasicProverEnvironment<?>> proverConstructor, BooleanFormula instance)
+      throws InterruptedException {
+
+    try (BasicProverEnvironment<?> pe = proverConstructor.get()) {
+      Thread t =
+          new Thread(
+              () -> {
+                try {
+                  Thread.sleep(delay);
+                  pe.getShutdownManagerForProver().requestShutdown("Shutdown Request");
+                } catch (InterruptedException exception) {
+                  throw new UnsupportedOperationException("Unexpected interrupt", exception);
+                }
+              });
       pe.push(instance);
       t.start();
       assertThrows(InterruptedException.class, pe::isUnsat);
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractProverWithAllSat.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractProverWithAllSat.java
@@ -28,16 +28,14 @@ import org.sosy_lab.java_smt.api.SolverException;
  */
 public abstract class AbstractProverWithAllSat<T> extends AbstractProver<T> {
 
-  protected final ShutdownNotifier shutdownNotifier;
   private final BooleanFormulaManager bmgr;
 
   protected AbstractProverWithAllSat(
       Set<ProverOptions> pOptions,
       BooleanFormulaManager pBmgr,
       ShutdownNotifier pShutdownNotifier) {
-    super(pOptions);
+    super(pShutdownNotifier, pOptions);
     bmgr = pBmgr;
-    shutdownNotifier = pShutdownNotifier;
   }
 
   @Override
@@ -68,7 +66,7 @@ public abstract class AbstractProverWithAllSat<T> extends AbstractProver<T> {
       AllSatCallback<R> callback, List<BooleanFormula> importantPredicates)
       throws SolverException, InterruptedException {
     while (!isUnsat()) {
-      shutdownNotifier.shutdownIfNecessary();
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
 
       ImmutableList.Builder<BooleanFormula> valuesOfModel = ImmutableList.builder();
       try (Evaluator evaluator = getEvaluatorWithoutChecks()) {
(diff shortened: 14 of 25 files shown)


src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -202,4 +204,18 @@ interface AllSatCallback<R> {
...
+   * SolverContext}, resulting in shutdown when the {@link SolverContext}s notifier is shutting
+   * down. The notifier returned here can be shut down independently of the creating contexts
+   * notifier.
+   *
+   * @return a {@link ShutdownManager} who is the child of the {@link ShutdownNotifier} used in the
+   *     creating {@link SolverContext}, that can be used to shut down only this {@link
+   *     ProverEnvironment}.
+   * @throws UnsupportedOperationException if the solver does not support prover specific shutdown.
+   */
+  ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException;
```
kfriedberger: Due to API change, better provide a default implementation that throws the UnsupportedException.

Additionally, his method is a plain getter. Why does a plain getter throw such an exception? The getter itself does not depend on solver-features that could be unsupported, but only on JavaSMT-internal code. Perhaps just remove the exception from signature.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -158,4 +167,17 @@ public void close() {
     closeAllEvaluators();
     closed = true;
   }
+
+  @Override
+  public ShutdownManager getShutdownManagerForProver() throws UnsupportedOperationException {
+    return getShutdownManagerForProverImpl();
+  }
+
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
```
kfriedberger: As this method is not generic, we do not need it. just use inline it in the normal `getShutdoenManagerForProver`.
With a default implementation in the API, we can even remove this implmentation completely and just mark the overridden method as `abstract`.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
```diff
@@ -38,7 +39,10 @@ class BitwuzlaTheoremProver extends AbstractProverWithAllSat<Void> implements Pr
       new Terminator() {
         @Override
         public boolean terminate() {
-          return shutdownNotifier.shouldShutdown(); // shutdownNotifer is defined in the superclass
+          return proverShutdownManager
+              .getNotifier()
+              .shouldShutdown(); // shutdownNotifer is defined in the
+          // superclass
```
kfriedberger: The comment is formatted ugly, and it might not be understandable, please rephrase.

Is `getNotifier()` a "fast" method call or does it construct any objects? The callback is expected to be executed quite often from native solver code. We could store the notifier directly within the class and just reuse it.

Addtiionally, we already have a method for getting the manager. Why do we then use direct access here? Please use the getter.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
```diff
@@ -119,7 +123,8 @@ private boolean readSATResult(Result resultValue) throws SolverException, Interr
       return false;
     } else if (resultValue == Result.UNSAT) {
       return true;
-    } else if (resultValue == Result.UNKNOWN && shutdownNotifier.shouldShutdown()) {
+    } else if (resultValue == Result.UNKNOWN
+        && proverShutdownManager.getNotifier().shouldShutdown()) {
```
kfriedberger: Is it guaranteed that the manager always returns the same notifier-instance and the notifier always returns the same flag?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
```diff
@@ -246,6 +251,11 @@ protected BitwuzlaModel getEvaluatorWithoutChecks() {
             Collections2.transform(getAssertedFormulas(), creator::extractInfo)));
   }
 
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
+  }
```
kfriedberger: One could move this method higher up in the hierarchy and just provide it in the abstract super-class.

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
```diff
@@ -49,7 +49,7 @@ protected BoolectorAbstractProver(
     this.manager = manager;
     this.creator = creator;
     this.btor = btor;
-    terminationCallback = shutdownNotifier::shouldShutdown;
+    terminationCallback = proverShutdownManager.getNotifier()::shouldShutdown;
```
kfriedberger: Just for comparison to above: Here, we keep the same notifier and reuse it for all callback-calls.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
```diff
@@ -100,21 +98,33 @@ private long buildConfig(Set<ProverOptions> opts) {
...
   }
 
+  /**
+   * Get a termination callback for the current prover (who's parent is the contexts callback). The
+   * callback can be registered upfront, i.e., before calling a possibly expensive computation in
+   * the solver to allow a proper shutdown.
+   */
+  private TerminationCallback getTerminationTest() {
+    return () -> {
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
```
kfriedberger: We could extract the reference to the notifier and just reuse.

Might look like this:


```
final ShutdownNotifier notifier = getShutdownManagerForProver().getNotifier()
return () -> {
  notifier.shutdownIfNecessary();
  return false;
};
```

src/org/sosy_lab/java_smt/test/TimeoutTest.java
```diff
@@ -146,4 +202,25 @@ private void testBasicProverTimeout(
...
+                  pe.getShutdownManagerForProver().requestShutdown("Shutdown Request");
+                } catch (InterruptedException exception) {
+                  throw new UnsupportedOperationException("Unexpected interrupt", exception);
+                }
+              });
+      pe.push(instance);
+      t.start();
+      assertThrows(InterruptedException.class, pe::isUnsat);
+    }
+  }
```
kfriedberger, line 597: None of the tests checks whether the solver is usable after a shutdown.
None of the tests checks that, out of multiple parallel instances, if one is terminated, then all other provers remain running and deliver results.

Could you add such tests?

kfriedberger (commented):
Overall, I find the idea good. It matches our architectural guidelines and solves a clear problem. The PR needs some minor imporovements.

kfriedberger:
This PR might provide a solution for a quite old issue: https://github.com/sosy-lab/java-smt/issues/100 :-)

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: Thanks for the feedback! I removed the exception from the signature and added a default implementation.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: True. I imagined that i need more solver specific code, but i didn't. I removed it. Thanks for the feedback!

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: I removed the comment, as the new default should be obvious.

`getNotifier()` does not construct any objects, it just returns the notifier that is created when the `ShutdownManager` is created. You can argue that it is faster to have the reference directly, but this is built once when the prover is created. I changed it so that the notifier is generally accessible directly in the superclass.

Thanks for the feedback!

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: Yes! The notifier is created final and only once, see here: `private final ShutdownNotifier notifier = new ShutdownNotifier(this);` The flag can not be changed once a shutdown is triggered.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: I removed it completely, as we didn't really need it.

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
baierd: We build a new manager in every prover creation (see [here](https://github.com/sosy-lab/java-smt/pull/489/files#diff-3ef1d1629e51825e4d06bdc6fe00b365066088658fd72432476e4e5d1f50c220R67)) and then use this new managers notifier to request shutdowns. So we always use the one notifier bound to the one manager of the prover (that is a child of the contexts `ShutdownNotifier`)

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
baierd: Done! Thanks for the hint.

src/org/sosy_lab/java_smt/test/TimeoutTest.java
baierd: I was just writing some tests and noticed that we do not really define how a solver (context and prover) should behave after a context-wide shutdown is requested.
Should prover creation throw a exception if the context is already shut down?
What operations are still usable and guaranteed to be correct/safe after a shutdown in your current prover?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
kfriedberger: Referencing some existing source code is no "guarantee".
The guarantee would be a JavaDoc specification for the ShutdownManager-API, which is missing:
https://sosy-lab.github.io/java-common-lib/api/org/sosy_lab/common/ShutdownManager.html#getNotifier()

A user can provide his own implementation of the shutdown-manager and change its behaviour.
I really would prefer to keep a local reference to this notifier.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
```diff
@@ -202,11 +203,12 @@ public boolean isUnsat() throws InterruptedException, SolverException {
...
     Result result;
-    try (ShutdownHook hook = new ShutdownHook(shutdownNotifier, smtEngine::interrupt)) {
-      shutdownNotifier.shutdownIfNecessary();
+    try (ShutdownHook hook =
+        new ShutdownHook(proverShutdownManager.getNotifier(), smtEngine::interrupt)) {
+      proverShutdownManager.getNotifier().shutdownIfNecessary();
       result = smtEngine.checkSat();
     }
-    shutdownNotifier.shutdownIfNecessary();
+    proverShutdownManager.getNotifier().shutdownIfNecessary();
```
kfriedberger: I would prefer to avoid calling `getNotifier()` more than once here.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
PhilippWendler: There is no intention of a guarantee that `getNotifier()` always returns the same instance, but there is definitively the intention that this does not matter and that all notifiers handed out by a manager always return the shutdown state of the manager, and that `shouldShutdown()` will keep returning `true` forever after it has returned `true` once.

Since the split in separate manager and notifier classes this was not fully written down in the JavaDoc, but I improved this now: https://github.com/sosy-lab/java-common-lib/commit/108faa65ec0ec60a085232de43831bac1d453db3

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -202,4 +204,24 @@ interface AllSatCallback<R> {
...
+   * SolverContext}, resulting in shutdown when the {@link SolverContext}s notifier is shutting
+   * down. The notifier returned here can be shut down independently of the creating contexts
+   * notifier.
+   *
+   * @return a {@link ShutdownManager} who is the child of the {@link ShutdownNotifier} used in the
+   *     creating {@link SolverContext}, that can be used to shut down only this {@link
+   *     ProverEnvironment}.
+   * @throws UnsupportedOperationException if the solver does not support prover specific shutdown.
+   */
+  default ShutdownManager getShutdownManagerForProver() {
```
PhilippWendler: This is kind of backwards compared to how the API of `ShutdownManager` was designed and is supposed to be used. Typically, shutdown requests flow downwards in the call graph: For example, everything in an application should terminate because of an external request like Ctrl+C, some component should terminate because the caller wants a time limit, etc. What is unwanted are (accidental) shutdown requests from a component to its "sister" components or its parent. Of course, this is especially important in case of larger hierarchies with layered shutdown managers and notifiers, because if a component terminates as a side effect of some other unrelated component (like a "nephew" in the call graph) shutting down, this easily creates a hard-to-debug mess. 

The `ShutdownManager` API is designed such that this architecture is easy to implement and (accidental) misuse is hard / harder than correct use, and also such that it is easy to see and understand how shutdown requests can and cannot flow. If you write or see a component creating a `ShutdownManager` and handing out `ShutdownNotifier`s to the components it calls, you can be sure that only the parent component decides when to shutdown. Thus, `ShutdownManager`s are best kept as private property of the component that creates them. Then the owner of the `ShutdownManager` as visible in the code is obviously the same as the conceptual owner of the `ShutdownManager` (i.e., who is allowed to trigger shutdowns).

The suggested API here breaks with this pattern. This means that it will be harder to understand which component will issue shutdown requests and where do the shutdown requests flow. For example, right now it is unclear whether JavaSMT would ever trigger shutdown requests on the returned `ShutdownManager` on its own, which would prevent client code from using it for any other purpose, or whether JavaSMT will not do this and thus client code could reuse this `ShutdownManager`'s notifier for other components that it also needs to terminate at the same time as the prover. And even if an answer to this question is added to the documentation, it remains in the documentation. If the client code would create the `ShutdownManager` on its own and only pass a `ShutdownNotifier` to JavaSMT, the answer to this question would be obvious from the code.

It also makes it harder to use the API. For me, one of the most obvious use cases of this feature is for example a time limit for an unsat+interpolation query during CEGAR refinement of a verifier (if the query is too expensive, we might want to try refining a different path first, or use a different strategy to get predicates). But the component that defines and handles the time limit would not be the immediate client code of JavaSMT that deals with low-level formula and prover handling, but something higher up the call chain that is concerned with handling the overall refinement procedure and switching between different refinement strategies (for example, in CPAchecker these two classes would be `InterpolationManager` and `PredicateCPARefiner`, respectively).

This means I am left with two suboptimal choices: (A) I could also implement getters to pass the `ShutdownManager` created by JavaSMT upwards in the call graph from the prover-handling component to the component that defines the time limit / shutdown requests. But then I get a confusing mixture of some shutdown notifiers being passed downwards (those for the context) and some upwards (those for the provers), and it becomes easy to accidentally use the latter for some other component and create such a "nephew" situation as described above. I also always have to first create the lower-level component, then call the getter for the `ShutdownManager`, then register it for a time limit (or other shutdown requests) somewhere, and then call the lower-level component again to let it do its actual work. This would be incompatible with an API that just consists of a single method like `doWorkInProverUntilShutdownIsRequested(SolverContext, Formula, ShutdownNotifier)`. And of course, the more layers are between the prover-handling component (e.g., `InterpolationManager`) and the shutdown-requesting component (e.g., `PredicateCPARefiner`), the more complex this becomes because on every layer I have to implement all of this again. The (not-so-obvious) alternative (B) would be to still pass `ShutdownNotifier`s only downwards in my code, and then in the low-level class that calls JavaSMT manually connect the `ShutdownNotifier` that was received as a parameter with the `ShutdownManager` that was created by JavaSMT. But of course this is still suboptimal, and one can forget to keep a reference to the `ShutdownRequestListener` that connects the two, and then the GC would break this connection.

Besides the concrete effect on callers of this API, I also see a further effect: Adding this API effectively communicates to JavaSMT that is fine to pass around `ShutdownManager`s and obscure the connection between the conceptual owner of the manager and the code that handles it. But we do not want to train users to think so, we want to discourage them from doing so.

Also note the inconsistency between the context, which accepts a `ShutdownNotifier` as parameter, and the prover, which returns a `ShutdownManager`.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: @PhilippWendler i just noticed that there is/can be a delay in between the shutdown being requested and the children being informed. See the 2 tests [here](https://github.com/sosy-lab/java-smt/commit/e73b88924df72d1807869ec29a74b5297fadc976#diff-9166e448ca2535d7e132659d20b7dd6f9d3e5ef9651026a627f4aa52fed3ebbcR180-R234). Is this intentional?

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
baierd: I just forgot to push everything. Should be removed now. Sorry for the delay.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: I see. 
We can of course change it so that we return only a notifier and take a manager or notifier when creating a prover.
The only reason i did it this way is that i can be sure that the contexts notifier is the parent of our notifier in the prover.

Would it be possible to add the ability to get or set a parent in the `ShutdownNotifier`?
I understand why a setter would maybe not be a good idea. But i guess it would be fine if i can at least check that the contexts notifier is the parent of the notifier given as a prover argument.

baierd:
I cleaned up the default behavior of the prover API (isUnsat(), getModel(), getUnsatCore() etc.) so that they share the common prerequisites and no inconsistencies within the solver implementations exist.

I found 2 things i want to discuss:
1. The JavaDoc of `unsatCoreOverAssumptions()` states that: `Returns an UNSAT core (if it exists, otherwise {@code Optional.empty()}), over the chosen assumptions. Does NOT require the {@link ProverOptions#GENERATE_UNSAT_CORE} option to work.` But we have and check for the option `GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS`. So i guess the JavaDoc should be updated?
2. the 2 public methods `getSeqInterpolants0()` and `getTreeInterpolants0()` are not named well in my opinion. I suggest better names like `getSingletonSeqInterpolants)`? (i am happy about better name suggestions ;D)

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
PhilippWendler: I am not sure what exactly you are referring to. But if you have multi-threaded code, then you can of course never be sure when each of the threads is scheduled in relation to the others.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
PhilippWendler: > I see. We can of course change it so that we return only a notifier and take a manager or notifier when creating a prover.

Why even return something? My suggestion would be to do the same as every other case where this API is used: just accept a `ShutdownNotifier` as parameter.
But most importantly: Do not accept a `ShutdownManager` as parameter. It should be clear from my post that this also has the same drawbacks of unclear ownership as the current code.

> The only reason i did it this way is that i can be sure that the contexts notifier is the parent of our notifier in the prover.
> 
> Would it be possible to add the ability to get or set a parent in the `ShutdownNotifier`? I understand why a setter would maybe not be a good idea. But i guess it would be fine if i can at least check that the contexts notifier is the parent of the notifier given as a prover argument.

Do you really want to force users to build a `ShutdownNotifier` that is a child of the context's one? Consider the use case that I described above, with a particular time limit for the refinement step, but broader use of the context. In this case I could imagine easily that due to the way how components in CPAchecker are created, the `ShutdownNotifier`s for context and prover are maybe not in a parent-child relation ship, just both children of some other `ShutdownNotifier`, even though in practice likely the context's one always also triggers if the prover's one triggers. Then the caller code would be back to being forced to manually link two `ShutdownNotifier`s together.

The easiest way for users of JavaSMT, and the API with the least potential for misuse, would of course be to simply accept an arbitrary `ShutdownNotifier`. And the prover would be terminated if either the given `ShutdownNotifier` triggers or the context terminates.

If this is too difficult to implemented or unwanted, maybe a less optimal API has to be chosen. But so far I have not seen arguments indicating this.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: Yes, but i would have expected that the chain of signals in the shutdown notifiers/managers is built in such a way that they are all informed in a synchronized way.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: We can at any point just check the shutdown signal for context and prover notifiers. I'll test it out and report back!

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
PhilippWendler: What do you mean with "synchronized way"? "synchronized" is not some magic global property, one can only synchronize against specific things.

If you want help, you need to be more specific. What is the sequence of operations that you are observing and that is unexpected/unwanted?

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -22,8 +22,13 @@
...
  */
 public interface BasicProverEnvironment<T> extends AutoCloseable {
+  String NO_MODEL_HELP = "Model computation failed. Are the pushed formulae " + "satisfiable?";
 
-  String NO_MODEL_HELP = "Model computation failed. Are the pushed formulae satisfiable?";
+  String NO_UNSAT_CORE_HELP =
+      "UnsatCore computation failed. Are the pushed formulae " + "unsatisfiable?";
+
+  String STACK_CHANGED_HELP =
+      "Computation failed. The prover state has changed since the last " + "call to isUnsat().";
```
daniel-raffler: Could you concatenate those Strings?

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,6 +66,28 @@ enum ProverOptions {
...
+   *     provers can be created and used. If a {@link ShutdownNotifier} has been given to the
+   *     context that is used to call this method, both notifiers can be used to stop the prover
+   *     returned by this method. Note that once a shutdown-request has been given to the contexts
+   *     {@link ShutdownNotifier}, no prover can ever be used again on that context instance.
+   *     Solvers that don't support isolated prover shutdown throw a {@link
+   *     UnsupportedOperationException} for this method and {@link
+   *     #newProverEnvironment(ProverOptions...)} should be used instead.
+   * @param options Options specified for the prover environment. All the options specified in
+   *     {@link ProverOptions} are turned off by default.
+   */
```
daniel-raffler: So far we've left it pretty open if a solver can be reused after a shutdown request. Should the JavaDoc for `SolverContextFactory` also be updated?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
```diff
@@ -8,22 +8,167 @@
...
+    }
+    super.close();
+  }
+
+  /*
+   * Boolector should throw its own exceptions that tell you what went wrong!
+   */
+  @Override
+  protected boolean isUnsatImpl() throws SolverException, InterruptedException {
+    Preconditions.checkState(!closed);
```
daniel-raffler: Do we still need to check here?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
```diff
@@ -8,22 +8,167 @@
...
+
+  @Override
+  protected void pushImpl() throws InterruptedException {
+    BtorJNI.boolector_push(manager.getEnvironment(), 1);
+  }
+
+  @Override
+  protected boolean isUnsatWithAssumptionsImpl(Collection<BooleanFormula> pAssumptions)
+      throws SolverException, InterruptedException {
+    Preconditions.checkState(!closed);
```
daniel-raffler: Same

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
```diff
@@ -8,22 +8,167 @@
...
+    for (BooleanFormula assumption : pAssumptions) {
+      BtorJNI.boolector_assume(btor, BoolectorFormulaManager.getBtorTerm(assumption));
+    }
+    return isUnsat();
+  }
+
+  @SuppressWarnings("resource")
+  @Override
+  protected Model getModelImpl() throws SolverException {
+    Preconditions.checkState(wasLastSatCheckSat, NO_MODEL_HELP);
```
daniel-raffler: Same

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
```diff
@@ -228,10 +200,9 @@ private boolean convertSatResult(Result result) throws InterruptedException, Sol
   }
 
   @Override
-  public List<BooleanFormula> getUnsatCore() {
-    Preconditions.checkState(!closed);
-    checkGenerateUnsatCores();
-    Preconditions.checkState(!changedSinceLastSatQuery);
+  protected List<BooleanFormula> getUnsatCoreImpl() {
+    Preconditions.checkState(!wasLastSatCheckSat);
+    Preconditions.checkState(!stackChangedSinceLastQuery);
```
daniel-raffler: Are these checks still necessary?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -220,10 +200,9 @@ private boolean convertSatResult(Result result) throws InterruptedException, Sol
   }
 
   @Override
-  public List<BooleanFormula> getUnsatCore() {
-    Preconditions.checkState(!closed);
-    checkGenerateUnsatCores();
-    Preconditions.checkState(!changedSinceLastSatQuery);
+  protected List<BooleanFormula> getUnsatCoreImpl() {
+    Preconditions.checkState(!stackChangedSinceLastQuery);
+    Preconditions.checkState(!wasLastSatCheckSat);
```
daniel-raffler: Same

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
```diff
@@ -251,4 +230,10 @@ public void close() {
     }
     super.close();
   }
+
+  /* TODO: revisit once CVC5 supports interruption
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
+  }*/
```
daniel-raffler: I think this can be removed now?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5SolverContext.java
```diff
@@ -215,11 +216,19 @@ public Solvers getSolverName() {
...
   @Override
-  public ProverEnvironment newProverEnvironment0(Set<ProverOptions> pOptions) {
+  public ProverEnvironment newProverEnvironment0(
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> pOptions) {
     Preconditions.checkState(!closed, "solver context is already closed");
+
+    if (pProverShutdownNotifier != null) {
+      // TODO: CVC5 does not support shutdown at all currently. Remove completely?
+      throw new UnsupportedOperationException("CVC5 does not support interruption of provers");
+    }
```
daniel-raffler, line 234: I'm not sure an exception should be thrown here. We accept a shutdown notifier for the context, and this seems no different. A shutdown request doesn't guarantee that the solver will abort immediately, but just that it will do so when it's convenient. So it should still be fine if we just check the shutdown notifier manually every now and then.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5SolverContext.java
```diff
@@ -233,11 +242,18 @@ protected boolean supportsAssumptionSolving() {
...
   @Override
   protected InterpolatingProverEnvironment<?> newProverEnvironmentWithInterpolation0(
-      Set<ProverOptions> pOptions) {
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> pOptions) {
     Preconditions.checkState(!closed, "solver context is already closed");
+
+    if (pProverShutdownNotifier != null) {
+      // TODO: CVC5 does not support shutdown at all currently. Remove completely?
+      throw new UnsupportedOperationException("CVC5 does not support interruption of provers");
+    }
```
daniel-raffler, line 259: Same

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
```diff
@@ -263,22 +242,21 @@ public boolean isUnsat() throws InterruptedException, SolverException {
   }
 
   @Override
-  public List<BooleanFormula> getUnsatCore() {
-    Preconditions.checkState(!closed);
-    checkGenerateUnsatCores();
-    Preconditions.checkState(!changedSinceLastSatQuery);
+  protected List<BooleanFormula> getUnsatCoreImpl() {
+    Preconditions.checkState(!wasLastSatCheckSat);
+    Preconditions.checkState(!stackChangedSinceLastQuery);
```
daniel-raffler: Can those be removed?

src/org/sosy_lab/java_smt/solvers/princess/PrincessSolverContext.java
```diff
@@ -74,7 +74,13 @@ public static SolverContext create(
 
   @SuppressWarnings("resource")
   @Override
-  protected ProverEnvironment newProverEnvironment0(Set<ProverOptions> options) {
+  protected ProverEnvironment newProverEnvironment0(
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> options) {
+
+    if (pProverShutdownNotifier != null) {
+      throw new UnsupportedOperationException("Shutdown is not supported for Princess.");
+    }
```
daniel-raffler, line 82: Same as for CVC5: I'm not sure we need an exception here

src/org/sosy_lab/java_smt/solvers/princess/PrincessSolverContext.java
```diff
@@ -85,14 +91,19 @@ protected ProverEnvironment newProverEnvironment0(Set<ProverOptions> options) {
   @SuppressWarnings("resource")
   @Override
   protected InterpolatingProverEnvironment<?> newProverEnvironmentWithInterpolation0(
-      Set<ProverOptions> options) {
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> options) {
+
+    if (pProverShutdownNotifier != null) {
+      throw new UnsupportedOperationException("Shutdown is not supported for Princess.");
+    }
```
daniel-raffler, line 98: Same

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolAbstractProver.java
```diff
@@ -244,10 +238,17 @@ public <R> R allSat(AllSatCallback<R> callback, List<BooleanFormula> important)
...
     }
     return callback.getResult();
   }
+
+  /* TODO: the shutdownNotifier is bound to the Script. But this is done in the context. It seems
+       like it might be re-usable after a shutdown.
+  @Override
+  protected ShutdownManager getShutdownManagerForProverImpl() throws UnsupportedOperationException {
+    return proverShutdownManager;
+  }*/
```
daniel-raffler: Do we still need this?

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolSolverContext.java
```diff
@@ -226,26 +226,46 @@ private SMTInterpol getSmtInterpol() {
 
   @SuppressWarnings("resource")
   @Override
-  protected ProverEnvironment newProverEnvironment0(Set<ProverOptions> options) {
+  protected ProverEnvironment newProverEnvironment0(
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> options) {
+
+    if (pProverShutdownNotifier != null) {
+      // TODO: check re-usability of SMTInterpol after shutdown.
```
daniel-raffler, line 233: Have you looked into that yet?

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolSolverContext.java
```diff
@@ -226,26 +226,46 @@ private SMTInterpol getSmtInterpol() {
...
   }
 
   @SuppressWarnings("resource")
   @Override
   protected InterpolatingProverEnvironment<?> newProverEnvironmentWithInterpolation0(
-      Set<ProverOptions> options) {
+      @Nullable ShutdownNotifier pProverShutdownNotifier, Set<ProverOptions> options) {
+
+    if (pProverShutdownNotifier != null) {
+      // TODO: check re-usability of SMTInterpol after shutdown.
```
daniel-raffler, line 249: Same

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
```diff
@@ -67,7 +67,8 @@ public ImmutableList<ValueAssignment> asList() {
         Native.decRef(z3context, funcDecl);
       }
     } catch (Z3Exception e) {
-      throw z3creator.handleZ3ExceptionAsRuntimeException(e);
+      // TODO: Do we need the prover shutdown notifier here?
```
daniel-raffler: Is this working now? Or do we still need to add something?

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
```diff
@@ -388,7 +389,8 @@ protected Long evalImpl(Long formula) {
     try {
       satisfiableModel = Native.modelEval(z3context, model, formula, false, resultPtr);
     } catch (Z3Exception e) {
-      throw z3creator.handleZ3ExceptionAsRuntimeException(e);
+      // TODO: Do we need the prover shutdown notifier here?
```
daniel-raffler: Same

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
baierd: I was merely curious ;D
We discussed this offline sufficiently. Thank you!

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: The current version should resolve the issues brought up by you. Could you take another look?

We can now hand over an optional notifier argument to prover creation.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: Done. Thanks for spotting!

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Good question. @kfriedberger do we want to put out a policy or just give out all info that we have and leave it to the users?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
baierd: Nope. Thanks for spotting, i removed it!

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
baierd: Nope. Thanks for spotting, i removed it!

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorTheoremProver.java
baierd: Nope. Thanks for spotting, i removed it!

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolAbstractProver.java
baierd: Nope. Thanks for spotting, i removed it!

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
baierd: I fixed it, should be fine now.

src/org/sosy_lab/java_smt/solvers/z3/Z3Model.java
baierd: I fixed it, should be fine now.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5SolverContext.java
baierd: Yes and we should not be accepting a notifier if it can't interrupt. So the correct way of handling this is actually removing the context notifier.

src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtAbstractProver.java
baierd: Yes! Thanks for spotting, removed.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
baierd: Yes! Thanks for spotting, removed.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5AbstractProver.java
baierd: Yes! Thanks for spotting, removed.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4TheoremProver.java
baierd: No! Thanks for spotting, removed.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -22,9 +22,17 @@
...
   String NO_MODEL_HELP = "Model computation failed. Are the pushed formulae satisfiable?";
 
+  String NO_UNSAT_CORE_HELP =
+      "UnsatCore computation failed. Are the pushed formulae unsatisfiable?";
+
+  String STACK_CHANGED_HELP =
+      "Computation failed. The prover state has changed since the last call to isUnsat().";
+
+  // Used as prefix concatenated with the reason in the IllegalStateException thrown for shutdowns
+  String SHUTDOWN_EXCEPTION_PREFIX = "Prover is not usable due to interrupt with message: ";
```
PhilippWendler: Are these 3 strings supposed to be part of your public API? If not, they should not be here. Every user of a prover environment will see them for auto-completion, for example. (Same for `NO_MODEL_HELP` already.)

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
```diff
@@ -36,6 +37,21 @@ protected void clearAssumptions() {
...
+    for (int i = 0; i < solverAssumptionsAsFormula.size(); i++) {
+      try {
+        delegate.pop();
+      } catch (IllegalStateException ise) {
+        if (ise.getMessage().startsWith(AbstractProver.SHUTDOWN_EXCEPTION_PREFIX)) {
+          throw new InterruptedException(
+              ise.getMessage()
+                  .replace("Prover is not usable due " + "to shutdown with message: ", ""));
+        }
+      }
```
PhilippWendler: Why so complicated and even using String matching?

The catch block could simply call `shutdownIfNecessary()`, for example. Or the `IllegalStateException` could have an `InterruptedException` as cause.

Also, there is a bug here: If there is an `IllegalStateException` for a different reason, it is swallowed!

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
```diff
@@ -44,13 +60,16 @@ public void pop() {
 
   @Override
   public T addConstraint(BooleanFormula constraint) throws InterruptedException {
-    clearAssumptions();
+    // This might throw the "wrong" exception because it pops the assumptions
+    // (which does not return a InterruptedException for interrupts)
```
PhilippWendler: Why does `clearAssumptionsWithInterruptedException()` throw a wrong exception? And why is there only this comment mentioning this instead of code that fixes it? And why only for this single call to `clearAssumptionsWithInterruptedException()`?

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -47,7 +54,13 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
 
   private static final String TEMPLATE = "Please set the prover option %s.";
 
-  protected AbstractProver(Set<ProverOptions> pOptions) {
+  protected final @Nullable ShutdownNotifier proverShutdownNotifier;
```
PhilippWendler: Why make it `@Nullable` instead of using a dummy `ShutdownNotifier`? Do you think the additional volatile read on every `shutdownIfNecessary()` is too expensive compared to the null check for users who do not need this feature?

And why are the `ShutdownNotifier`s protected? Why should subclasses access them and not just `shutdownIfNecessary()`?

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -107,11 +115,7 @@ default Evaluator getEvaluator() throws SolverException {
    * <p>Note that if you need to iterate multiple times over the model it may be more efficient to
    * use this method instead of {@link #getModel()} (depending on the solver).
    */
-  default ImmutableList<Model.ValueAssignment> getModelAssignments() throws SolverException {
-    try (Model model = getModel()) {
-      return model.asList();
-    }
-  }
+  ImmutableList<Model.ValueAssignment> getModelAssignments() throws SolverException;
```
PhilippWendler: For the record: Removal of the default implementation is a breaking change if someone external has implemented `BasicProverEnvironment`. It is up to the maintainers to decide whether they want to guarantee API stability for implementors of the `api` package or not, I just wanted to make this clear.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,6 +66,28 @@ enum ProverOptions {
    */
   ProverEnvironment newProverEnvironment(ProverOptions... options);
 
+  /**
+   * Create a fresh new {@link ProverEnvironment} which encapsulates an assertion stack and can be
+   * used to check formulas for unsatisfiability. Allows the shutdown of the prover instance
+   * returned with the given {@link ShutdownNotifier}, without stopping other provers of the context
```
PhilippWendler: "Allows the shutdown" -> "Allows shutdown"

"Allows shutdown of the prover instance returned with the given ShutdownNotifier" is hard to read, I could at first not understand what "returned" refers to to. Better rephrase, for example "Allows shutdown of the returned prover instance" or better "The returned prover instance can be shut down using the given ShutdownNotifier".

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,6 +66,28 @@ enum ProverOptions {
...
+   * calling this (except for provers also connected via the given {@link ShutdownNotifier}).
+   *
+   * @param pProverShutdownNotifier a {@link ShutdownNotifier} that stops the prover returned by
+   *     this method. The prover is not usable anymore after a shutdown has been requested and only
+   *     ever returns {@link InterruptedException}s. The context can be used normally and new
```
PhilippWendler: "only ever returns" -> "in the following its methods only throw"

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,6 +66,28 @@ enum ProverOptions {
...
+   *     provers can be created and used. If a {@link ShutdownNotifier} has been given to the
+   *     context that is used to call this method, both notifiers can be used to stop the prover
+   *     returned by this method. Note that once a shutdown-request has been given to the contexts
+   *     {@link ShutdownNotifier}, no prover can ever be used again on that context instance.
+   *     Solvers that don't support isolated prover shutdown throw a {@link
+   *     UnsupportedOperationException} for this method and {@link
+   *     #newProverEnvironment(ProverOptions...)} should be used instead.
```
PhilippWendler: A lot of this information is not specific to this method and especially not to the `pProverShutdownNotifier` parameter, but it is hidden in its documentation. I suggest to write only "Note that as for all provers, the prover returned by this method can also be shut down by a shut down of the whole context using the ShutdownNotifier that was given when creating the context."

The sentence about unusable contexts should be moved to a general context documentation. The sentence about the exception should be moved to the general documentation of this method (out of `@param`) or `@throws`.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -76,6 +99,31 @@ enum ProverOptions {
    */
   InterpolatingProverEnvironment<?> newProverEnvironmentWithInterpolation(ProverOptions... options);
 
+  /**
+   * Create a fresh new {@link InterpolatingProverEnvironment} which encapsulates an assertion stack
+   * and allows generating and retrieve interpolants for unsatisfiable formulas. If the SMT solver
+   * is able to handle satisfiability tests with assumptions please consider implementing the {@link
+   * InterpolatingProverEnvironment} interface, and return an Object of this type here. Allows the
```
PhilippWendler: API documentation that is only relevant for implementors should be separated from the documentation for callers in order to not confuse the latter (of which there are many more). In particular, it should not be the second sentence of the method description when there are many more important things to say.

`@implNote` seems like a better place.

src/org/sosy_lab/java_smt/delegate/logging/PackageSanityTest.java
```diff
@@ -9,5 +9,12 @@
 package org.sosy_lab.java_smt.delegate.logging;
 
 import com.google.common.testing.AbstractPackageSanityTests;
+import org.sosy_lab.common.ShutdownManager;
 
-public class PackageSanityTest extends AbstractPackageSanityTests {}
+public class PackageSanityTest extends AbstractPackageSanityTests {
+
+  // WTF?
```
PhilippWendler, line 16:
```suggestion
```

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedBasicProverEnvironment.java
```diff
@@ -82,6 +84,11 @@ public Model getModel() throws SolverException {
     }
   }
 
+  @Override
+  public ImmutableList<ValueAssignment> getModelAssignments() throws SolverException {
+    return delegate.getModelAssignments();
```
PhilippWendler, line 91: Does this need to be synchronized?

It seems like all of the changes in `/delegate/` should be double checked whether there is something missing.

src/org/sosy_lab/java_smt/delegate/debugging/DebuggingBasicProverEnvironment.java
```diff
@@ -77,6 +79,11 @@ public Model getModel() throws SolverException {
     return new DebuggingModel(delegate.getModel(), debugging);
   }
 
+  @Override
+  public ImmutableList<ValueAssignment> getModelAssignments() throws SolverException {
+    return delegate.getModelAssignments();
```
PhilippWendler, line 87: Shouldn't this have the sanity checks?

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
```diff
@@ -97,24 +96,36 @@ private long buildConfig(Set<ProverOptions> opts) {
...
+  /**
+   * Get a termination callback for the current prover (who's parent is the contexts callback). The
+   * callback can be registered upfront, i.e., before calling a possibly expensive computation in
+   * the solver to allow a proper shutdown.
+   */
+  private TerminationCallback getTerminationTest() {
+    return () -> {
+      shutdownIfNecessary();
+      return false;
+    };
```
PhilippWendler: There is no real reason to have a method return a lambda. Simply implement the content of the lambda as a method and use a method reference instead of calling `getTerminationTest()`.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -35,6 +40,8 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
   private final boolean generateUnsatCoresOverAssumptions;
   protected final boolean enableSL;
   protected boolean closed = false;
+  protected boolean wasLastSatCheckSat = false;
+  protected boolean stackChangedSinceLastQuery = false;
```
PhilippWendler: Are you sure it is a good idea to have non-private mutable state, which is accessible by classes in ~10 different packages, and is even actually mutated by at least some of them? I would assume that it will become highly difficult to properly understand and maintain the logic for this state.

src/org/sosy_lab/java_smt/test/TimeoutTest.java
```diff
@@ -90,43 +101,348 @@ public void testTacticTimeout() {
...
+  public void testProverInterruptWithSubsequentNewProverUsageBv()
+      throws InterruptedException, SolverException {
+    requireBitvectors();
+    requireIsolatedProverShutdown();
+
+    ShutdownManager sm1 = ShutdownManager.create();
+
+    testBasicProverTimeoutBv(() -> context.newProverEnvironment(sm1.getNotifier()), sm1);
+    assertThat(shutdownManager.getNotifier().shouldShutdown()).isFalse();
+    assertThat(sm1.getNotifier().shouldShutdown()).isTrue();
```
PhilippWendler: Using `shutdownManager` and `sm1` as names for two objects of the same type is confusing and makes the test hard to understand.

src/org/sosy_lab/java_smt/test/TimeoutTest.java
```diff
@@ -90,43 +101,348 @@ public void testTacticTimeout() {
...
+
+    HardBitvectorFormulaGenerator gen = new HardBitvectorFormulaGenerator(bvmgr, bmgr);
+    try (BasicProverEnvironment<?> pe = context.newProverEnvironment()) {
+      pe.push(gen.generate(8));
+      assertThat(pe.isUnsat()).isTrue();
```
PhilippWendler, line 130: Does it really make sense to ask the solver for a hard query just in order to confirm that it still works? Will this not make the tests need unnecessarily long time?

I note that for example in https://gitlab.com/sosy-lab/software/java-smt/-/jobs/10318170445 `TimeoutTest` takes 42s, more than any other test it seems, and a significant portion of the overall test time.

Another contributing factor to this seems to be that many tests are duplicate for ints and bitvectors. While I see that both are required, I don't think that more than one of them needs to run for each solver. There could be a test method that runs either or. Or even better would be to make the test code be independent of the formula type. It looks like this would be trivial, only the formula generator differs between the tests.

baierd: I reduced the query to `3`.
`3` is very fast, as the growth is exponential. We could also just ask to solve a trivial query.

It is true that this test class takes very long now. My suggestion would be to run only a subset of the time-limits per default, and run all of them weekly in the CI for example.

baierd: Note: i pushed more tests, increasing the run-time even more, and would argue that we need a split into all possible prover variants (theorem, interpolation, optimization) to test everything, which would again increase the run-time.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -134,6 +267,25 @@ protected ImmutableSet<T> getAssertedConstraintIds() {
     return builder.build();
   }
 
+  protected void checkInterpolationArguments(Collection<T> formulasOfA) {
+    checkArgument(
+        getAssertedConstraintIds().containsAll(formulasOfA),
+        "Interpolation can only be done over previously asserted formulas. %s",
+        MoreStrings.lazyString(() -> getErrorString(formulasOfA)).toString());
```
PhilippWendler: Calling `toString()` forces the lazy string to be eagerly evaluated. I would simply use an `if` branch and `throw`. Not everything needs to use fancy features.

src/org/sosy_lab/java_smt/test/TimeoutTest.java
baierd: Fair. I renamed them.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: Good point. I made all private and added getters/setters where necessary.

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
baierd: Good point! Thanks for noticing. I changed it!

src/org/sosy_lab/java_smt/delegate/debugging/DebuggingBasicProverEnvironment.java
baierd: My guess is that this was missing as it calls `getModel()` usually. But it does not hurt, so i added it.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Good spot. I added the tag.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Thanks for the hint! Changed.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Thanks for the hint! I changed it!

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: No, they are meant for us internally. This was just a central point to access them. I moved them to be hidden from users, but now we have `SHUTDOWN_EXCEPTION_PREFIX` double (but with info to keep it in sync). I could not find a way to share this field without breaking changes, may you know a way?

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: Sorry. It's re-added.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: No, its `null` because i need a way of discerning if we have a usable prover based notifier or not, as some solvers do not support this. If there were a check `isDummy()` i could use a dummy.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
baierd: We have no way of accessing the notifiers here. And we can not get to them without making them public due to our architecture.

I wanted the error message to be equal to the message of the notifier, hence the modification of the String.

And yes that's a bug, good catch, i fixed it!

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
baierd: Because one would expect a `InterruptedException` when interrupting, not a `IllegalStateException`.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Thanks for the suggestion! I updated the JavaDoc of all 3 methods.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
PhilippWendler: I don't understand. The breakage is in deleting the other existing public fields. `SHUTDOWN_EXCEPTION_PREFIX` was not part of the code yet, how can moving it be breaking?

Also, I still question whether even using an exception-message string to convey information that needs to be machine usable is a good thing.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
PhilippWendler: > We have no way of accessing the notifiers here.

Why?

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
PhilippWendler: Yes, but why is there only this comment mentioning this instead of code that fixes it?

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
PhilippWendler: But why and where is this information actually needed?

src/org/sosy_lab/java_smt/api/SolverContext.java
PhilippWendler: I do not see a change here.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -65,17 +66,63 @@ enum ProverOptions {
...
   /**
    * Create a fresh new {@link InterpolatingProverEnvironment} which encapsulates an assertion stack
-   * and allows generating and retrieve interpolants for unsatisfiable formulas. If the SMT solver
-   * is able to handle satisfiability tests with assumptions please consider implementing the {@link
-   * InterpolatingProverEnvironment} interface, and return an Object of this type here.
+   * and allows generating and retrieve interpolants for unsatisfiable formulas.
    *
+   * @implNote If the SMT solver is able to handle satisfiability tests with assumptions please
+   *     consider implementing the {@link InterpolatingProverEnvironment} interface, and return an
+   *     Object of this type here.
```
PhilippWendler, line 94: I don't understand this comment. The return type of the method is `InterpolatingProverEnvironment`, one always needs to return such an instance. What does this have to do with "solving with assumptions"?

src/org/sosy_lab/java_smt/api/SolverContext.java
PhilippWendler: Several of my suggestions were not adopted, don't you agree with them?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
baierd: This is now outdated due to the changes.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: Since you can't set the privacy modifier to private or protected in interfaces, we would need to move it. But since this is needed in `BasicProverWithAssumptionsWrapper` and `AbstractProver`, all my changes lead to breaking changes. There might be a way, but i couldn't see it. My current solution hides these fields, but 1 has to be present twice.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
baierd: Because they are placed in the `AbstractProver`, and this wrapper is not inheriting from it. The delegate might be some other wrapper that also doesn't. 

We could of course try to access the nested delegates until we get to where it is accessible in all wrappers. (This would also work for the `SHUTDOWN_EXCEPTION_PREFIX` above).

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
PhilippWendler: I still see no explanation why adding a public constant somewhere (for example in `AbstractProver`) should be a breaking change.

But as mentioned, IMHO this shouldn't even be done. Please don't use strings and exception messages in order to convey information between classes. It is certainly a bad idea that one class relies on the specific format of the exception message produced by a different class. For example this breaks as soon as someone uses the `BasicProverWithAssumptionsWrapper` for an SMT solver that happens to not use the `AbstractProver` base class.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
PhilippWendler: > Because they are placed in the `AbstractProver`, and this wrapper is not inheriting from it. The delegate might be some other wrapper that also doesn't.

Why shouldn't `BasicProverWithAssumptionsWrapper` get the shutdown notifier? It can simply be passed in as constructor parameter.

> We could of course try to access the nested delegates until we get to where it is accessible in all wrappers.

I don't understand this sentence, sorry.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: This is a comment for developers (e.g. students) so that they get info what to implement. We have this at several locations and for example helped me back in my bachelors thesis.

baierd:
@kfriedberger @PhilippWendler i now merged #501 into this to get the Solver/InterruptedExceptions for the model and prover API. I also added them to `getUnsatCore()`. So now nearly every method in the prover throws both. This allows us to remove all that special handling for the exceptions (except `iterator()` ;D).

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: Good point! Thank you! Fixed.

PhilippWendler:
I just realized that ca14eff was even more important than I thought before.

If the intention of `checkShutdownState()` was "the method must not be called with shutdown already being triggered before", then this is not even possible. The following sequence can always happen:

1. Caller checks that no shutdown is triggered and decides to call a method like `getUnsatCore`.
2. Other thread triggers shutdown, e.g., due to timeout.
3. `checkShutdownState()` is called and throws.

This means that no matter what an application does, if it uses the shutdown interface as it is intended, it would randomly crash with `IllegalStateException`, which would be a bug.

So the only way to handle shutdown requests properly can be to throw `InterruptedException`, as it is done now I guess?

But note that it is unnecessary to check for shutdown immediately when entering a method. Because of the asynchronicity no method can and needs to guarantee that it always notices a shutdown. The only place where one needs to check for shutdowns is in loops that execute for potentially long time.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: I wouldn't say that. I incorporated your suggestion about the sentence, so i clearly agree there. 

For the changes to `@param` and/or `@throws` i would argue that it does not matter, the important thing is that the information is present. As a consequence i moved it as suggested now.

As for the info about the context and its notifier, i would argue that the info about the differences should be present in BOTH documentations, so that users are informed about how to do things in both situations. This prevents users from thinking they have to use one of the two, when the other exists.

src/org/sosy_lab/java_smt/api/SolverContext.java
baierd: Maybe i didn't push back then, sorry.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: I don't remember the reason why currently. I know that i tried it and found some problem with it.

With all the new changes this seems to work now though! Which obviously makes better code, and i think we can take the minimal overhead.

baierd:
> I just realized that [ca14eff](https://github.com/sosy-lab/java-smt/commit/ca14eff7af662bce4383b74ad0656f4742bf63b7) was even more important than I thought before.
> 
> If the intention of `checkShutdownState()` was "the method must not be called with shutdown already being triggered before", then this is not even possible. The following sequence can always happen:
> 
>     1. Caller checks that no shutdown is triggered and decides to call a method like `getUnsatCore`.
> 
>     2. Other thread triggers shutdown, e.g., due to timeout.
> 
>     3. `checkShutdownState()` is called and throws.
> 
> 
> This means that no matter what an application does, if it uses the shutdown interface as it is intended, it would randomly crash with `IllegalStateException`, which would be a bug.
> 
> So the only way to handle shutdown requests properly can be to throw `InterruptedException`, as it is done now I guess?
> 
> But note that it is unnecessary to check for shutdown immediately when entering a method. Because of the asynchronicity no method can and needs to guarantee that it always notices a shutdown. The only place where one needs to check for shutdowns is in loops that execute for potentially long time.

Oh wow! Good catch! This means that all the API changes are actually fully justified.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
baierd: This has been removed now and the correct exception is thrown.

src/org/sosy_lab/java_smt/basicimpl/withAssumptionsWrapper/BasicProverWithAssumptionsWrapper.java
baierd: This problem has been removed and the correct exception is thrown without workaround.

src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
baierd: The remaining problem has been removed, as the correct exception is thrown without workaround.

src/org/sosy_lab/java_smt/api/SolverContext.java
PhilippWendler: Sure, having `@implNote`s is good. But what does this one want to say? Isn't  `InterpolatingProverEnvironment` always required, and what does implementing it have to do with "satisfiability tests with assumptions"?

baierd: No it is not always required. The default impl throws a `UnsupportedOperationException`.
It is meant to lead the developer to the conclusion that this needs to be implemented if you want to interpolate. Usually students (in my experience) notice once they run the tests that things are missing ;D

PhilippWendler: Ok, but then the comment should still refer to interpolation and not satisfiability tests with assumptions.

PR #491, opened 2025-07-13
Cleanup usage of bound variables

kfriedberger:
The usage of bound variables and quantifier usage in the API was never consistent:
- some solvers require bound variables in quantified formulas, some solvers just use regular variables.
- some solvers visit bound variables in formulas, some solvers just handle them as regular variables.
- some solvers can not use an empty list of variables in quantification, Princess does allow this.
This PR unifies the beaviour and marks the corresponding methods as deprecated:
- all solvers provide only free variables toward the API/user
- all sub-formulas are consistent when visiting quantified formulas, e.g. the body of a quantified formula is provided in a bound-free way, e.g, all bound variables are replaced by free counterparts (for Princess, we use a new temporary unique name, other solvers keep the original name of the variable).

The overall goal is an analysis of the blocking error in https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/234 which might be caused by solver-specific and non-consistent quantifier handling (Z3 vs. other solver).

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
@@ -393,6 +393,7 @@ public class BitwuzlaFormulaCreator extends FormulaCreator<Term, Sort, Void, Bit
     return bitwuzlaSortToType(pType);
   }
 
+  @SuppressWarnings("deprecation")
   @Override
   public <R> R visit(FormulaVisitor<R> visitor, Formula formula, Term f)
       throws UnsupportedOperationException {
--- a/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
@@ -12,15 +12,12 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Table;
 import com.google.common.primitives.Longs;
 import java.math.BigInteger;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import org.sosy_lab.java_smt.api.ArrayFormula;
 import org.sosy_lab.java_smt.api.BitvectorFormula;
@@ -29,10 +26,8 @@ import org.sosy_lab.java_smt.api.Formula;
 import org.sosy_lab.java_smt.api.FormulaType;
 import org.sosy_lab.java_smt.api.FormulaType.ArrayFormulaType;
 import org.sosy_lab.java_smt.api.FormulaType.FloatingPointType;
-import org.sosy_lab.java_smt.api.FunctionDeclarationKind;
 import org.sosy_lab.java_smt.api.visitors.FormulaVisitor;
 import org.sosy_lab.java_smt.basicimpl.FormulaCreator;
-import org.sosy_lab.java_smt.basicimpl.FunctionDeclarationImpl;
 import org.sosy_lab.java_smt.solvers.boolector.BoolectorFormula.BoolectorArrayFormula;
 import org.sosy_lab.java_smt.solvers.boolector.BoolectorFormula.BoolectorBitvectorFormula;
 import org.sosy_lab.java_smt.solvers.boolector.BoolectorFormula.BoolectorBooleanFormula;
@@ -47,9 +42,6 @@ public class BoolectorFormulaCreator extends FormulaCreator<Long, Long, Long, Lo
    * per var name, meaning there is only 1 column per row!
    */
   private final Table<String, Long, Long> formulaCache = HashBasedTable.create();
-
-  // Remember uf sorts, as Boolector does not give them back correctly
-  private final Map<Long, List<Long>> ufArgumentsSortMap = new HashMap<>();
 
   // Possibly we need to split this up into vars, ufs, and arrays
 
@@ -203,50 +195,6 @@ public class BoolectorFormulaCreator extends FormulaCreator<Long, Long, Long, Lo
         "Boolector has no methods to access internal nodes for visitation.");
   }
 
-  // Hopefully a helpful template for when visitor gets implemented
-  // Btor only has bitvec arrays and ufs with bitvecs and arrays of bitvecs
-  // (and quantifier with bitvecs only)
-  @SuppressWarnings({"deprecation", "unused"})
-  private <R> R visit1(FormulaVisitor<R> visitor, Formula formula, Long f) {
-    if (BtorJNI.boolector_is_const(getEnv(), f)) {
-      // Handles all constants (bitvec, bool)
-      String bits = BtorJNI.boolector_get_bits(getEnv(), f);
-      return visitor.visitConstant(formula, convertValue(f, parseBitvector(bits)));
-    } else if (BtorJNI.boolector_is_param(getEnv(), f)) {
-      // Quantifier have their own variables called param.
-      // They can only be bound once! (use them as bitvec)
-      int deBruijnIdx = 0; // TODO: Ask Developers for this because this is WRONG!
-      return visitor.visitBoundVariable(formula, deBruijnIdx);
-    } else if (false) {
-      // Quantifier
-      // there is currently no way to find out if the formula is a quantifier
-      // do we need them separately?
-      /*
-       * return visitor .visitQuantifier( (BoolectorBooleanFormula) formula, quantifier,
-       * boundVariables, new BoolectorBooleanFormula(body, getEnv()));
-       */
-    } else if (BtorJNI.boolector_is_var(getEnv(), f)) {
-      // bitvec var (size 1 is bool!)
-      return visitor.visitFreeVariable(formula, getName(f));
-    } else {
-      ImmutableList.Builder<Formula> args = ImmutableList.builder();
-
-      ImmutableList.Builder<FormulaType<?>> argTypes = ImmutableList.builder();
-
-      return visitor.visitFunction(
-          formula,
-          args.build(),
-          FunctionDeclarationImpl.of(
-              getName(f), getDeclarationKind(f), argTypes.build(), getFormulaType(f), f));
-    } // TODO: fix declaration in visitFunction
-    return null;
-  }
-
-  // TODO: returns kind of formula (add, uf etc....) once methods are provided
-  private FunctionDeclarationKind getDeclarationKind(@SuppressWarnings("unused") long f) {
-    return null;
-  }
-
   @Override
   public Long callFunctionImpl(Long pDeclaration, List<Long> pArgs) {
     Preconditions.checkArgument(
@@ -270,7 +218,6 @@ public class BoolectorFormulaCreator extends FormulaCreator<Long, Long, Long, Lo
     }
     long uf = BtorJNI.boolector_uf(getEnv(), sort, name);
     formulaCache.put(name, sort, uf);
-    ufArgumentsSortMap.put(uf, pArgTypes);
     return uf;
   }
 
--- a/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
@@ -206,7 +206,7 @@ public class BoolectorFormulaCreator extends FormulaCreator<Long, Long, Long, Lo
   // Hopefully a helpful template for when visitor gets implemented
   // Btor only has bitvec arrays and ufs with bitvecs and arrays of bitvecs
   // (and quantifier with bitvecs only)
-  @SuppressWarnings("unused")
+  @SuppressWarnings({"deprecation", "unused"})
   private <R> R visit1(FormulaVisitor<R> visitor, Formula formula, Long f) {
     if (BtorJNI.boolector_is_const(getEnv(), f)) {
       // Handles all constants (bitvec, bool)
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4QuantifiedFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4QuantifiedFormulaManager.java
@@ -8,6 +8,8 @@
 
 package org.sosy_lab.java_smt.solvers.cvc4;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import edu.stanford.CVC4.Expr;
 import edu.stanford.CVC4.ExprManager;
 import edu.stanford.CVC4.Kind;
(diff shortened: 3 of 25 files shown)


src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
```diff
@@ -184,6 +184,11 @@ class PrincessFormulaCreator
    */
   private final Table<Sort, Sort, Sort> arraySortCache = HashBasedTable.create();
 
+  /** This map keeps track of the bound/free variables created when visiting quantified terms. */
+  // TODO should we use a WeakHashMap?
+  //  We do not cleanup in other places, too, and the number of quantified formulas is small.
+  private final Map<IFormula, String> boundVariableNames = new LinkedHashMap<>();
```
PhilippWendler, line 190: Do you really want to introduce a leak here? I know that for native solvers memory management is difficult and for some solvers all created formulas are stored forever. But for a pure Java solver I would (as a user of JavaSMT) certainly expect that GC works as usual.

I am also surprised how you can know that the number of quantified formulas is low. Doesn't this depend on what JavaSMT is used for?

I would also wonder whether such a global cache works correctly across all visitors? For example, what happens if there are multiple ongoing visitations, especially of the same formula?
And doesn't this change what formulas I get even if I just change what read operations the program performs? For example, if I create two (quantified) formulas `f` and `g`, then visit both of them and print the `body`, I would guess it is desired that the result doesn't depend on whether `f` or `g` are visited first. Otherwise it could be inconvenient when one is for example debugging a problem and introducing additional debugging statements (with printfs and maybe some visitors to find out facts about a formula), but then the problematic formulas change.

PhilippWendler (commented):
>  * all sub-formulas are consistent when visiting quantified formulas, e.g. the body of a quantified formula is provided in a bound-free way, e.g, all bound variables are replaced by free counterparts

Isn't this costly (or could be potentially costly at least)? It seems like it introduces a O(n) operation (with n being the DAG size of the formula) for an operation (visiting top-level node of formula) that one intuitively assumes to be O(1). I guess there are many cases of visitors that are used to only find out some cheap facts about the formula, but now these suddenly become expensive when passed a quantified formula. Even JavaSMT probably has plenty of these cases, such as for example `conjunctionFinder` and `disjunctionFinder`. IIRC using a visitor is in some (many?) cases the only way to get some information about a particular formula in JavaSMT, which would of course be suboptimal if it is not cheap even in cases where one does not need the expensive information.

> The overall goal is an analysis of the blocking error in [gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/234](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/234) which might be caused by solver-specific and non-consistent quantifier handling (Z3 vs. other solver).

What blocking error is there? If there is such an error, it would be good if it is explained and discussed in that MR, but I don't see any related unresolved thread.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4QuantifiedFormulaManager.java
```diff
@@ -72,23 +74,22 @@ protected Expr eliminateQuantifiers(Expr pExtractInfo)
...
-      Expr substBody = pBody;
-      // every free needs a bound copy. As the internal Id is different for every variable, even
-      // with the same name, this is fine.
-      for (Expr var : pVars) {
-        Expr boundCopy = ((CVC4FormulaCreator) formulaCreator).makeBoundCopy(var);
-        vec.add(boundCopy);
-        substBody = substBody.substitute(var, boundCopy);
-      }
-      Expr quantifiedVars = exprManager.mkExpr(Kind.BOUND_VAR_LIST, vec);
+    checkArgument(
```
baierd, line 77: This should be unnecessary now with the changes in `AbstractQuantifiedFormulaManager`.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5QuantifiedFormulaManager.java
```diff
@@ -63,22 +65,21 @@ protected Term eliminateQuantifiers(Term input) throws SolverException, Interrup
...
-      List<Term> boundVars = new ArrayList<>();
-      Term substBody = pBody;
-      // every free needs a bound copy. As the internal Id is different for every variable, even
-      // with the same name, this is fine.
-      for (Term var : pVars) {
-        Term boundCopy = ((CVC5FormulaCreator) formulaCreator).makeBoundCopy(var);
-        boundVars.add(boundCopy);
-        substBody = substBody.substitute(var, boundCopy);
-      }
+    checkArgument(
```
baierd, line 68: This should be unnecessary now, as it is handled in the `AbstractQuantifiedFormulaManager`.

src/org/sosy_lab/java_smt/solvers/princess/PrincessQuantifiedFormulaManager.java
```diff
@@ -40,24 +39,17 @@ class PrincessQuantifiedFormulaManager
 
   @Override
   public IExpression mkQuantifier(Quantifier q, List<IExpression> vars, IExpression body) {
+    checkArgument(!vars.isEmpty(), "Missing variables for quantifier '%s' and body '%s'.", q, body);
```
baierd, line 42: This should be unnecessary now, as it is handled in the `AbstractQuantifiedFormulaManager`.

src/org/sosy_lab/java_smt/solvers/z3/Z3QuantifiedFormulaManager.java
```diff
@@ -29,7 +29,12 @@ class Z3QuantifiedFormulaManager extends AbstractQuantifiedFormulaManager<Long,
 
   @Override
   public Long mkQuantifier(Quantifier q, List<Long> pVariables, Long pBody) {
-    checkArgument(!pVariables.isEmpty(), "List of quantified variables can not be empty");
+    checkArgument(
```
baierd, line 32: This should be unnecessary now, as it is handled in the `AbstractQuantifiedFormulaManager`.

baierd (changes requested):
Very good thought to clean this up! I like it!
I don't see any problems in the code besides the small redundancy. Is this a breaking change? People relying on the old implementation can not use it fully (even if ignoring the deprecation) as far as i can see.

baierd:
> > * all sub-formulas are consistent when visiting quantified formulas, e.g. the body of a quantified formula is provided in a bound-free way, e.g, all bound variables are replaced by free counterparts
> 
> Isn't this costly (or could be potentially costly at least)? It seems like it introduces a O(n) operation (with n being the DAG size of the formula) for an operation (visiting top-level node of formula) that one intuitively assumes to be O(1). I guess there are many cases of visitors that are used to only find out some cheap facts about the formula, but now these suddenly become expensive when passed a quantified formula. Even JavaSMT probably has plenty of these cases, such as for example `conjunctionFinder` and `disjunctionFinder`. IIRC using a visitor is in some (many?) cases the only way to get some information about a particular formula in JavaSMT, which would of course be suboptimal if it is not cheap even in cases where one does not need the expensive information.

Short answer: yes. But since the solvers are not consistent when it comes to bound/free variables, we handle it this way, as it is easier for users when creating quantified formulas.

Also, since we are usually dealing with native objects (besides the Java/Scala solvers) we generally need to request information about terms from the solver, which we only want to do when necessary. We could save more information about formulas directly in the formula-wrappers when creating them, but this would increase overhead.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
```diff
@@ -393,6 +393,7 @@ public FormulaType<?> getFormulaType(Term formula) {
     return bitwuzlaSortToType(pType);
   }
 
+  @SuppressWarnings("deprecation")
```
daniel-raffler, line 396: Can't we just update `visit` for all the solvers to no longer use `visitBoundVariable`? As far as I can see `Yices2` might be the only solver where we don't substitute bound variables with free variables. All other cases for bound variables would have been dead code anyway.

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
```diff
@@ -538,6 +532,43 @@ public <R> R visit(FormulaVisitor<R> visitor, final Formula f, final IExpression
...
   }
 
+  private <R> R visitQuantifier(FormulaVisitor<R> visitor, BooleanFormula f, IQuantified input) {
+    Quantifier quantifier =
+        input.quan().equals(ap.terfor.conjunctions.Quantifier.apply(true)) ? FORALL : EXISTS;
+    IFormula body = input.subformula();
+
+    // Princess uses de-Bruijn indices, so we have index 0 here for the most outer quantified scope
+    IVariable boundVariable = input.sort().boundVariable(0);
+    String boundVariableName = getFreshVariableNameForBody(body);
```
daniel-raffler, line 542: We could also  store the old variable names when a quantified formula is created. That way we don't have to create new variable names for the visitor. This might be more consistent with what the other solvers are doing, but it would probably add some overhead.

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
```diff
@@ -206,7 +206,7 @@ public <R> R visit(FormulaVisitor<R> visitor, Formula formula, Long f) {
   // Hopefully a helpful template for when visitor gets implemented
   // Btor only has bitvec arrays and ufs with bitvecs and arrays of bitvecs
   // (and quantifier with bitvecs only)
-  @SuppressWarnings("unused")
+  @SuppressWarnings({"deprecation", "unused"})
```
daniel-raffler: I think this method could just be removed entirely. `Boolector` is no longer being developed and we're not going to add visitor support to it.

daniel-raffler (dismissed):
Seems like a good idea overall. We should probably check to make sure that nobody was using the old interface, even if was broken for most solvers

PhilippWendler:
> > > * all sub-formulas are consistent when visiting quantified formulas, e.g. the body of a quantified formula is provided in a bound-free way, e.g, all bound variables are replaced by free counterparts
> > 
> > Isn't this costly (or could be potentially costly at least)? It seems like it introduces a O(n) operation (with n being the DAG size of the formula) for an operation (visiting top-level node of formula) that one intuitively assumes to be O(1). I guess there are many cases of visitors that are used to only find out some cheap facts about the formula, but now these suddenly become expensive when passed a quantified formula. Even JavaSMT probably has plenty of these cases, such as for example `conjunctionFinder` and `disjunctionFinder`. IIRC using a visitor is in some (many?) cases the only way to get some information about a particular formula in JavaSMT, which would of course be suboptimal if it is not cheap even in cases where one does not need the expensive information.
> 
> Short answer: yes. But since the solvers are not consistent when it comes to bound/free variables, we handle it this way, as it is easier for users when creating quantified formulas.

It is clear that one cannot have quantifier handling fast and easy at the same time in this case. But is it really necessary that everyone pays the performance price whenever they use a visitor on a formula that happens to be quantified, even if they do not need it anyway? I see as the great advantage of JavaSMT that it allows one to use a lot of features of SMT solvers in a solver-independent manner without a disadvantage. This way I don't have to worry about the question whether I should use JavaSMT or access the solver directly. But if I have to pay a price when using JavaSMT (e.g., in reduced performance compared to direct access) then the question of using JavaSMT becomes a trade off and is no longer clear cut. So I would recommend to invest more effort here in a solution.

> Also, since we are usually dealing with native objects (besides the Java/Scala solvers) we generally need to request information about terms from the solver, which we only want to do when necessary. We could save more information about formulas directly in the formula-wrappers when creating them, but this would increase overhead.

Not sure how this relates to the current discussion? Is there a trade off here between storing more information and the performance?

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
kfriedberger: Thank you for your feedback on my explicitly noted "TODO" in the code. :-)

On memory leaks, I agree that Java users expect standard GC behavior. The current cache assumes manageable formula counts. Regarding quantified formulas, you’re right that their number depends on usage. Our “low” assumption is based on typical cases, but we’re open to adaptive caching strategies. Any specific workloads where this fails, e.g., in CPAchecker?

For visitor determinism, we aim to ensure consistent results regardless of visitation order. We could use a deterministic naming scheme (e.g., based on formula structure or formula-hash) to avoid issues with randomly ordered visitations or debugging. For now, I would like to stay with the minimal working solution.

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4QuantifiedFormulaManager.java
kfriedberger: Better check twice. And there were internal calls to such methods. I do not expect that check to harm anyone.

src/org/sosy_lab/java_smt/solvers/princess/PrincessQuantifiedFormulaManager.java
kfriedberger: same as above: better check twice.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5QuantifiedFormulaManager.java
kfriedberger: same as above: better check twice.

src/org/sosy_lab/java_smt/solvers/z3/Z3QuantifiedFormulaManager.java
kfriedberger: same as above: better check twice.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFormulaCreator.java
kfriedberger: Cleanup can be done later.
I wanted to get a working and solver-independent consistent visitor implementation for quantifiers.

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
kfriedberger: Good point. I thought so as well.
However, then I found some problematic use cases:

- we also support parsing SMT-LIB directly, where we do not directly create quantified formulas.
- the solver can return quantified formulas in interpolants or unsat cores, or other formula-producing methods. 

We can not track bound variables there and need to find a more general solution.

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorFormulaCreator.java
kfriedberger: Cleanup can be done later. But thanks for the hint.

kfriedberger:
You’re right that the current approach introduces an O(n) cost for visiting quantified formulas in Princess, which can be unexpected for operations assumed to be O(1), such as accessing data directly. Our implementation currently prioritizes consistency of use for creating and visiting quantified formulas over performance in some cases.

However, there are even more operations in Princess that are not in constant time, even if expected. E.g., Princess applies a shift to all bound variables from nested quantifications when a new free variable is made bound by quantification. Building a plain `EXISTS(x, body)` requires O(n) where n is formula size.

If the user shows with some more details (like measurements or benchmarks) that the visiting of quantified formulas is an issue), then we can report this to the Princess developers and try to find a better solution. For now, I prefer a consistent behavious (not performance) of the JavaSMT API.

kfriedberger:
> Very good thought to clean this up! I like it! I don't see any problems in the code besides the small redundancy. Is this a breaking change? People relying on the old implementation can not use it fully (even if ignoring the deprecation) as far as i can see.

Actually yes, this change breaks exactly those cases where bound variables were visited before. That happend to be Z3 and Yices2 before commit e700fb71bd48d01bee7347aee4b92da616297bd0, e.g., we already broke the behaviour at that old commit.

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
PhilippWendler: > On memory leaks, I agree that Java users expect standard GC behavior. The current cache assumes manageable formula counts. Regarding quantified formulas, you’re right that their number depends on usage. Our “low” assumption is based on typical cases, but we’re open to adaptive caching strategies. Any specific workloads where this fails, e.g., in CPAchecker?

I personally do not use Princess, so I will not be affected. But do you really want to base the development of JavaSMT on the assumption that it will not be used for large amounts of formulas? As I wrote already, I see the big selling point of JavaSMT in being something that I can adopt without having to ponder about disadvantages, because in almost every case there are no noticeable disadvantages. If you publicly write that JavaSMT is not intended for large amounts of formulas, this could convince me of the opposite.

> For visitor determinism, we aim to ensure consistent results regardless of visitation order. We could use a deterministic naming scheme (e.g., based on formula structure or formula-hash) to avoid issues with randomly ordered visitations or debugging. For now, I would like to stay with the minimal working solution.

What does "for now" here mean, i.e., when would this change? I don't really understand what the difference will be between now and some point in the future.

For me, nondeterministic behavior has the potential to easily cause hours of additional debugging effort even if just one case happens in the future and it triggers. Plus, it is then usually unclear where the nondeterministic behavior comes from, and finding out its cause could easily take more hours, so delaying the implementation will just be more costly.

PhilippWendler:
> You’re right that the current approach introduces an O(n) cost for visiting quantified formulas in Princess, which can be unexpected for operations assumed to be O(1), such as accessing data directly. Our implementation currently prioritizes consistency of use for creating and visiting quantified formulas over performance in some cases.

Is consistency of use really contrary to performance here? I don't really see that. After all, the solver does provide introspection methods that are cheap, it is just JavaSMT's handling that imposes O(n) cost. And it is not obvious that it needs to couple the creation of a usable formula for the body of a quantifier to the only method that allows me to determine whether the formula I have is a disjunction or conjunction.

> However, there are even more operations in Princess that are not in constant time, even if expected. E.g., Princess applies a shift to all bound variables from nested quantifications when a new free variable is made bound by quantification. Building a plain `EXISTS(x, body)` requires O(n) where n is formula size.

The difference is that in your example (a) only the actual operation is more costly than some would assume, (b) it is a creation operation, and (c) if the solver does not provide any other method it is not JavaSMT's problem. With the current MR, (a) other *unrelated* methods in the JavaSMT API (such as `toConjunctionArgs` and all client visitors) become potentially O(n), even though the "costly" quantifier is not relevant here, (b) these operations include trivial "read" operations that do not create formulas (so non-O(1) is even less expected), and (c) the solver is not the reason.

PR #499, opened 2025-07-27
update CVC5 to latest version from 2025-07-24

kfriedberger:


kfriedberger:
There are the following regressions:

- ALLSAT computation in `SolverAllSatTest::allSatTest_xor` crashes with an invalid model. The constraint `(i=1) XOR (i=2)` seems to produce a model with `i=0`.
- SAT check in `ModelTest::testDeeplyNestedFormulaLIA` and in`ModelTest::testDeeplyNestedFormulaBV` use a predicate `basis` in the formula, and query its value from the model. It should be `true`, but CVC5 reports `false`.

Both cases appear with model evaluation. Maybe a starting point for analysis and debugging?

@daniel-raffler : Could you take a look at the regressions in CVC5. Maybe a starting point for analysis and debugging? Thanks.

baierd:
Thank you very much for the quick work @kfriedberger !

This might be related to the changes of the proof granularity (see changelog [here](https://github.com/cvc5/cvc5/releases/tag/cvc5-1.3.0)). We should try running/compiling CVC5 with the previous default. It might be sufficient to export the problematic cases into SMT2 and try to execute them on the binaries with distinct settings.

daniel-raffler:
> This might be related to the changes of the proof granularity (see changelog [here](https://github.com/cvc5/cvc5/releases/tag/cvc5-1.3.0)). We should try running/compiling CVC5 with the previous default. It might be sufficient to export the problematic cases into SMT2 and try to execute them on the binaries with distinct settings.

Thanks for the hint! I'll try to debug the issue later today

daniel-raffler:
I've opened a new CVC5 issues here:
https://github.com/cvc5/cvc5/issues/12038

daniel-raffler:
It's been fixed in version `2025-07-29-6dacfa5`

@kfriedberger
Could you update the binaries once more?

kfriedberger:
Thanks for reporting to CVC5. Problem is fixed.

PR #495, opened 2025-07-21
Draft: Prepare Release 6.0.0

baierd:
To prepare the release of 6.0.0 we should discuss all changes that we included and want to include, as well as update our README and info material to reflect the current state of JavaSMT.

TODO:
- [ ] All breaking changes that we plan to do should be included in this PR. For example #491.
- [ ] We need to discuss which changes currently under development are supposed to be finished for this release.
- [ ] We need to update our info material to reflect the current version of JavaSMT. We should be clear about future developments (e.g. we want to update to Java 17 soon and 21 next year), as well as the current requirements and dependencies (i.e. what is needed to run or build our solvers).
- [ ] We need to write a thorough release message with the many changes that piled up, as well as our plans for the future regarding Java versions etc.

PR #508, opened 2025-08-28
Allow users to build new rounding mode formulas

daniel-raffler:
Hello,
in this PR we add support to explicitly create new rounding mode formulas. So far this has not been possible and rounding-mode formulas were added internally when building new floating-point formulas. This leads to some issues when using visitors on floating-point formulas, as the rounding mode can not be changed when rebuilding the same formula. Let's take the term `(fp.sqrt roundTowardPositive a)` as an example. Here the rounding mode `roundTowardPositive` is represented as a `FloatingPointRoundingModeFormula` in JavaSMT and can in fact be visited. However, since there is no way to instantiate `FloatingPointRoundingModeFormula`, all the visitor can do is return the same rounding mode again. In the same way there is no way to access the `FloatingPointRoundingModeFormula` to see which rounding mode is being used by the term.

In this PR we add a new method `makeRoundingMode(FloatingPointRoundingMode pRoundingMode)` to the `FloatingPointFormulaManager` that allows user to create new rounding mode formulas. We also fixed the visitor code in the backends to properly map solver rounding mode values to JavaSMT rounding modes:
```
mgr.transformRecursively(t, new FormulaVisitor<Formula>() {
  @Override
  public Formula visitConstant(Formula f, Object value) {
    if(f instanceof FloatingPointRoundingModeFormula && value instanceof FloatingPointRoundingMode) {
      // Switch the rounding mode to "neareast-ties-to-even"
      return fpmgr.makeRoundingMode(FloatingPointRoundingMode.NEAREST_TIES_TO_EVEN);
    } else {
      return f;
    }
  }
}
```

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5FormulaCreator.java
@@ -127,6 +127,8 @@ import org.sosy_lab.java_smt.api.EnumerationFormula;
 import org.sosy_lab.java_smt.api.FloatingPointFormula;
 import org.sosy_lab.java_smt.api.FloatingPointNumber;
 import org.sosy_lab.java_smt.api.FloatingPointNumber.Sign;
+import org.sosy_lab.java_smt.api.FloatingPointRoundingMode;
+import org.sosy_lab.java_smt.api.FloatingPointRoundingModeFormula;
 import org.sosy_lab.java_smt.api.Formula;
 import org.sosy_lab.java_smt.api.FormulaType;
 import org.sosy_lab.java_smt.api.FormulaType.ArrayFormulaType;
@@ -289,6 +291,12 @@ class Mathsat5FormulaCreator extends FormulaCreator<Long, Long, Long, Long> {
   protected FloatingPointFormula encapsulateFloatingPoint(Long pTerm) {
     assert getFormulaType(pTerm).isFloatingPointType();
     return new Mathsat5FloatingPointFormula(pTerm);
+  }
+
+  @Override
+  protected FloatingPointRoundingModeFormula encapsulateRoundingMode(Long pTerm) {
+    assert getFormulaType(pTerm).isFloatingPointRoundingModeType();
+    return new Mathsat5FloatingPointRoundingModeFormula(pTerm);
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


daniel-raffler:
Thanks! I'll add some test later today

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5FormulaCreator.java
```diff
@@ -333,6 +340,20 @@ public <R> R visit(FormulaVisitor<R> visitor, Formula formula, final Long f) {
       return visitor.visitConstant(formula, true);
     } else if (msat_term_is_false(environment, f)) {
       return visitor.visitConstant(formula, false);
+    } else if (msat_is_fp_roundingmode_type(environment, msat_term_get_type(f))) {
+      long decl = msat_term_get_decl(f);
+      switch (msat_decl_get_name(decl)) {
```
baierd: You could also use the functions MathSAT5 provides to check the exact type of FP-rounding-mode ;D

daniel-raffler: Thanks for the tip! I've updated the code now

daniel-raffler:
Should we add a function to get the rounding mode from a rounding mode formula? We currently have to do something like this, which is rather awkward:
```
private FloatingPointRoundingMode getRoundingMode(
    FloatingPointRoundingModeFormula pRoundingModeFormula) {
  return visit(
      pRoundingModeFormula,
      new DefaultFormulaVisitor<>() {
        @Override
        protected FloatingPointRoundingMode visitDefault(Formula f) {
          throw new IllegalStateException();
        }

        @Override
        public FloatingPointRoundingMode visitConstant(Formula f, Object value) {
          return (FloatingPointRoundingMode) value;
        }
      });
}
```

Alternatively we could change the API in `FloatingPointFormulaManager` to require rounding mode **formulas**, instead of rounding modes as arguments. So, for instance:
```
FloatingPointFormula sqrt(FloatingPointFormula number, FloatingPointRoundingMode roundingMode);
```
would become:
```
FloatingPointFormula sqrt(FloatingPointFormula number, FloatingPointRoundingModeFormula roundingMode);
```
The original API could then be restored with a default implementation:
```
default FloatingPointFormula sqrt(FloatingPointFormula number, FloatingPointRoundingMode rm) {
  return sqrt(number, makeRoundingMode(rm));
}
```

daniel-raffler:
Another way to implement this is to use indices for the rounding mode. So `(fp.sqrt rm f)` becomes `((_ fp.sqrt rm) f)` and the round mode argument is tucked away from the user. This may be more intuitive as the operations now have the right number of arguments (1 for `sqrt`, 2 for `add`, etc) and all arguments are floating point formulas.

The downside is that indices have to be constant, so something like `((_ fp.sqrt (ite c roundTowardNegative roundTowardPositive) f)` would not work. On the other hand such cases are rare, and it should always be possible to lift the operation: ```(ite c ((_ fp.sqrt  roundTowardNegative) f) ((_ fp.sqrt  roundTowardPositive) f)```

kfriedberger:
@daniel-raffler I added the reversed function for retrieval of RoundingMode from a RoundingModeFormula into FPManager. This is a small and backwards-compatible API change.

kfriedberger (approved):

baierd:
This is a quite important and big change. @daniel-raffler would you be so kind and add a draft of the changes to https://github.com/sosy-lab/java-smt/pull/495?

PR #509, opened 2025-08-28
Add more FunctionDeclarationKinds for the visitor

daniel-raffler:
Hello,
this PR fixes some minor issues:
- In the MathSAT backend we now properly recognize terms for division. These used to be mapped to `FunctionDeclarationKind.OTHER` when they should be `FunctionDeclarationKind.DIV`
- We added support for `int_to_bv`, `ubv_to_int`  and `sbv_to_int`. These are need to convert between bitvector and integer terms, and while were able to create such terms with `BitvectorFormulaManager.makeBitvector(int length, IntegerFormula pI)` and `BitvectorFormulaManager.toIntegerFormula)` they were not properly supported by the visitor

**TODO**
- Wait for https://github.com/sosy-lab/java-smt/pull/507 as `int_to_bv` is an indexed function and will need to be supported in the visitor

### DIFF
--- a/src/org/sosy_lab/java_smt/test/SolverVisitorTest.java
+++ b/src/org/sosy_lab/java_smt/test/SolverVisitorTest.java
@@ -311,6 +311,57 @@ public class SolverVisitorTest extends SolverBasedTest0.ParameterizedSolverBased
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


kfriedberger (approved):
lgtm

PR #512, opened 2025-09-01
Add IEEE-754 Floating Point to Bitvector Conversion Fallback

### LINKED ISSUE #504: IEEE Floating-point to Bitvector Conversion Not Supported in All Solvers
IEEE Floating-point to Bitvector conversion is not supported in Bitwuzla and CVC5, possibly others. The standard says two things about this conversion:

```
:note
 "There is no function for converting from (_ FloatingPoint eb sb) to the
  corresponding IEEE 754-2008 binary format, as a bit vector (_ BitVec m) with 
  m = eb + sb, because (_ NaN eb sb) has multiple, well-defined representations.
  Instead, an encoding of the kind below is recommended, where f is a term
  of sort (_ FloatingPoint eb sb):

   (declare-fun b () (_ BitVec m))
   (assert (= ((_ to_fp eb sb) b) f))
 "
```

So it is nice that some solvers give us this method, but in general we can not expect them to add it (for Bitwuzla we asked and got told that it will not be added).

And there is an implementation that we can follow, but it has 2 problems:
- A constraint is created and needs to be added to provers that use this variable (and as far as i can see ONLY those). So we would need to add a constraint without the users knowledge.
- A new BV variable is created that the user also does not know about.

We could of course add a method that returns this info and the constraint as well as the BV variable.

This came up in CPAchecker [here](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/299).

Also, we already have an implementation in Bitwuzla.

In my opinion, this looks like something a user should add instead of us, but i am completely open to suggestions/ideas.

baierd:
This PR adds a fallback method for `FloatingPointManager.toIeeeBitvector(FloatingPointFormula)`:

- `FloatingPointManager.toIeeeBitvector(FloatingPointFormula, BitvectorFormula)`, returns a BooleanFormula that is the result of the equality of the FP input with the BV input (BV is equal to the floating-points bit representation according to the IEEE 754-2008 floating-point format). Special FP numbers are returned as the solver sees fit.

Tests for this method have been added. This includes the SMTLIB2 output of it, and the native `toIeeeBitvector()` method, showing us what solver can parse what output.
 
Furthermore, i added common FP type checks in the wrapping function of `AbstractFloatingPointFormulaManager`


baierd:
@kfriedberger the SMTLIB2 standard defines the size of an FP as mantissa + exponent with the mantissa including the sign bit. Our current implementation excludes it, and we do not document this properly in all cases, especially the FP type constructor.

This is problematic, as the standard defines the size of a BV representation of a FP as `mantissa + exponent`, which is off by 1 in our returned values (this is how i noticed btw.).

My proposal (already WIP):
- Fix FP type etc. to include the sign bit, and switch to what the standard defines in our internal implementation. Only MathSAT5 expects the mantissa to not include the sign bit.
- Announce the FP type constructor to be deprecated in the near future due to this.
- Add new constructors for FP type that specify the inclusion in the method name.
- Add methods to get the sizes of mantissa and exponent to `FloatingPointFormulaManager` based on the solvers terms instead of our types (similar to what we do for BV width), allowing assertions and checking the solvers interpretation for our type.

I don't want to change the current FP-Type constructor, as users expect this to not change behavior.

What do you think? I am happy about thoughts/ideas.

Edit: it would probably be a good idea to also update the API for `getMantissaSize()` to `getMantissaSizeWithSignBit()` and `getMantissaSizeWithoutSignBit()` and deprecate the old method.

baierd:
Do we want to document the typical returns of solvers for special FP numbers? There are multiple well defined return values for NaN as bitvector for example.

baierd:
> @kfriedberger the SMTLIB2 standard defines the size of an FP as mantissa + exponent with the mantissa including the sign bit. Our current implementation excludes it, and we do not document this properly in all cases, especially the FP type constructor.
> 
> This is problematic, as the standard defines the size of a BV representation of a FP as `mantissa + exponent`, which is off by 1 in our returned values (this is how i noticed btw.).
> 
> My proposal (already WIP):
> 
>     * Fix FP type etc. to include the sign bit, and switch to what the standard defines in our internal implementation. Only MathSAT5 expects the mantissa to not include the sign bit.
> 
>     * Announce the FP type constructor to be deprecated in the near future due to this.
> 
>     * Add new constructors for FP type that specify the inclusion in the method name.
> 
>     * Add methods to get the sizes of mantissa and exponent to `FloatingPointFormulaManager` based on the solvers terms instead of our types (similar to what we do for BV width), allowing assertions and checking the solvers interpretation for our type.
> 
> 
> I don't want to change the current FP-Type constructor, as users expect this to not change behavior.
> 
> What do you think? I am happy about thoughts/ideas.
> 
> Edit: it would probably be a good idea to also update the API for `getMantissaSize()` to `getMantissaSizeWithSignBit()` and `getMantissaSizeWithoutSignBit()` and deprecate the old method.

I mostly finished this as proposed. You can take a look. I feel like the API changes reduce the magic `+ 1` and `- 1` by a large amount, increasing readability of the code. My guess is that users, especially new users, will benefit from that. The deprecation warning needs to be discussed/done.

baierd:
I now split all changes in regard to FP mantissa into a dedicated [PR here](https://github.com/sosy-lab/java-smt/pull/513). We should first handle that PR, merge it, and then continue here.

src/org/sosy_lab/java_smt/api/FormulaType.java
```diff
@@ -205,14 +205,61 @@ public String toSMTLIBString() {
...
+   * The mantissa size is expected to not include the sign bit.
+   *
+   * @param exponentSize size of the exponent for the base of the floating-point
+   * @param mantissaSizeWithSignBit size of the mantissa (also called a coefficient or significant),
+   *     including the sign bit.
+   * @return the newly constructed {@link FloatingPointType}.
+   */
+  public static FloatingPointType getFloatingPointTypeWithSignBit(
+      int exponentSize, int mantissaSizeWithSignBit) {
+    return new FloatingPointType(exponentSize, mantissaSizeWithSignBit);
```
kfriedberger: Where is the difference in this method to the one before? Both methods create the same object with the same parameters, even if they are named differently.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -80,8 +81,29 @@ public final boolean getSign() {
 
   public abstract int getExponentSize();
 
+  /**
+   * Returns the size of the mantissa (also called a coefficient or significant), excluding the sign
+   * bit.
+   */
+  // TODO: mark as soon to be deprecated
```
kfriedberger: Do we really want to remove this simple method? Is the documentation not sufficient?

Please provide a replacement code snippet for inlining deprecated code.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -256,7 +270,95 @@ public BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber) {
...
-  protected abstract TFormulaInfo toIeeeBitvectorImpl(TFormulaInfo pNumber);
+  @SuppressWarnings("unused")
+  protected TFormulaInfo toIeeeBitvectorImpl(TFormulaInfo pNumber) {
+    throw new UnsupportedOperationException(
+        "The chosen solver does not support transforming "
+            + "FloatingPointFormula to IEEE bitvectors. Try using the fallback "
+            + "methods toIeeeBitvector"
+            + "(FloatingPointFormula, String) and/or "
+            + "toIeeeBitvector(FloatingPointFormula, String, Map)");
+  }
```
kfriedberger, line 322: This message is quite misleading: The message will be seen by a user of the application, not by the developer calling JavaSMT. How will the user of the application (e.g., CPAchecker) change that call? That might be impossible.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -519,4 +621,28 @@ protected static String getBvRepresentation(BigInteger integer, int size) {
     }
     return String.copyValueOf(values);
   }
+
+  public static final class BitvectorFormulaAndBooleanFormula {
```
kfriedberger: Why not just use a Google-Auto-class? In the future this might be a record.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -256,7 +270,95 @@ public BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber) {
...
+  }
+
+  @Override
+  public BitvectorFormulaAndBooleanFormula toIeeeBitvector(
+      FloatingPointFormula f, String bitvectorConstantName) {
+    return toIeeeBitvector(f, bitvectorConstantName, ImmutableMap.of());
+  }
+
+  @Override
+  public BitvectorFormulaAndBooleanFormula toIeeeBitvector(
```
kfriedberger: This method is weird. Does is really handle NaN as expected?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -467,4 +529,13 @@ FloatingPointFormula multiply(
...
   BooleanFormula isNegative(FloatingPointFormula number);
+
+  /**
+   * Returns the size of the mantissa (also called a coefficient or significant), including the sign
+   * bit, for the given {@link FloatingPointFormula}.
+   */
+  int getMantissaSizeWithSignBit(FloatingPointFormula number);
+
+  /** Returns the size of the exponent for the given {@link FloatingPointFormula}. */
+  int getExponentSize(FloatingPointFormula number);
```
kfriedberger: Do we really need these two methods?
The user already can access the FormulaType directly and extract exponentSize and mantissaSize from the returned FP-FormulaType.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
```diff
@@ -148,6 +147,16 @@ protected Term castToImpl(
...
+  @Override
+  protected int getMantissaSizeWithSignBitImpl(Term fp) {
+    return fp.sort().fp_sig_size();
+  }
+
+  @Override
+  protected int getExponentSizeImpl(Term fp) {
+    return fp.sort().fp_exp_size();
+  }
+
```
kfriedberger: I doupt the requirement for these two methods.

kfriedberger (commented):
I am unsure whether this special handling for FP2IEEEBV is really required. This feature looks errorprone and quite complex to understand. I would prefer that solvers provide their internal implementation for this, and JavaSMT just provides the wrapper/bindings instead of such parts. There might be potential, but also the burden for the developer, especially when using a method that returns additional constraints.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
baierd: You are correct. We already expose this through the type.

We could still keep them internally as sanity checks/assertions though. Would that be OK for you?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: See my answer below.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: This is clearly a record once we update.

I didn't think about auto-classes while writing. But i would prefer to switch to records once we update. (Which is 100% personal preference. I don't mind if we go with auto-classes.)

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
kfriedberger: you can have as many private methods as you want.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
kfriedberger: see my answer below, too :smile:

baierd:
Note: [as written above](https://github.com/sosy-lab/java-smt/pull/512#issuecomment-3244350919) we extracted many changes into PR #513. Hence this branch is now based on #513 and it makes sense to first the other PR. Everyone feel free to provide feedback for the API/ideas with the fallback for IEEE FP to BV at any time though!

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
PhilippWendler: Note that it is not possible to add internal methods to the `FloatingPointFormulaManger` interface.

I would also recommend to remove them, or at most add them as `protected abstract` to `AbstractFloatingPointormulaManager`. But if you force solver packages to implement methods for retrieving sizes more than once, the risk for errors and inconsistencies also rises.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -13,9 +13,11 @@
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Map;
 import org.sosy_lab.common.rationals.Rational;
 import org.sosy_lab.java_smt.api.FloatingPointNumber.Sign;
 import org.sosy_lab.java_smt.api.FormulaType.FloatingPointType;
+import org.sosy_lab.java_smt.basicimpl.AbstractFloatingPointFormulaManager.BitvectorFormulaAndBooleanFormula;
```
PhilippWendler: This is an internal package that should not be part of the public API.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -305,6 +307,71 @@ FloatingPointFormula castFrom(
...
+   *
+   * <p>(declare-fun b () (_ BitVec m))
+   *
+   * <p>(assert (= ((_ to_fp eb sb) b) f))
+   *
+   * <p>Note: SMTLIB2 output of this method uses the SMTLIB2 keyword 'to_fp' as described above and
+   * no keyword variation for floating-point to bitvector conversion like 'fp.as_ieee_bv'.
+   *
+   * @param number the {@link FloatingPointFormula} to be converted into an IEEE bitvector.
+   * @param bitvectorConstantName the name of the returned {@link BitvectorFormula}.
```
PhilippWendler: What is "the name of a formula"? I wouldn't know what this refers to.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -305,6 +307,71 @@ FloatingPointFormula castFrom(
...
+   * @return {@link BitvectorFormulaAndBooleanFormula} consisting of the transformed input
+   *     floating-point as a {@link BitvectorFormula} and the additional constraint as {@link
+   *     BooleanFormula}.
+   */
+  BitvectorFormulaAndBooleanFormula toIeeeBitvector(
+      FloatingPointFormula number, String bitvectorConstantName);
```
PhilippWendler: Overloading methods can be confusing and error prone, and should be used only in cases where the semantics are clear and it is obvious what the differences are from the differing parameters.

But here this is clearly not the case. We have an operation (converting from float to bv) that has the obvious signature `FloatingPointFormula` -> `BitvectorFormula`. Nobody will understand why this operation should also take a `String`, and what the difference is between the same operation with and without the `String`.

So I recommend to reconsider the choice for an overload here and suggest renaming the new methods such that their names explain what they do.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -305,6 +307,71 @@ FloatingPointFormula castFrom(
...
+   *
+   * <p>Note: SMTLIB2 output of this method uses the SMTLIB2 keyword 'to_fp' as described above and
+   * no keyword variation for floating-point to bitvector conversion like 'fp.as_ieee_bv'.
+   *
+   * @param number the {@link FloatingPointFormula} to be converted into an IEEE bitvector.
+   * @param bitvectorConstantName the name of the returned {@link BitvectorFormula}.
+   * @param specialFPConstantHandling a {@link Map} defining the returned {@link BitvectorFormula}
+   *     for special {@link FloatingPointFormula} constant numbers. You are only allowed to specify
+   *     a mapping for special FP numbers with more than one well-defined bitvector representation,
+   *     i.e. NaN and +/- Infinity. The width of the mapped bitvector values has to match the
```
PhilippWendler: Are `NaN` and `Infinity` the only allowed values as the "i.e." seems to indicate? If yes, why does the sentence not consist only of a simple "Only NaN and Infinity values are allowed as keys."? If no, please fix. Also note that the JavaDoc of the other overload contains "special FP values (NaN, Inf, etc.)", so the JavaDocs are inconsistent with regards to whether there are special values outside of NaN/Infinity.

The comment also leaves open if one can specify differing NaN values, or whether there must only be one NaN in the map.

Furthermore, why does this map have `FloatingPointFormula` as key type? Shouldn't this be something like `FloatingPointValue`? What happens for example if I add a key that is equivalent to Infinity, but not just the result from `makeInfinity()`?

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
PhilippWendler: If you want to make this a record in the future, you should already now design the (non-private parts of the) API such that making it a record will be compatible.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -542,4 +648,28 @@ protected static String getBvRepresentation(BigInteger integer, int size) {
...
+    private final BitvectorFormula bitvectorFormula;
+    private final BooleanFormula booleanFormula;
+
+    private BitvectorFormulaAndBooleanFormula(
+        BitvectorFormula pBitvectorFormula, BooleanFormula pBooleanFormula) {
+      bitvectorFormula = checkNotNull(pBitvectorFormula);
+      booleanFormula = checkNotNull(pBooleanFormula);
+    }
+
+    private static BitvectorFormulaAndBooleanFormula of(
```
PhilippWendler: What use does a private factory method have that has exactly the same semantics as the constructor?

src/org/sosy_lab/java_smt/api/FormulaType.java
```diff
@@ -320,7 +320,6 @@ public boolean isFloatingPointType() {
       return true;
     }
 
-    /** Returns the size of the exponent. */
```
PhilippWendler, line 329: This should be reverted.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -279,7 +293,99 @@ public BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber) {
...
+      String bitvectorConstantName,
+      Map<FloatingPointFormula, BitvectorFormula> specialFPConstantHandling) {
+
+    FormulaType.FloatingPointType fpType =
+        (FloatingPointType) getFormulaCreator().getFormulaType(f);
+    // The BV is sign bit + exponent + mantissa without hidden bit
+    int mantissaSizeWithoutHiddenBit = fpType.getMantissaSizeWithoutHiddenBit();
+    int exponentSize = fpType.getExponentSize();
+    BitvectorFormula bvFormula =
+        bvMgr.makeVariable(1 + exponentSize + mantissaSizeWithoutHiddenBit, bitvectorConstantName);
```
PhilippWendler: Why does this manually compute the size instead of using an appropriate method?

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -279,7 +293,99 @@ public BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber) {
...
+
+    // When building new Fp types, we don't include the sign bit
+    FloatingPointFormula fromIeeeBitvector =
+        fromIeeeBitvector(
+            bvFormula,
+            getFloatingPointTypeFromSizesWithoutHiddenBit(
+                exponentSize, mantissaSizeWithoutHiddenBit));
```
PhilippWendler: This is an anti-pattern. Almost no code should handle the individual ints that are part of a float type, because it is way too easy to mix them up. In this method there should also be absolutely no reason for accessing the individual sizes: just use the `FloatingPointType` instances that you have already anyway.

The fact that the above comment is wrong confirms my point.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -279,7 +293,99 @@ public BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber) {
...
+
+    // assignment() allows a value to be NaN etc.
+    // Note: All fp.to_* functions are unspecified for NaN and infinity input values in the
+    // standard, what solvers return might be distinct.
+    BooleanFormula additionalConstraint = assignment(fromIeeeBitvector, f);
+
+    // Build special numbers so that we can compare them in the map
+    FloatingPointType precision =
+        getFloatingPointTypeFromSizesWithoutHiddenBit(exponentSize, mantissaSizeWithoutHiddenBit);
```
PhilippWendler: What is the difference between `fpType` and `precision`?

PhilippWendler (commented):
From the comments I do not understand why and when I should use the old or new `toIeeeBitvector` methods. Please explain this in terms that a JavaSMT user could understand.

Clearly this explanation also needs to be added to the JavaDoc of the old method.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: It is the same for FPs in SMTLIB2. But i renamed it (all occurrences in this class) to avoid confusion! Thanks for noticing!

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: Good spot! You are of course correct. I updated it! Thanks.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: I wanted to make the difference of the BV structure compared to the SMTLIB2 definition clear. But since i added it to the documentation of `FloatingPointType::getTotalSize()`, i switched to it and added some doc to make sure.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
baierd: @kfriedberger was right. We already have a implementation for those 2 methods implicitly as part of the `FormulaManager` when asking for the type of a formula! So we can just compare the type given to create a FP formula with the type derived from the created term. (Which was my goal)

Edit: i removed the 2 methods in favor of a single one with a bv formula as input. Users can map special values using ITE.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: If it is OK for you i can remove it and we only reference it in the JavaDoc?

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: I think this is now obsolete due to the already merged PR.

src/org/sosy_lab/java_smt/api/FormulaType.java
baierd: I think this is now obsolete due to the already merged PR.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: SMTLIB2 does not define a default NaN beyond its representation in [SMTLIB2](https://smt-lib.org/theories-FloatingPoint.shtml). But it warns that there are multiple possible binary representations. It states that:
```
 :note
 "There is no function for converting from (_ FloatingPoint eb sb) to the
  corresponding IEEE 754-2008 binary format, as a bit vector (_ BitVec m) with 
  m = eb + sb, because (_ NaN eb sb) has multiple, well-defined representations.
  Instead, an encoding of the kind below is recommended, where f is a term
  of sort (_ FloatingPoint eb sb):

   (declare-fun b () (_ BitVec m))
   (assert (= ((_ to_fp eb sb) b) f))
 "
```

Hence i thought it might be beneficial to have 2 methods; one that handles NaN simply the way the solver handles it, the other in such a way that you can configure it (and other special values, as infinity suffers from a similar problem as NaN). If a user decides to use a distinct encoding of NaN that is not used by the solver per default, it is possible with this method.

However, we can easily make the argument that every user can wrap the `public BitvectorFormulaAndBooleanFormula toIeeeBitvector(FloatingPointFormula f, String bitvectorConstantName)` method in ITE formulas themselves.

Do you think we should just remove the marked method and add some info in the JavaDoc?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: The reason i named it like this is that this is how the SMTLIB2 standard actually recommends this operation to be implemented. It performs the same operation as `BitvectorFormula toIeeeBitvector(FloatingPointFormula pNumber)`, just with an extra input.

The main advantage is that users familiar with `toIeeeBitvector()` find the new call(s) immediately, and it is clear what they do (i.e. they actually perform the same operation).

Note: i now removed the 2 `toIeeeBitvector()` fallback methods you reviewed in favor of a single new one.

baierd:
@kfriedberger @PhilippWendler i now removed the 2 `toIeeeBitvector()` fallback methods that you reviewed, and moved to a single one that simply takes a `BitvectorFormula` and returns the `BooleanFormula` representing the FP = BV equality.

src/org/sosy_lab/java_smt/api/FormulaType.java
baierd: Done. Thanks for spotting this!

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: I removed the reference from the exception now and extended the JavaDoc.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: The class has been removed.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: I removed the method. A user can simply map this at all times by using ITE expressions.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: The method has been removed.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: The class has been removed.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: Note: i replaced the 2 methods with a single one taking only a FP and a BV formula. The behavior is the same as in this method.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: The class has been removed.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -75,44 +81,75 @@ public FloatingPointRoundingMode fromRoundingModeFormula(
...
+  /**
+   * @param pTerm the term to be wrapped in a {@link FloatingPointFormula}.
+   * @param pTypeForAssertions the {@link FloatingPointType} used to create pTerm. This argument is
+   *     only used to verify the exponent and mantissa sizes of pTerm. Ignored if null.
+   * @return input term pTerm in a {@link FloatingPointFormula}.
+   */
+  protected FloatingPointFormula wrapFloatingPointAndAssertType(
+      TFormulaInfo pTerm, @Nullable FloatingPointType pTypeForAssertions) {
+    checkArgument(
+        getFormulaCreator().getFormulaType(pTerm).isFloatingPointType(),
```
PhilippWendler: The expression `getFormulaCreator().getFormulaType(pTerm)` is evaluated up to four times on every wrap call. Why not refactor it out into a variable?

Furthermore, checking whether the type is a floating-point type if we check `equals()` later on anyway is redundant. So I would recommend to execute only one of these checks.

Then the method becomes basically `if ... check1 else check2`. Now it is useless to delegate from `wrapFloatingPoint` to `wrapFloatingPointAndAssertType`. So just put a single check into each of these two methods - much simpler than now.

src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
```diff
@@ -166,10 +166,6 @@ protected BitvectorFormula encapsulateBitvector(TFormulaInfo pTerm) {
   }
 
   protected FloatingPointFormula encapsulateFloatingPoint(TFormulaInfo pTerm) {
-    checkArgument(
-        getFormulaType(pTerm).isFloatingPointType(),
-        "Floatingpoint formula has unexpected type: %s",
-        getFormulaType(pTerm));
```
PhilippWendler, line 171: Why remove these and similar checks?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
PhilippWendler: I just noticed the following: It seems one can use the currently proposed method symmetrically, both for converting from a float to a bitvector and vice versa. After all it creates only an equality between a bv and a float, so I could use it to create an equality between a bitvector constant and a nondet float variable, right?

That would mean that the name `toIeeeBitvector` is bad because it specifies only one direction and instead the name should indicate the mentioned equality between a float and its bitvector representation. That would also eliminate the awkward `bitvectorFormulaSetToBeEqualToFpNumber` parameter name.

Then the method should also be referenced in the JavaDoc of the current inverse of `toIeeeBitvector`.

Btw.: Are there tests for something like I mentioned above?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,50 @@ FloatingPointFormula castFrom(
...
 
+  /**
+   * Create a {@link BooleanFormula} representing the equality of the bitvector representation of
+   * the given {@link FloatingPointFormula}s value with the given {@link BitvectorFormula},
+   * conforming to the IEEE 754-2008 floating-point format. The size m of the given {@link
+   * BitvectorFormula} has to be equal to the sum of the sizes of the exponent eb and mantissa sb
+   * (including the hidden bit) of the given {@link FloatingPointFormula}. This implementation can
+   * be used independently of {@link #toIeeeBitvector(FloatingPointFormula)}, as it does not rely on
+   * an SMT solvers support for {@link #toIeeeBitvector(FloatingPointFormula)}. Behavior for special
+   * FP values (NaN, Inf, etc.) is not defined, and returned values are solver dependent. This
```
PhilippWendler: First, there are no "returned values" here. Second, if behavior is simply mentioned as "not defined" it means that probably almost no caller can safely use it without wrapping it in lots of ITEs, creating a trap in the API. After all, the best way to do something should also be the easiest way.

But I suspect that "not defined" is not really true and one can rely on some behavior rules, just that these are more vague than for other operations, right? What are these rules?

For example, can I rely on the fact that if the float is a NaN that there is at least one bv value that is determined as equal to it? And that no bv value that has a defined float representation could compare as equal to a NaN value?

src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
baierd: They are not removed, just lifted to a central location (the calling function).

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: Good points! I incorporated them. Thank you!


I originally wanted to switch to using "assert" one we ran all tests and confirmed that we use the API as expected. We might still do that. @kfriedberger whats your opinion on this?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: > But I suspect that "not defined" is not really true and one can rely on some behavior rules, just that these are more vague than for other operations, right? What are these rules?

SMTLIB2 applies. Meaning that, from the point of view of SMTLIB2, there is only one NaN etc. The solvers do accept all valid NaN BV representations when transforming from BV to FP as NaN. But they return just one canonical FP NaN representation per default. I added a test for this.
If a user wants custom behavior for special values, it has to be added by the user uniformly over all FPs. (When i realized that, i removed the version of this method that had the ability to add FP number mappings ;D )

Btw. i checked the default BV representations of NaN in all solvers. Some examples:
MathSAT5: `0 11111111 10000000000000000000000`
Z3:               `0 11111111 00000000000000000000001`
Bitwuzla:    `0 11111111 11111111111111111111111`
CVC5:          `1 11111111 11111111111111111111111`

baierd: > For example, can I rely on the fact that if the float is a NaN that there is at least one bv value that is determined as equal to it? And that no bv value that has a defined float representation could compare as equal to a NaN value?

All BV NaN representations seem to work as NaN. When transformed to FP, they are canonized. They do not compare equal as FPs.

Fun fact: no SMT solver returns the same bit representation for NaN as bitvector that is used in the floating-point number when transforming via `toIeeeBitvector()`.

PhilippWendler: Now the JavaDoc just says "Behavior for special FP values (NaN, Inf, etc.), is solver dependent." Wasn't there an explanation about how to handle this at some point, i.e., the hint to handle special values manually by adding conditions?

baierd: Yes, but everyone working with SMT should know how to do this. It might even be dangerous to mention it here; if they don't also modify other sources of BV/FP constants/conversions, unexpected behavior might ensue.

If we knew enough about how these things work or the details, i would be in favor of providing this information. Currently, i don't think that we have the full picture.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -445,9 +445,23 @@ private String sanitize(String formulaStr) {
...
+      if (illegalArgumentException.getMessage() != null
+          && (illegalArgumentException.getMessage().contains("to_ieee_bv")
+              || illegalArgumentException.getMessage().contains("as_ieee_bv"))) {
+        String additionalMessage =
+            "; Note: operations 'to_ieee_bv' and 'as_ieee_bv' are not supported in most SMT"
+                + " solvers. You can try using the SMTLIB2 standards preferred way to encode this"
+                + " operation by utilizing the 'to_fp' operation.";
+        throw new IllegalArgumentException(
+            illegalArgumentException.getMessage() + additionalMessage, illegalArgumentException);
+      }
```
kfriedberger: 4x getMessage --> maybe use a variable for that.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -75,44 +80,77 @@ public FloatingPointRoundingMode fromRoundingModeFormula(
...
   }
 
-  protected FloatingPointFormula wrap(TFormulaInfo pTerm) {
+  /**
+   * @param pTerm the term to be wrapped in a {@link FloatingPointFormula}.
+   * @param pTypeForAssertions the {@link FloatingPointType} used to create pTerm. This argument is
+   *     only used to verify the exponent and mantissa sizes of pTerm.
+   * @return input term pTerm in a {@link FloatingPointFormula}.
+   */
+  protected FloatingPointFormula wrapFloatingPointAndAssertType(
```
kfriedberger: Why not just keep the name `wrap` and make it so complex?

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
```diff
@@ -75,44 +80,77 @@ public FloatingPointRoundingMode fromRoundingModeFormula(
...
+        type.equals(checkNotNull(pTypeForAssertions)),
+        "Floating-Point formula %s type %s is not equal to expected type %s",
+        pTerm,
+        type,
+        pTypeForAssertions);
+
+    return getFormulaCreator().encapsulateFloatingPoint(pTerm);
+  }
+
+  protected FloatingPointFormula wrapFloatingPoint(TFormulaInfo pTerm) {
```
kfriedberger: Is this second method really needed? Why not just name it `wrap` in this context?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
```diff
@@ -196,48 +194,6 @@ protected Term fromIeeeBitvectorImpl(Term pNumber, FloatingPointType pTargetType
...
-        termManager.mk_term(
-            Kind.ITE,
-            termManager.mk_term(Kind.FP_IS_NAN, pNumber),
-            termManager.mk_term(Kind.EQUAL, bvVar, bvNaN),
-            termManager.mk_term(
-                Kind.EQUAL,
-                termManager.mk_term(Kind.FP_TO_FP_FROM_BV, bvVar, sizeExp, sizeSig),
-                pNumber));
-
-    bitwuzlaCreator.addConstraintForVariable(newVariable, equal);
```
kfriedberger: Removing this piece of code will remove the reason for `BitwuzlaFormulaCreator::constraintsForVariables`. Please clean up the unused code there.

kfriedberger:
The idea of this PR is quite similar to the implementation of `IntegerFormulaManager::modularCongruence`, i.e., a left and right operand are combined and a boolean result is returned.

Some request from my side:
- the naming is not optimal: `toIEEEBitvector(fp, bv)` should be renamed to `equal/bitsEqual/bitwiseEqual(fp, bv)`, because this is the exact meaning: assignment/equality of bits and a symmetrical nature. The parameter name is ugly, too.
- cleaning up special logic for basic contraints in solvers.

Overall, the idea is good. If it works for special cases like NaN, Inf, Subnormal, etc, then fine for me.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
kfriedberger: Do we so urgent need to theck the formula-type here, such that we need to extract the type in every call to the wrapper-function? We should not add too many checks to avoid overhead.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaFloatingPointManager.java
baierd: Good spot, thanks! I removed it.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
baierd: Done! Thanks for noticing!

I also extracted it into a method and moved the call due to how parsing was changed in `parseAll()`.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: I renamed both and simplified/combined them. This can of course also be done in a distinct way, but this implementation guarantees that we will always check the correct assertion.

baierd: > Is this second method really needed?

It makes it easier to understand. The assertions were helpful when developing the BV-FP conversion, but are not strictly necessary. We can also use a single method that uses `wrap(term, null)` or so if you want.

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: The more detailed assertions were helpful when developing the BV-FP conversion, as the solvers handle things distinctly.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: I added some new tests for this; when transforming bitvectors to FP numbers, their model values still represent the original BV. So there should not be a problem.

If problems arise, we could provide an option that ensures that this operation, as well as its inverse function, return values that are equal in all solvers that support it. 

I would recommend that CPAchecker switches to using the newly implemented method of this PR per default, as then the SMTLIB2 output of each solver supporting BV to/from FP is parsable by every other solver that supports it. This would also result in some integration/usage testing. (I can take care of this)

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: > Do we so urgent need to theck the formula-type here, such that we need to extract the type in every call to the wrapper-function? We should not add too many checks to avoid overhead.

It was very useful during development to spot issues early. But preconditions are most likely overkill and i would use assertions instead. Are assertions acceptable for you?

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: The check(s) are now assertions

src/org/sosy_lab/java_smt/basicimpl/AbstractFloatingPointFormulaManager.java
baierd: The check(s) are now assertions

baierd:
> the naming is not optimal: toIEEEBitvector(fp, bv) should be renamed to equal/bitsEqual/bitwiseEqual(fp, bv), because this is the exact meaning: assignment/equality of bits and a symmetrical nature. The parameter name is ugly, too.

I chose `toIEEEBitvector(fp, bv)` as the SMTLIB2 standard defines this new implementation as the canonical way of implementing the functionality that we offer with the method `toIEEEBitvector(fp)`. I would argue that changing the name could lead to users not understanding that this offers the same functionality.

Note: It returns a `BooleanFormula` only because the standard defines it that way. We could theoretically switch the bitvector and bool, such that a `BitvectorFormula` is returned, and a `BooleanFormula` is needed as argument. Alternativly: we can always just return a tuple of `<BitvectorFormula, BooleanFormula>`.

> If it works for special cases like NaN, Inf, Subnormal, etc, then fine for me.

It works due to SMTLIB(2) only defining each special FP number only once. Hence "all" bitvector NaN representations return the "same" NaN (as defined by SMTLIB2). If translated back, the original bitvector is returned. I added tests for all special FP numbers that check this.

PhilippWendler:
I agree that the current name is bad as mentioned in https://github.com/sosy-lab/java-smt/pull/512#discussion_r2754935885.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
```diff
@@ -103,9 +102,6 @@ public void popImpl() {
   @CanIgnoreReturnValue
   protected int addConstraint0(BooleanFormula constraint) {
     Term formula = creator.extractInfo(constraint);
-    for (Term t : creator.getConstraintsForTerm(formula)) {
-      formula = creator.getEnv().mk_term(Kind.AND, formula, t);
```
PhilippWendler: I found it surprising that there are such seemingly unrelated changes in this MR. I have no idea whether this is correct, maybe it is unintentional? There are quite a few such changes in the Bitwuzla package.

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
```diff
@@ -299,18 +299,34 @@ protected final void requireBitvectorToInt() {
   }
 
   @SuppressWarnings("CheckReturnValue")
-  protected final void requireFPToBitvector() {
+  protected final void requireNativeFPToBitvector() {
     requireFloats();
     try {
       fpmgr.toIeeeBitvector(fpmgr.makeNumber(0, getSinglePrecisionFloatingPointType()));
     } catch (UnsupportedOperationException e) {
```
PhilippWendler: Catch blocks should only contain the minimal amount of code, in particular here only the actual method call for which you want to catch the exception, not others.

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
```diff
@@ -299,18 +299,34 @@ protected final void requireBitvectorToInt() {
...
+                  + "String, Map) and/or FloatingPointFormulaManager#toIeeeBitvector"
+                  + "(FloatingPointFormula, String).",
+              solverToUse())
           .that(solverToUse())
           .isNull();
     }
   }
 
+  @SuppressWarnings("CheckReturnValue")
+  protected final boolean solverSupportsNativeFPToBitvector() {
```
PhilippWendler: This method is currently unused and there is no obvious situation IMHO where it would be relevant. I would just remove it instead of adding clutter to the API.

Adding methods later on is easier than removing methods that we do not want (to maintain) anymore.

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
```diff
@@ -299,18 +299,34 @@ protected final void requireBitvectorToInt() {
   }
 
   @SuppressWarnings("CheckReturnValue")
-  protected final void requireFPToBitvector() {
+  protected final void requireNativeFPToBitvector() {
```
PhilippWendler: Is this a method that is likely relevant in test scenarios outside of JavaSMT (i.e., for tests of client code that uses JavaSMT), or only for tests of the actual solver-integration code inside JavaSMT?

For things like `requireFloats()` the answer is clearly the former, but for this special method I am not so sure. If the answer is that it is not relevant outside of JavaSMT, I would recommend to make it package private such that it does not clutter the API and it can be evolved without considering compatibility issues.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   * generated containing formulas originating from this method is often not parsable by other
+   * solvers. You can use the method {@link #bitwiseEqual(FloatingPointFormula, BitvectorFormula)}
+   * to avoid both problems.
    */
   BitvectorFormula toIeeeBitvector(FloatingPointFormula number);
 
+  /**
+   * Create a {@link BooleanFormula} representing the equality of the bitvector representation of
+   * the given {@link FloatingPointFormula}s value with the given {@link BitvectorFormula},
+   * conforming to the IEEE 754-2008 floating-point format. The size m of the given {@link
```
PhilippWendler: Not the "equality [comparison]" conforms to the IEEE 754-2008 floating-point format, but the "bitvector representation of the given FloatingPointFormulas value". So the reference to the standard should be attached to the latter in the sentence, but right now it refers to the comparison. You could add the standard reference in parentheses after "representation" or "value".

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
   FloatingPointFormula fromIeeeBitvector(BitvectorFormula number, FloatingPointType pTargetType);
 
   /**
-   * Create a formula that produces a representation of the given floating-point value as a
-   * bitvector conforming to the IEEE 754-2008 FP format. The bit size of the resulting bitvector is
-   * equal to the total size of the {@link FloatingPointNumber}s precision.
+   * Returns a {@link BitvectorFormula} equal to the representation of the given floating-point
+   * value as a bitvector conforming to the IEEE 754-2008 FP format. The bit size of the resulting
+   * bitvector is equal to the total size of the {@link FloatingPointFormula}s {@link
```
PhilippWendler: Add reference to `FloatingPointType#getTotalSize()` here.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   * to avoid both problems.
    */
   BitvectorFormula toIeeeBitvector(FloatingPointFormula number);
 
+  /**
+   * Create a {@link BooleanFormula} representing the equality of the bitvector representation of
+   * the given {@link FloatingPointFormula}s value with the given {@link BitvectorFormula},
+   * conforming to the IEEE 754-2008 floating-point format. The size m of the given {@link
+   * BitvectorFormula} has to be equal to the sum of the sizes of the exponent eb and mantissa sb
+   * (including the hidden bit) of the given {@link FloatingPointFormula}. This implementation can
```
PhilippWendler: 1. Why does this refer to the hidden bit and not the sign bit? After all, the representation as bitvector contains the sign bit, and not the hidden bit, right?
2. Why is the size described differently than in the docs of `toIeeeBitvector()`? If it uses the same size, I guess the description should be the same such that one does not wonder whether one overlooked a difference.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   * be used independently of {@link #toIeeeBitvector(FloatingPointFormula)}, as it does not rely on
+   * an SMT solvers support for {@link #toIeeeBitvector(FloatingPointFormula)}. Behavior for special
+   * FP values (NaN, Inf, etc.), is solver dependent. This method is based on a suggestion in the
```
PhilippWendler:
```suggestion
   * FP values (NaN, Inf, etc.) is solver dependent. This method is based on a suggestion in the
```

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   *
+   * <p>(= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber)
+   *
+   * <p>Example usage in SMTLIB2, asserting the equality of the 2 parameters:
+   *
+   * <p>(declare-fun bitvectorFormulaSetToBeEqualToFpNumber () (_ BitVec m))
+   *
+   * <p>(assert (= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber))
+   *
+   * <p>Note: SMTLIB2 output of this method uses the SMTLIB2 keyword 'to_fp' as described above.
```
PhilippWendler: I don't hink `to_fp` is a keyword, is it?

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   * hidden bit, and eb + sb equal to the bit size of the used {@link BitvectorFormula} parameter,
+   * illustrated in SMTLIB2 as:
+   *
+   * <p>(= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber)
+   *
+   * <p>Example usage in SMTLIB2, asserting the equality of the 2 parameters:
+   *
+   * <p>(declare-fun bitvectorFormulaSetToBeEqualToFpNumber () (_ BitVec m))
+   *
+   * <p>(assert (= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber))
```
PhilippWendler: I find this hard to read and confusing. For example, all this uses `eb` and `sb`, which are not part of the method's signature and not at all relevant to JavaSMT users (which never touch mantissa and significant individually).

Why not say something like "this method is a shortcut for calling `assignment(fromIeeeBitvector(bvValue), floatValue)`, which is what SMT-LIB recommends as replacement for converting from floats to bitvectors".

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   * <p>Note: SMTLIB2 output of this method uses the SMTLIB2 keyword 'to_fp' as described above.
+   *
+   * @param fpNumber the {@link FloatingPointFormula} to be converted into an IEEE bitvector.
+   * @param bitvectorFormulaSetToBeEqualToFpNumber a {@link BitvectorFormula} that is set to be
+   *     equal to the IEEE bitvector representation of the {@link FloatingPointFormula} parameter.
+   * @return a {@link BooleanFormula} representing the result of the equality of the two parameters,
+   *     i.e. (= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber).
+   */
+  BooleanFormula bitwiseEqual(
+      FloatingPointFormula fpNumber, BitvectorFormula bitvectorFormulaSetToBeEqualToFpNumber);
```
PhilippWendler: The parameter names should be improved.

First, an equality comparison is in principle commutative and the method has two equally important parameters. While the method signature forces us to define one parameter as the first, there is no reason to call one "A" and the other "BsetToBeEqualToA", where having "B" and "AsetToBeEqualToB" would be equally valid. Please just use simple terms that describe each parameter. For the same reason the documentation of the parameters also should be updated.

Second, "number" is confusing because floats can be literally "not a number" or other special values that are not numbers (like infinity).

I think the names should be `floatValue` and `bvValue` or something like this.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
```diff
@@ -299,12 +299,49 @@ FloatingPointFormula castFrom(
...
+   *
+   * <p>Example usage in SMTLIB2, asserting the equality of the 2 parameters:
+   *
+   * <p>(declare-fun bitvectorFormulaSetToBeEqualToFpNumber () (_ BitVec m))
+   *
+   * <p>(assert (= ((_ to_fp eb sb) bitvectorFormulaSetToBeEqualToFpNumber) fpNumber))
+   *
+   * <p>Note: SMTLIB2 output of this method uses the SMTLIB2 keyword 'to_fp' as described above.
+   *
+   * @param fpNumber the {@link FloatingPointFormula} to be converted into an IEEE bitvector.
```
PhilippWendler: This is not true (anymore).

PhilippWendler (commented):
I had a look over the API, but no in-depth review.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
baierd: These changes are only seemingly unrelated. This is part of the removal of the previous (insufficient) solution to this problem that was only present in Bitwuzla.
 
I'll extract the Bitwuzla changes into another PR and de-clutter this PR.

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
baierd: Fine for me. (But i really don't see a difference to before. We already knew that the 2 calls won't fail due to `requireFloats()`)

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
baierd: I removed it.

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
baierd: I changed it.

I think it has limited use for users, but i want to incentivize using the new method anyways (as it avoids the problems associated with this one).

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: Shouldn't this even be `FP values (NaN, Inf, etc.) are solver dependent.`?

(Application from here failed... thanks GitHub. I applied it in a normal commit)

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: It is 
image
(Source: https://smt-lib.org/theories-FloatingPoint.shtml)

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
PhilippWendler: The point is that there could be a bug in the methods making them fail unexpectedly.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
PhilippWendler: I would say "behavior for special values *is* solver dependent" is correct.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
PhilippWendler: But is a function symbol. Are these key words?

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
baierd: That i don't dispute, but many other tests would then also fail. But i agree of course that now there is less potential for problems.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: Ups. That's a clear case of "it pays to read the whole sentence". My bad. Fixed.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: Ah now i get the question above! It is of course a function symbol! Thank you! Fixed.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: I would argue it is still true, but the wording can be improved with the new naming.

I propose: `fpNumber the {@link FloatingPointFormula} to be converted into a {@link BitvectorFormula}, equal to the IEEE 754-2008 floating-point format representation of the given floating-point number`

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: I fully agree! I changed it. Good thought!

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: This part is a direct copy from the SMTLIB2 standards recommendation. But i agree that expressing it in our API is better. I changed it and moved the SMTLIB2 example. Please tell me what you think of this new version.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: > 1. Why does this refer to the hidden bit and not the sign bit? After all, the representation as bitvector contains the sign bit, and not the hidden bit, right?

Because that is how the SMT-LIB2 standard defines this.
I am not a big fan of their definitions, but deviating from them introduces even more confusion.

>     2. Why is the size described differently than in the docs of `toIeeeBitvector()`? If it uses the same size, I guess the description should be the same such that one does not wonder whether one overlooked a difference.

I now changed it and use our FP total size method as a reference. Please tell me whether you think this is good. (I like it more than the previous version)

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: I did in the sentence below.

PhilippWendler: This is not pushed so far (only for the JavaDoc of the other method).

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: Fair point. I fixed it. Thank you!

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
baierd: My bad, i had the other methods JavaDoc in mind. I changed it now.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
baierd: Done.

src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
PhilippWendler: You are saying "I would recommend that CPAchecker switches to using the newly implemented method" and also "everyone working with SMT should know how to do this" with respect to handling of special values. But honestly, from reading the JavaDoc it would not be clear to me what I have to do in order to properly migrate the code in CPAchecker to the new method.

baierd: > You are saying "I would recommend that CPAchecker switches to using the newly implemented method" and also "everyone working with SMT should know how to do this" with respect to handling of special values.

I recommend switching because then more solvers can be used with this feature and the SMTLIB2 output of solvers using this new call, instead of `toIeeeBitvector()`, can be parsed by all solvers that support `fromIeeeBitvector()`.
Since every solver that implements `toIeeeBitvector()` on its own uses its own function symbol for it, SMTLIB2 output can only be parsed by the solver that generated it.

As for the special FP values; I don't think that we should be the canonical source of information about these beyond the standard or the solvers. It is unfortunate that there is not much to go by, but i want to avoid putting wrong information out there. The tests give a base sense of what is happening, but we can not guarantee that the behavior is always like this!

> But honestly, from reading the JavaDoc it would not be clear to me what I have to do in order to properly migrate the code in CPAchecker to the new method.

You create a new bitvector variable, replace `toIeeeBitvector()` with the new call and assert the `BooleanFormula` (so that the equality of the FP and BV always holds). The bitvector variable is used in the same way as the return value of `toIeeeBitvector()` was.

The problem here is less the new API in my opinion, but the old one that does not stick to the standard.

PhilippWendler: But how do I handle special values? This is what you referred to with "everyone working with SMT should know how to do this", but I don't know what I need to do and you also didn't answer this in your response.

baierd: The easiest is using ITE constructs, but you can of course use any logical formula that fits your purpose. You can for example check the bits in the bitvector (that is equal to your FP number) for the NaN pattern, and if it is some NaN, you return a single, canonical NaN, else the initial bitvector.

Note: maybe a public example (not a test) that shows exactly this would be a good idea?

PR #514, opened 2025-09-02
Use Solver Native Implementation of Bitvector to IEEE FP for CVC4 and CVC5

baierd:
This PR changes the implementation of CVC4 and 5 to use their built-in version of `fromIeeeBitvector()`.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FloatingPointFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FloatingPointFormulaManager.java
@@ -408,29 +408,18 @@ public class CVC5FloatingPointFormulaManager
   }
 
   @Override
-  protected Term fromIeeeBitvectorImpl(Term bitvector, FloatingPointType pTargetType) {
-    int mantissaSize = pTargetType.getMantissaSize();
-    int exponentSize = pTargetType.getExponentSize();
-    int size = pTargetType.getTotalSize();
-    assert size == mantissaSize + exponentSize + 1;
-
-    Op signExtract;
-    Op exponentExtract;
-    Op mantissaExtract;
+  protected Term fromIeeeBitvectorImpl(Term pBitvector, FloatingPointType pTargetType) {
     try {
-      signExtract = termManager.mkOp(Kind.BITVECTOR_EXTRACT, size - 1, size - 1);
-      exponentExtract = termManager.mkOp(Kind.BITVECTOR_EXTRACT, size - 2, mantissaSize);
-      mantissaExtract = termManager.mkOp(Kind.BITVECTOR_EXTRACT, mantissaSize - 1, 0);
-    } catch (CVC5ApiException e) {
-      throw new IllegalArgumentException(
-          "You tried creating a invalid bitvector extract in term " + bitvector + ".", e);
+      return termManager.mkTerm(
+          termManager.mkOp(
+              Kind.FLOATINGPOINT_TO_FP_FROM_IEEE_BV,
+              pTargetType.getExponentSize(),
+              pTargetType.getMantissaSize() + 1), // add sign bit
+          pBitvector);
+    } catch (CVC5ApiException pE) {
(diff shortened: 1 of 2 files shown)


baierd:
@kfriedberger i think this branch is ready to be reviewed. The CI fails for some timeout reason and needs to be restarted.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FloatingPointFormulaManager.java
```diff
@@ -408,29 +408,18 @@ protected Term isNegative(Term pParam) {
...
-    } catch (CVC5ApiException e) {
-      throw new IllegalArgumentException(
-          "You tried creating a invalid bitvector extract in term " + bitvector + ".", e);
+      return termManager.mkTerm(
+          termManager.mkOp(
+              Kind.FLOATINGPOINT_TO_FP_FROM_IEEE_BV,
+              pTargetType.getExponentSize(),
+              pTargetType.getMantissaSize() + 1), // add sign bit
+          pBitvector);
+    } catch (CVC5ApiException cvc5ApiException) {
```
kfriedberger, line 419: We catch other CVC5ApiExceptions as IllegalStateException, because it shows a bug in JavaSMT or a missed illegal usage of the API. Can we replace this plain RuntimeException?

kfriedberger (approved):
Overall, a good catch. We seem to have missed this simple operation in CVC4/CVC5 some years ago.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FloatingPointFormulaManager.java
baierd: Good question!

The documentation of CVC5 is a little sparse when it comes to the reason why exceptions are thrown:
```java
/**
   * Create operator of Kind:
   * <ul>
   *   <li>BITVECTOR_EXTRACT</li>
   *   <li>FLOATINGPOINT_TO_FP_FROM_IEEE_BV</li>
   *   <li>FLOATINGPOINT_TO_FP_FROM_FP</li>
   *   <li>FLOATINGPOINT_TO_FP_FROM_REAL</li>
   *   <li>FLOATINGPOINT_TO_FP_FROM_SBV</li>
   *   <li>FLOATINGPOINT_TO_FP_FROM_UBV</li>
   * </ul>
   * See enum {@link Kind} for a description of the parameters.
   * @param kind The kind of the operator.
   * @param arg1 The first unsigned int argument to this operator.
   * @param arg2 The second unsigned int argument to this operator.
   * @return The operator.
   * @throws CVC5ApiException on error
   */
 ```
 However, from the code it is clear that the reasons can be negative integer input, and whatever exception is thrown by native code is also thrown in the JNI wrapper. The native code checks that the given FP size arguments fit the bitvector size as far as i can see. We could/should just check the given FP precision against the BV size ourselves, and this exception is never thrown.

baierd:
> Overall, a good catch. We seem to have missed this simple operation in CVC4/CVC5 some years ago.

Their naming (especially in CVC4) seems a little counter-intuitive at first. I only noticed because i had the SMTLib2 standard and the CVC5 API documentation open while working on FP stuff.

Actually, we could add your implementation as a default implementation, then we would not have to worry about any future solvers not supporting this.

baierd:
> Actually, we could add your implementation as a default implementation, then we would not have to worry about any future solvers not supporting this.

OK, that's not possible in general. The solvers generally do not allow the creation of new FP numbers from 3 BV terms like CVC4/5. And since all solver support the BV to IEEE FP operation, i won't bother.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FloatingPointFormulaManager.java
baierd: I've added a default check for the used sizes and extended the documentation in CVC5 and switched to an `IllegalArgumentException`.

kfriedberger (approved):

PR #515, opened 2025-09-03
replace all usages of "List.of" with more readable "ImmutableList.of/copyOf"

kfriedberger:
… to directly show the immutability.

 This might optimise internal operations, e.g.,
 avoid copies from List to other Lists, if all are ImmutableLists.

### DIFF
--- a/src/org/sosy_lab/java_smt/api/BitvectorFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/api/BitvectorFormulaManager.java
@@ -8,6 +8,7 @@
 
 package org.sosy_lab.java_smt.api;
 
+import com.google.common.collect.ImmutableList;
 import java.math.BigInteger;
 import java.util.List;
 import org.sosy_lab.java_smt.api.FormulaType.BitvectorType;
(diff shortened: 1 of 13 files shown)


src/org/sosy_lab/java_smt/solvers/cvc4/CVC4NativeAPITest.java
```diff
@@ -766,7 +766,7 @@ public void checkUnsatCore() {
 
     // UnsatCores are iterable
     for (Expr e : unsatCore) {
-      assertThat(e.toString()).isIn(Arrays.asList("(not (or a b))", "(and a b)"));
+      assertThat(e.toString()).isIn(ImmutableList.of("(not (or a b))", "(and a b)"));
```
PhilippWendler: For this and similar cases I would simply use `isAnyOf()`.

kfriedberger: done

baierd:
Good idea!

Question: do we also want to switch to `ImmutableList` etc. for internal API declarations?
E.g. `protected abstract TFormulaInfo distinctImpl(ImmutableList<TFormulaInfo> pNumbers);`.

kfriedberger:
For the internal method signature, the benefit is quite small. Keeping signatures with `List` seems less strict for implementations, and avoids potential copies if a method already gets a list from calling code.

PhilippWendler:
This PR has the same problem as I mentioned in https://github.com/sosy-lab/java-smt/commit/d344e3a797c783a8f44bcf4a1d6d9978ae0881cb#commitcomment-165053827: It *adds* additional unnecessary copies by replacing `Arrays.asList` with a copy.

PR #517, opened 2025-09-06
Improve separation logic (CVC4 and CVC5)

kfriedberger:
Our implementation for SL seems to have been broken since a long time.

This PR improves the situation:
- for CVC4, which is deprecated, we noticed a broken method for SEP_NIL (i.e., the NULL element in SL) in the bindings. There is currently  no way to fix it. :cry: 
- for CVC5, we properly implement the required options for solving SL queries and fix the behaviour of the non-incremental solver stack used for SL in CVC5.
- we add tests for SL, which were missing since the initial implementation. :cry:

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4SLFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4SLFormulaManager.java
@@ -45,6 +45,14 @@ public class CVC4SLFormulaManager extends AbstractSLFormulaManager<Expr, Type, E
 
   @Override
   protected Expr makeNilElement(Type pType) {
+    // TODO CVC4 bindings for Java do not support creation of SEP_NIL via mkTerm.
+    //  It is unclear whether this method ever worked.
+    //  CVC4 is deprecated and we will not investigate here.
+    //
+    // Executing this method will cause the following exception:
+    //     Illegal argument detected: CVC4::Expr CVC4::ExprManager::mkExpr(CVC4::Kind, CVC4::Expr)
+    //     `kind' is a bad argument; ... Only operator-style expressions are made with mkExpr();
+    //     to make variables and constants, see mkVar(), mkBoundVar(), and mkConst().
     return exprManager.mkExpr(Kind.SEP_NIL, pType.mkGroundTerm());
   }
 }
--- a/src/org/sosy_lab/java_smt/api/SLFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/api/SLFormulaManager.java
@@ -2,7 +2,7 @@
 // an API wrapper for a collection of SMT solvers:
 // https://github.com/sosy-lab/java-smt
 //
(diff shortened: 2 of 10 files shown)


src/org/sosy_lab/java_smt/solvers/cvc4/CVC4SLFormulaManager.java
```diff
@@ -45,6 +45,14 @@ protected Expr makeEmptyHeap(Type pT1, Type pT2) {
...
   protected Expr makeNilElement(Type pType) {
+    // TODO CVC4 bindings for Java do not support creation of SEP_NIL via mkTerm.
+    //  It is unclear whether this method ever worked.
+    //  CVC4 is deprecated and we will not investigate here.
+    //
+    // Executing this method will cause the following exception:
+    //     Illegal argument detected: CVC4::Expr CVC4::ExprManager::mkExpr(CVC4::Kind, CVC4::Expr)
+    //     `kind' is a bad argument; ... Only operator-style expressions are made with mkExpr();
+    //     to make variables and constants, see mkVar(), mkBoundVar(), and mkConst().
     return exprManager.mkExpr(Kind.SEP_NIL, pType.mkGroundTerm());
```
daniel-raffler: I looked it up and `exprManager.mkNullaryOperator(pType, Kind.SEP_NIL)` seems to do the trick. Two of the tests are still broken on CVC4, though

daniel-raffler (approved):
Documentation and tests are always an improvement

LGTM!

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4SLFormulaManager.java
kfriedberger: Thanks. I followed your hint and got the working solution. :smile: :+1: 

PS: The words `I looked it up` should always have some reference **where** this was looked up. :receipt:

daniel-raffler: > PS: The words `I looked it up` should always have some reference **where** this was looked up. 🧾

It's straight from the C++ API definitions [here](https://github.com/CVC4/CVC4-archived/blob/40eac7f0529176bcc8464d6c4c8804fbde628c2b/src/api/cvc4cpp.cpp#L3301):
```
Term Solver::mkSepNil(Sort sort) const
{
  CVC4_API_SOLVER_TRY_CATCH_BEGIN;
  CVC4_API_ARG_CHECK_EXPECTED(!sort.isNull(), sort) << "non-null sort";
  CVC4_API_SOLVER_CHECK_SORT(sort);

  Expr res = d_exprMgr->mkNullaryOperator(*sort.d_type, CVC4::kind::SEP_NIL);
  (void)res.getType(true); /* kick off type checking */
  return Term(this, res);

  CVC4_API_SOLVER_TRY_CATCH_END;
}
```

I'm not sure why the function didn't make it to the Java API 😅. The [documentation](https://cvc4.github.io/separation-logic) suggest we should use `exprManager.mkUniqueVar(T,kind::SEP_NIL)`, but that function no longer seems to exist

PR #518, opened 2025-09-06
Princess/Ostrich update

daniel-raffler:
Hello everyone,
this will update Princesss to version  2025-06-25 and Ostrich to 2.0. Our current version is still from the end of last year and there have been several improvements since then. Specifically, Princess now has special `IFunction`s for rational operations like additional or division. The old version would rewrite rational formulas as integer-only formulas, which makes it hard to restore the original formula for the visitor. In Ostrich one of our [bugs](https://github.com/uuverifiers/princess/issues/18) was closed, and `str_from_code` is now supported.

@kfriedberger: Could you please upload the packages from Maven? Here is what I used locally for testing:
```
ant install -Dorganisation=io.github.uuverifiers -Dmodule=princess_2.13 -Drevision=2025-06-25
ant install -Dorganisation=io.github.uuverifiers -Dmodule=ostrich_2.13 -Drevision=2.0
```


src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
```diff
@@ -142,7 +135,7 @@ protected IExpression multiply(IExpression number1, IExpression number2) {
 
   @Override
   protected IExpression divide(IExpression number1, IExpression number2) {
-    return Rationals.div(toType(number1), toType(number2));
+    return Rationals.divWithSpecialZero(toType(number1), toType(number2));
```
kfriedberger, line 141: What is special about zero in this case? Please document as comment or provide a test that shows the behaviour, if the test does not yet exist.

kfriedberger (approved):
Overall valid.

Princess comes already from Maven. There is no need to upload it to Maven. We (aka SosyLab) just upload it to the SosyLab-Ivy (I will check this).

- Maven: https://central.sonatype.com/artifact/io.github.uuverifiers/princess_2.13
- Ivy: https://www.sosy-lab.org/ivy/io.github.uuverifiers/princess_2.13

kfriedberger:
@daniel-raffler You could ask @PhilippWendler for an SVN account to upload such files directly in the future. Sosy-Ivy has strict rules, e.g., it is append-only and quite strcitly structured, and commits follow a specific pattern.

daniel-raffler:
> @daniel-raffler You could ask @PhilippWendler for an SVN account to upload such files directly in the future. Sosy-Ivy has strict rules, e.g., it is append-only and quite strcitly structured, and commits follow a specific pattern.

Thanks for your help with the binaries! I already have an account from when CPAchecker still used svn, but last time I tried to upload binaries to the Ivy repository it was still rejected. However, this could just be a problem with my configuration, so I'll try again next time

src/org/sosy_lab/java_smt/solvers/princess/PrincessRationalFormulaManager.java
daniel-raffler: I've added a short comment [here](https://github.com/sosy-lab/java-smt/pull/518/commits/11d1c9ba460dc848eb8fd43fc79fac7dfb35f762) and updated one of the tests to show the difference

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
```diff
@@ -360,8 +363,11 @@ private String getName(IExpression input) {
     }
   }
 
-  /** Returns true if the expression is a constant number. */
-  private static boolean isConstant(IFunApp pExpr) {
+  /** Returns true if the expression is a constant value. */
+  private static boolean isValue(IFunApp pExpr) {
+    if (!pExpr.fun().equals(Rationals.frac()) && !pExpr.fun().equals(Rationals.fromRing())) {
+      return false;
+    }
```
kfriedberger: Does this work for non-rational numbers? I would assume that Intgers are not longer handled here.

src/org/sosy_lab/java_smt/test/NonLinearArithmeticTest.java
```diff
@@ -224,6 +216,14 @@ public void testDivisionByZero() throws SolverException, InterruptedException {
...
 
     assertThatFormula(f).isSatisfiable();
+
+    // Division by zero is still a function. So, if (/0 a) = b and (/0 a) = c, then b=c must hold
+    BooleanFormula g =
+        bmgr.and(
+            nmgr.equal(nmgr.divide(a, zero), nmgr.makeNumber(2)),
+            nmgr.equal(nmgr.divide(a, zero), nmgr.makeNumber(4)));
+
+    assertThatFormula(g).isUnsatisfiable();
```
kfriedberger, line 229: This example does not check `((/0 a)=b && (/0 a)=c) => b=c`, but only `(/0 a)!=2 || (/0 a)!=4`, which does not tell us anything aboud the result of the division at all..

daniel-raffler: I've updated the test to use the actual formula

src/org/sosy_lab/java_smt/solvers/princess/PrincessFormulaCreator.java
daniel-raffler: No, it will only detect rational number literals. Other types of literals were handled separately in the visitor and have their own tests. However, since the name was a little confusing I reorganized the code in https://github.com/sosy-lab/java-smt/pull/518/commits/8d5fcd1454103ebfd1f53abd842555eea07f2e05 and all literals are now handled by `isValue`

kfriedberger (approved):
can be merged

src/org/sosy_lab/java_smt/test/RationalFormulaManagerTest.java
```diff
@@ -126,11 +126,11 @@ public void forallFloorIsLessThanValueTest() throws SolverException, Interrupted
   public void visitFloorTest() {
     requireRationals();
     requireRationalFloor();
-    // TODO Princess will rewrite floor. Add backtranslation in the visitor
+    // TODO Princess will rewrite floor. Add back translation in the visitor
```
kfriedberger, line 129: No, we do not need to add a precise backwards-mapping. The solver is allowed to rewrite the formula.

kfriedberger: The issue here is that Princess uses quantifiers and we do not handle quantified formulas correctly in some cases.

kfriedberger (approved):

src/org/sosy_lab/java_smt/test/RationalFormulaManagerTest.java
daniel-raffler: Yes, the comment was about quantifiers and/or epsilon terms that Princess likes to introduce for operations like these

Issue #520, opened 2025-09-09
Add support for arrays with multiple indices in the Princess model


daniel-raffler:
Hello everyone,
Princess fails `ModelTest.testGetArrays3` with an exception:
```
java.lang.IllegalArgumentException: unhandled model value store(const(const(0)), 3, store(const(124), 1, 123)) of type class ap.parser.IFunApp
	at org.sosy_lab.java_smt.solvers.princess.PrincessFormulaCreator.convertValue(PrincessFormulaCreator.java:250)
	at org.sosy_lab.java_smt.solvers.princess.PrincessFormulaCreator.convertValue(PrincessFormulaCreator.java:80)
	at org.sosy_lab.java_smt.solvers.princess.PrincessModel.getAssignmentsFromArraySelect(PrincessModel.java:221)
	at org.sosy_lab.java_smt.solvers.princess.PrincessModel.getAssignments(PrincessModel.java:156)
	at org.sosy_lab.java_smt.solvers.princess.PrincessModel.asList(PrincessModel.java:88)
	at org.sosy_lab.java_smt.basicimpl.CachingModel.asList(CachingModel.java:40)
	at org.sosy_lab.java_smt.api.Model.iterator(Model.java:55)
	at org.sosy_lab.java_smt.test.ModelTest.testModelIterator(ModelTest.java:1623)
```

The problem is on our side as the backend for Princess does not support array with more than one index while creating the model. We should add this missing feature. While we're at it we may also want to look into moving (part) of the conversion into an abstract class. The algorithm is always the same and we currently implement it for several solvers separately

daniel-raffler:
I've added some tests [here](https://github.com/sosy-lab/java-smt/commit/316a3f3d53f778fca739b5f22928dcccf01dd6cc) and it turns out that model generation for array values is broken in most solvers:
- Princess, Z3 and Yices2 may crash if multiple indices are used
- Z3 often misses a mapping as it simplifies `(store array idx val)` to `(const val)` if there are no other mappings (and we can't handle default values)
- CVC4 and CVC5 return `Store` terms as values
- Bitwuzla doesn't generate any assignments

[...]

Issue #519, opened 2025-09-09
CVC5 fails while calculating an interpolant sequence


daniel-raffler:
Hello everyone,
CVC5 crashes in two tests from `InterpolatingProverTest`:  `sequentialInterpolation` and `sequentiaBVlInterpolation`. Both test are currently disabled for CVC5 due to performance issues.However, theses issues seem to be resolved, and instead we're now getting an exception:
```
io.github.cvc5.CVC5ApiException: invalid call to 'cvc5::Sort cvc5::Term::getSort() const', expected non-null object
	at io.github.cvc5.Term.getSort(Native Method)
	at io.github.cvc5.Term.getSort(Term.java:140)
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5FormulaCreator.getFormulaType(CVC5FormulaCreator.java:214)
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5FormulaCreator.encapsulateBoolean(CVC5FormulaCreator.java:298)
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5InterpolatingProver.getSeqInterpolants(CVC5InterpolatingProver.java:124)
	at org.sosy_lab.java_smt.basicimpl.withAssumptionsWrapper.InterpolatingProverWithAssumptionsWrapper.getSeqInterpolants(InterpolatingProverWithAssumptionsWrapper.java:61)
	at org.sosy_lab.java_smt.test.InterpolatingProverTest.sequentialInterpolation(InterpolatingProverTest.java:293)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
```
The issue is reproducible for me when I run the test locally. However, in the [CI](https://gitlab.com/sosy-lab/software/java-smt/-/pipelines/2028420782) it only seemed to appear for the run on Java23.

This might be a phantom reference issues as we use a new `ProverEnvironment` to calculate each of the interpolants of the sequence. I believe that when this solver gets closed it also deletes the calculated interpolant.

Calculating all interpolants on the same stack does seem to fix the issue. However, we should look further into it and report the bug to the CVC5 developers if it's really a problem with the solver

kfriedberger:
This looks like an invalid interpolation query from our side. The returned interpolant is "null", wrapped in a CVC5-term. The terms phi+ and phi- seem to have disjunct variables, aka there is no possible interpolant. We need to check our interpolation algorithm here.

[...]

PR #521, opened 2025-09-09
Re-enable tests that were previously disabled

daniel-raffler:
Hello,
there are several tests that were disabled (for some solvers) due to bugs that have now been resolved. This PR tries to re-enable as many of them as possible. Currently there are still two issues that were previously hidden by the disabled tests and should be resolved first. I've created separate issues here:
- https://github.com/sosy-lab/java-smt/issues/520
- https://github.com/sosy-lab/java-smt/issues/519


daniel-raffler:
Thanks for the merge! I think the rest can be handled elsewhere

PR #522, opened 2025-09-10
fix invalid interpolation query for CVC5.

kfriedberger:
Since class Term implements Iterable of Terms,
the append-method was iterating over the terms children, and for TRUE/FALSE, there are none.
Actually a tricky, quite simple bug.

### DIFF
--- a/src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
+++ b/src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
@@ -421,11 +421,6 @@ public class InterpolatingProverTest extends SolverBasedTest0.ParameterizedSolve
   public <T> void sequentialBVInterpolation() throws SolverException, InterruptedException {
     requireBitvectors();
 
-    assume()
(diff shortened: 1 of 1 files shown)


src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -262,11 +262,6 @@ public <T> void sequentialInterpolation() throws SolverException, InterruptedExc
     InterpolatingProverEnvironment<T> stack = newEnvironmentForTest();
     requireIntegers();
 
-    assume()
-        .withMessage("Solver %s runs into timeout on this test", solverToUse())
-        .that(solverToUse())
-        .isNotEqualTo(Solvers.CVC5);
-
```
daniel-raffler, line 269: You could also turn `sequentialBVInterpolation` back on

kfriedberger: yes. done

PR #523, opened 2025-09-11
Add support for arrays in Yices2

daniel-raffler:
Hello,
in this PR we add support for arrays to the Yices2 backend. Yices has supported arrays for a while, however, the implementation is a bit non-standard. Instead of the usual array operations from SMTLIB we have to use function terms:
- `Const` become `(lambda (x::T) c)` where `T` is the index type and `c` the default element
- `Select` becomes regular function application
- `Store` become [function update](https://yices.csl.sri.com/doc/term-operations.html#c.yices_update) where we "fix" the function for a single value

Array types from SMTLIB then map to (curried) function types in Yices2

There can be some confusion between UF terms and arrays as both are just function terms in Yices2. We work around issue this by explicitly keeping track of all our UF declarations in the `FormulaCreator` and assume all other function terms must be arrays

### DIFF
--- a/src/org/sosy_lab/java_smt/test/SolverFormulaIOTest.java
+++ b/src/org/sosy_lab/java_smt/test/SolverFormulaIOTest.java
@@ -532,7 +532,6 @@ public class SolverFormulaIOTest extends SolverBasedTest0.ParameterizedSolverBas
     } catch (AssumptionViolatedException ave) {
       // ignore, i.e., do not report test-case as skipped.
     }
-    ;
   }
 
   private BooleanFormula genBoolExpr() {


src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
```diff
@@ -398,11 +466,26 @@ private <R> R visitFunctionApplication(
...
+        if (ufSymbols.contains(fun)) {
+          functionKind = FunctionDeclarationKind.UF;
+          functionArgs = getArgs(pF);
+          functionName = yices_get_term_name(functionArgs.get(0));
+          functionDeclaration = functionArgs.get(0);
+          functionArgs.remove(0);
+        } else {
+          functionKind = FunctionDeclarationKind.SELECT;
+          functionArgs = getArgs(pF);
+          functionName = "select";
```
kfriedberger, line 479: is every YICES_APP_TERM that is not a UF automatically a SELECT?

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
```diff
@@ -170,6 +182,14 @@ public class Yices2FormulaCreator extends FormulaCreator<Integer, Integer, Long,
    */
   private final Table<String, Integer, Integer> formulaCache = HashBasedTable.create();
 
+  /**
+   * List of all UF function declarations.
+   *
+   * <p>We need this in the visitor to tell the difference between functions and arrays. Both are
+   * modeled internally by function terms.
+   */
+  private final Set<Integer> ufSymbols = new HashSet<>();
```
kfriedberger, line 191: We need to keep this in mind:
- formula created via API -> handled with this set
- formulas created via parser -> Yices2 does not support parsing SMTLIB -> irrelevant for now
- formulas returned from Yices2, e.g., interpolants or unsat cores -> can only contain existing UFs.

The set is a good starting point, however, we might want to avoid it.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApi.java
```diff
@@ -89,6 +89,7 @@ private Yices2NativeApi() {}
   // MAX_INT avoids collisions with existing constants
   public static final int YICES_AND = Integer.MAX_VALUE - 1;
   public static final int YICES_BV_MUL = Integer.MAX_VALUE - 2;
+  public static final int YICES_ARRAY_EVAL = Integer.MAX_VALUE - 3;
```
kfriedberger: Can we just name it "YICES_ARRAY_SELECT" to recognize its usage in SMTLIB context?

src/org/sosy_lab/java_smt/test/SolverTheoriesTest.java
```diff
@@ -716,6 +716,9 @@ public void testMakeIntArray() {
         // INFO: OpenSmt changes the order of the terms in the sum
         assertThat(_b_at_i_plus_1.toString()).isEqualTo("(select b (+ 1 i))");
         break;
+      case YICES2:
+        assertThat(_b_at_i_plus_1.toString()).isEqualTo("(b (+ 1 i))");
```
kfriedberger, line 720: That is a really weird way of writing SELECT. Well, it seems to work for now.

Can we try to replace parts of the dumped output to recognize select more easily.
Example: replace all substrings like "(x y)" where x is an array (or not a default operator or UF) with "(select x y)"

src/org/sosy_lab/java_smt/solvers/yices2/Yices2NativeApi.java
kfriedberger: done.

daniel-raffler:
> The implementation looks good at first view.
> 
> Does it work with existing Yices2 bindings or do we need an update? I saw the other PR. I am just wondering, whether we can merge this first.

Hello Karlheinz,
thanks for the quick review! Arrays will work with the current version of Yices, so we could merge this before the update

src/org/sosy_lab/java_smt/test/SolverTheoriesTest.java
daniel-raffler: > Example: replace all substrings like "(x y)" where x is an array (or not a default operator or UF) with "(select x y)"

Yes, that should work, and probably wouldn't be too much overhead

If we also want to substitute the lambdas and function updates it might be easier to just use a visitor for formula printing. I was thinking about writing a generic formula printer anyway. However, its probably better to wait for https://github.com/sosy-lab/java-smt/pull/507 for this

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
daniel-raffler: I was hoping yes 😅

It's quite hard to track all array values, and the solver may also create new ones, for instance as part of the model. For UFs it should be easier, as we don't have parsing in Yices and solvers otherwise shouldn't create new UF symbols that are unknown to the user. (Or if they do, it's in a limited way, for instance Princess will include a new symbol `arrayConstant` in it's model, but these can easily be filtered out)

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
daniel-raffler: > The set is a good starting point, however, we might want to avoid it.

I'd certainly prefer to get rid of it. The issue is that Yices doesn't support Arrays directly. Instead it has "updatable" functions that can be used as arrays. From that perspective an array is just a function that maps from the index to the elements. However, when we try to translate this back to the SMTLIB view things get a little complicated as not all functions are arrays. That's why I think we'll need some way to remember which of the functions are UFs, and which are really arrays.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2FormulaCreator.java
kfriedberger: Thanks for the explanation. We can stay with this solution until we notice other issues.

daniel-raffler:
Thanks for merging! I think we still need to update `javasmt-yices2` before the next release? However, this can be done after we've updated to yices 2.7.0

PR #524, opened 2025-09-11
Update Yices to version 2.8.0

### LINKED ISSUE #402: Yices2 produces invalid SMTLIB output when dumping formulas
Hello everyone,
yices2 does not apply correct quotes while dumping SMTLIB formulas. An example for this can be seen in [VariableNamesTest.testBoolVariableDump()](https://github.com/sosy-lab/java-smt/blob/fed0e55e44243a26cc0ca00b90d6835aef6e5d55/src/org/sosy_lab/java_smt/test/VariableNamesTest.java#L557) where the formula for a (boolean) variable named "(" produces the following output:

```
(declare-fun |(| () Bool)
(assert ()
```
The correct output should use quotes for the "(" in the assertion:
```
(declare-fun |(| () Bool)
(assert |(|)
```

Without the quotes the SMTLIB string is invalid and will cause an exception during post-processing. (Which is why the test was temporarily disabled for yices2)

The problem is with the yices2 function `yices_term_to_string()`, which fails to apply the quotes while printing the formula. We should report this to the yices2 developers and wait for a fix. If this is not an option we may want to handle quoting on our end by overloading `Yices2FormulaManager.makeVariable.

daniel-raffler:
Hello,
this PR updates Yices to version 2.7.0 and closes #402. Other than that the new version mainly includes improvements to the MCSat solver, which we're not using at the moment. [Here](https://github.com/SRI-CSL/yices2/releases/tag/yices-2.7.0) is a full changelog.

Yices2 has some additional features that we may want to include:
- Nonlinear arithmetic is supported with the MCSat solver
- MCSat also support interpolation ([link](https://yices.csl.sri.com/doc/context-operations.html#c.yices_check_context_with_interpolation)), but not `unsat core` ([yet?](https://github.com/SRI-CSL/yices2/pull/578))
- Dpll (which we currently use) has `unsat core`, but no interpolation or non-linear arithmetic
- Only Dpll can be compiled with the `--thread-safety` option that allows multiple solver instances to run in parallel
- Arrays have been added separately, see [here](https://github.com/sosy-lab/java-smt/pull/523) for the other pull request
- Yices can use Cadical/Kissat/CryptoMiniSat as backend for SAT solving. This could be significantly faster, but is currently limited to `QF_BV` logic

In this PR I stuck with Dpll as solver and just enabled the `--thread-safety` option. We could still add MCSat support and maybe even the external SAT solvers. Yices then picks the solver based on the logic, which would give us the best feature set. The big downside is that MCSat is not compatible with `--thread-safety`, which means that multiple parallel solver instances are not possible

**Update**

The thread-safety issues in MCSAT have been fixed, and unsat core is now also supported in MCSAT. We therefore switched to MCSAT as the default solver
- A new option `solver.yices2.solverType=dpllt` has been added to use dpllt instead
- Interpolation is now supported, thanks to the MCSAT solver
- Nonlinear arithmetics and division by zero are also working now

We also switched the entire backend to the official Java bindings for Yices. In addition, a new build script has been added, and Windows binaries are now available

This PR is currently waiting on https://github.com/SRI-CSL/yices2/pull/614, but should be ready to merge otherwise

**Linked issues:**
- https://github.com/SRI-CSL/yices2/issues/608
- https://github.com/SRI-CSL/yices2/issues/613
- https://github.com/SRI-CSL/yices2/issues/615
- https://github.com/SRI-CSL/yices2/issues/616
- https://github.com/SRI-CSL/yices2_java_bindings/pull/8


daniel-raffler:
@kfriedberger
I can try uploading the binaries this time. What architecture should I compile for? Yices2 seems to use `--march=znver3` when I compile it on my laptop. Is that OK, or should I try to pick something else?

kfriedberger:
@daniel-raffler 
We did use the Ubuntu docker image (18.04 if possible without further overhead, else the 22.04) for building binary solvers.
The steps for publication are documented in https://github.com/sosy-lab/java-smt/blob/master/doc/Developers-How-to-Release-into-Ivy.md#publishing-yices2 ... wait ... that seems quite outdated for Yices2. I will have a deeper look the next days.

lib/native/source/yices2j/org_sosy_1lab_java_1smt_solvers_yices2_Yices2NativeApi.c
```diff
@@ -1598,6 +1598,13 @@ DEFINE_FUNC(int, 1is_1thread_1safe) WITHOUT_ARGS
 CALL0(int, is_thread_safe)
 INT_RETURN
 
+/*
+ * Check if yices was compiled with MCSAT support
+ */
+DEFINE_FUNC(int, 1has_1mcsat) WITHOUT_ARGS
+CALL0(int, has_mcsat)
+INT_RETURN
```
kfriedberger: We could maybe add BOOL_RETURN for such methods.

kfriedberger (commented):
Overall looks good. I will make a deeper test the next days.

daniel-raffler:
> @daniel-raffler We did use the Ubuntu docker image (18.04 if possible without further overhead, else the 22.04) for building binary solvers. The steps for publication are documented in https://github.com/sosy-lab/java-smt/blob/master/doc/Developers-How-to-Release-into-Ivy.md#publishing-yices2 ... wait ... that seems quite outdated for Yices2. I will have a deeper look the next days.

Thanks! I've been able to build the binaries in the 18.04 image

Should I build with [--thread-safety](https://github.com/SRI-CSL/yices2?tab=readme-ov-file#support-for-thread-safety) or [--enable-mcsat](https://github.com/SRI-CSL/yices2?tab=readme-ov-file#support-for-non-linear-arithmetic-and-mc-sat)? Unfortunately these options can't be combined and I'm not quite sure which one is more important to us

disteph:
> Yices2 has some additional features that we may want to include:
> 
>     * Nonlinear arithmetic is supported with the MCSat solver
> 
>     * MCSat also support interpolation ([link](https://yices.csl.sri.com/doc/context-operations.html#c.yices_check_context_with_interpolation)), but not `unsat core` ([yet?](https://github.com/SRI-CSL/yices2/pull/578))

Just FYI, support for unsat cores in MCSAT has been merged into master (both through SMTLib2 input and through API calls): [https://github.com/SRI-CSL/yices2/pull/578](https://github.com/SRI-CSL/yices2/pull/578).

It will be part of the upcoming Yices release 2.8.

baierd:
Thank you for that info! Both features are actually requested by users that i talk to a lot currently!
@daniel-raffler we should focus on finishing this PR. It will most likely not make 6.0.0, but we can just publish 6.1 once this is done.

daniel-raffler:
The branch has now been switched to MCSAT, and I'll add  interpolation support in the next few days. Unfortunately the `--enable-thread-safety` and `--enable-mcsat` options have to be set at build time and can't be combined. This means the new version will not be thread safe, and only a single thread may use Yices at a time

We may be able to ship two binaries, one for MCSAT and one that allows parallel contexts, similar to how the issue is handled in MathSAT/OptiMathSAT. At run-time users can then choose which version to use when opening their `SolverContext`

If this turns out to be too complex, we may also consider adding a global lock for Yices, and simply forbidding all access that does not come from the original thread

daniel-raffler:
Some update:
* We can now use `--enable-mcsat` and `--enable-thread-safety` in the same build and will only need one binary (See https://github.com/SRI-CSL/yices2/pull/607)
* There is another PR ([link](https://github.com/SRI-CSL/yices2/pull/606)) that could be interesting to us, so we should probably wait for the upcoming Yices 2.8.0 release before merging this
* Interpolation is working now
* This branch now uses the official Yices Java bindings ([here](https://github.com/SRI-CSL/yices2_java_bindings)). Thanks to @baierd for the hint
* The Java bindings are somewhat outdated. I've forked the repository [here](https://github.com/daniel-raffler/yices2_java_bindings) and added a few patches. Hopefully, this can be merged upstream once we're done
* We still need a new build script

daniel-raffler:
Some benchmark results:

Screenshot From 2026-03-18 09-56-02

From left to right: Yices old, Yices new (with dpllt), MathSAT (for reference)

Comparison between `dpllt` and `mcsat`:

Screenshot From 2026-03-18 09-55-37

We get significantly better results with `dpllt` (light brown), then with `mcsat`. Part of this is due to some open issues that cause segfaults in `mcsat`. However, this can't explain the difference entirely, and `dpllt` does currently seem to outperform `mcsat` at least on some tasks. (Note that interpolation always has to be done with `mcsat`, even when the `dpllt` option is used. So the comparison isn't entirely fair, and we would get much worse results for `dpllt` if it wasn't for this fallback)

daniel-raffler:
I've added binaries for Windows. They seem to be working in Wine, but we may need some more testing

Yices has github actions ([link](https://github.com/SRI-CSL/yices2/blob/master/.github/actions/build/action.yml)) to build the solver on MacOS. We could probably start from there if we also want to support that platform

daniel-raffler:
@kfriedberger Just a heads up: If you want to build the binaries, we need the [issue-613-mcsat-curried-function-model](https://github.com/SRI-CSL/yices2/tree/issue-613-mcsat-curried-function-model) branch

You can download yices like this:
```
ant download-yices2
```
Then go to `downloads/yices2` and change the branch:
```
cd downloads/yices2
git checkout issue-613-mcsat-curried-function-model
```
After that, just follow the build instructions

kfriedberger:
Is there a plan to integrate https://github.com/daniel-raffler/yices2_java_bindings.git into its original? Just asking, not urgent.

daniel-raffler:
> Is there a plan to integrate https://github.com/daniel-raffler/yices2_java_bindings.git into its original? Just asking, not urgent.

Yes, I've open a PR last week: https://github.com/SRI-CSL/yices2_java_bindings/pull/8

ThomasHaas:
> Yices has github actions ([link](https://github.com/SRI-CSL/yices2/blob/master/.github/actions/build/action.yml)) to build the solver on MacOS. We could probably start from there if we also want to support that platform

I would love Yices2 support on MacOS (ARM). It usually outperforms Z3 (and others) by quite a lot on our QF_BV encoding in Dartagnan (sometimes by a whole order of magnitude). It seems to do some serious black magic there :)

daniel-raffler:
Hello Thomas,

I've been able to build the MacOS (Arm) binaries with a github runner. We still have to figure out to sign and ship the binaries, but I can at least include the build script. It's not too complicated  to run and the binaries can then just be copied over

> > Yices has github actions ([link](https://github.com/SRI-CSL/yices2/blob/master/.github/actions/build/action.yml)) to build the solver on MacOS. We could probably start from there if we also want to support that platform
> 
> I would love Yices2 support on MacOS (ARM). It usually outperforms Z3 (and others) by quite a lot on our QF_BV encoding in Dartagnan (sometimes by a whole order of magnitude). It seems to do some serious black magic there :)

Is this with the default solver, or would we need one of [these](https://web.archive.org/web/20251212142346/https://yices.csl.sri.com/doc/install-sources.html#third-party-sat-solvers) delegates? So far we only have the internal solver, but we might still add one of the delegates

ThomasHaas:
> I've been able to build the MacOS (Arm) binaries with a github runner. We still have to figure out to sign and ship the binaries, but I can at least include the build script. It's not too complicated to run and the binaries can then just be copied over

Sounds great.

> > I would love Yices2 support on MacOS (ARM). It usually outperforms Z3 (and others) by quite a lot on our QF_BV encoding in Dartagnan (sometimes by a whole order of magnitude). It seems to do some serious black magic there :)
> 
> Is this with the default solver, or would we need one of [these](https://web.archive.org/web/20251212142346/https://yices.csl.sri.com/doc/install-sources.html#third-party-sat-solvers) delegates? So far we only have the internal solver, but we might still add one of the delegates

The default internal solver. Funnily enough, that one outpeforms Kissat and Cadical on our benchmarks.
To test that, we generated QF_BV encodings, used Yices2 to bitblast them into DIMACS and then re-parsed the file with Yices2 but different SAT solver backends (internal/kissat/cadical). The internal SAT solver was the fastest!
Also, Kissat does not support incremental solving, so it's not so interesting in the first place I think.
Similarly, MCSAT is extremely bad on bitvector arithmetic and I think multi-threaded solving is also quite bad...
Basically, we only care about the default setting for now :)

kfriedberger:
Thanks for the input, everyone. A couple of thoughts on the direction of the latest discussion here:

First, I think it's important that we keep JavaSMT vendor-independent when it comes to our build process. We should avoid being tied too closely to GitHub Actions for publishing; the goal is to ensure any user can still build the library reliably on their own local machine. Apple is kind of special here, maybe we find a way with tools like `osxcross/ldid/rcodesign` in the future.

Regarding the macOS-specific points, there are multiple solvers waiting for macOS:
- Yices2 (maybe outdated PR): #400
- MathSAT (integrated, but maybe only used once): #430

Should we move that conversation over to a separate issue? It feels like it deserves its own space for discussion and debugging without cluttering the Yices update.

kfriedberger:
@daniel-raffler :
The current state looks good. 👍
Do you see anything blocking a merge? If not, let's merge. ⏩ 
Further updates can be applied in another PR, e.g., after merging https://github.com/SRI-CSL/yices2_java_bindings/pull/8 and the "final" release of Yices 2.8.0. Version 2.7.0+X is sufficient for today.
You have done great work in updating these solver bindings. 🥇

kfriedberger (dismissed):

kfriedberger (approved):

daniel-raffler:
> Do you see anything blocking a merge? If not, let's merge. ⏩ Further updates can be applied in another PR, e.g., after merging [SRI-CSL/yices2_java_bindings#8](https://github.com/SRI-CSL/yices2_java_bindings/pull/8) and the "final" release of Yices 2.8.0. Version 2.7.0+X is sufficient for today. 

No, nothing major. I still have a few features on my todo list, but these can be added later

> You have done great work in updating these solver bindings. 🥇

Thanks! 😊

> First, I think it's important that we keep JavaSMT vendor-independent when it comes to our build process. We should avoid being tied too closely to GitHub Actions for publishing; the goal is to ensure any user can still build the library reliably on their own local machine. Apple is kind of special here, maybe we find a way with tools like osxcross/ldid/rcodesign in the future.

Agreed. Using github actions for publishing has some problems and would at least complicate the process for us. However, what I think we could do, is at least provide some additional build targets for the binaries on macOS. Users could then simply compile their own binaries from source without first having to work their way through our build system. We would then only rely on github actions for testing our build scripts

> Should we move that conversation over to a separate issue? It feels like it deserves its own space for discussion and debugging without cluttering the Yices update.

Yes, this affects a few solvers, so we could just open a new issue for it 👍

disteph:
> > Is there a plan to integrate https://github.com/daniel-raffler/yices2_java_bindings.git into its original? Just asking, not urgent.
> 
> Yes, I've open a PR last week: [SRI-CSL/yices2_java_bindings#8](https://github.com/SRI-CSL/yices2_java_bindings/pull/8)

Your PR is merged! I've also aligned the Windows JNI with it (with a test harness including windows).

disteph:
News: SAT-solver delegates are now available in Yices contexts, including support for push/pop behavior, but still limited to QF_BV logic:
https://github.com/SRI-CSL/yices2/pull/607

These other issues have been fixed and merged into master:
https://github.com/SRI-CSL/yices2/issues/608
https://github.com/SRI-CSL/yices2/issues/613
https://github.com/SRI-CSL/yices2/issues/615

The last one remaining in the issues listed in this thread is 
https://github.com/SRI-CSL/yices2/issues/616
I can't reproduce, but I have a tentative fix/hardening at https://github.com/SRI-CSL/yices2/pull/628

daniel-raffler:
> News: SAT-solver delegates are now available in Yices contexts, including support for push/pop behavior, but still limited to QF_BV logic: [SRI-CSL/yices2#607](https://github.com/SRI-CSL/yices2/pull/607)
> 
> These other issues have been fixed and merged into master: [SRI-CSL/yices2#608](https://github.com/SRI-CSL/yices2/issues/608) [SRI-CSL/yices2#613](https://github.com/SRI-CSL/yices2/issues/613) [SRI-CSL/yices2#615](https://github.com/SRI-CSL/yices2/issues/615)
> 
> The last one remaining in the issues listed in this thread is [SRI-CSL/yices2#616](https://github.com/SRI-CSL/yices2/issues/616) I can't reproduce, but I have a tentative fix/hardening at [SRI-CSL/yices2#628](https://github.com/SRI-CSL/yices2/pull/628)

Thanks for the update! I ran some benchmarks and the segfaults during interpolation are completely gone now. We still get exceptions for these tasks, but at least that's something we can react to, and maybe try a different strategy. Note there are still a few unrelated crashes that don't appear to be fixed by https://github.com/SRI-CSL/yices2/pull/628 either. However, the issue is incredibly rare (< 4 in 18000 tasks), so it might be fine to just ignore it for now

Good work!

disteph:
> > News: SAT-solver delegates are now available in Yices contexts, including support for push/pop behavior, but still limited to QF_BV logic: [SRI-CSL/yices2#607](https://github.com/SRI-CSL/yices2/pull/607)
> > These other issues have been fixed and merged into master: [SRI-CSL/yices2#608](https://github.com/SRI-CSL/yices2/issues/608) [SRI-CSL/yices2#613](https://github.com/SRI-CSL/yices2/issues/613) [SRI-CSL/yices2#615](https://github.com/SRI-CSL/yices2/issues/615)
> > The last one remaining in the issues listed in this thread is [SRI-CSL/yices2#616](https://github.com/SRI-CSL/yices2/issues/616) I can't reproduce, but I have a tentative fix/hardening at [SRI-CSL/yices2#628](https://github.com/SRI-CSL/yices2/pull/628)
> 
> Thanks for the update! I ran some benchmarks and the segfaults during interpolation are completely gone now. We still get exceptions for these tasks, but at least that's something we can react to, and maybe try a different strategy. Note there are still a few unrelated crashes that don't appear to be fixed by [SRI-CSL/yices2#628](https://github.com/SRI-CSL/yices2/pull/628) either. However, the issue is incredibly rare (< 4 in 18000 tasks), so it might be fine to just ignore it for now
> 
> Good work!

Now merged into master. I've pushed another commit fixing another data race; maybe it addresses the remaining 4 tasks. Feel free to rerun your tests on master.

PR #528, opened 2025-09-24
Use the partial model for formula evaluation in Princess

daniel-raffler:
Hello,
We stopped using the partial model in Princess when we added support for Rationals and Strings. This was needed due to several issues with evaluation. Our current work-around is to simply push the formula that should be evaluated to the assertion stack, re-run the SAT check, and then report the values from the model. While this does work it adds additional overhead and makes our code harder to maintain

The evaluation issues in Princess have since been fixed, and this PR will therefore remove the current work-around and return to using the Princess partial model for evaluation

Currently there are still 2 failing tests that have been temporarily disabled in the last commits:
- `StringFormulaManagerTest.testOutputUnescape` The partial model in Princess currently does not support evaluating constant String formulas. This issue was reported as https://github.com/uuverifiers/ostrich/issues/108 and has already been fixed upstream. We could wait for the next release, or just leave the test disabled for now.
- `ModelTest.testGetArrays3` This is a separate issue with our own backend that will be addressed in https://github.com/sosy-lab/java-smt/issues/520. We can just leave the test disabled for now and turn it back on after the fix

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/princess/PrincessAbstractProver.java
@@ -23,9 +23,7 @@ import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Deque;
-import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


kfriedberger:
Issues still not solved with Princess 2025-11-17 and Ostrich 2.0. We still need to wait for an updated version.

UPDATE: The update of ostrich will be done with #565. Lets close this issue here.

PR #535, opened 2025-10-15
Fix Z3 bound variable substitution

baierd:
Fix for the regression found by Philipp in CPAchecker. Z3 instantiates its bound variables reversed to the initial list of bound variables when building a quantified formula. We used them in the same order.

I also added some tests for this. (that should be extended at some point, but that can be done later as well)


baierd:
The Z3 issue with the solution provided by NikolajBjorner can be found here: https://github.com/Z3Prover/z3/issues/7970

kfriedberger (approved):
lgtm.
The link to the Z3 issue is good.

PR #536, opened 2025-10-17
Re-enable Z3 4.5.0 for better interpolation

leventeBajczi:
Old versions of z3 were really good at interpolation, and I want to add support for it in JavaSMT.

To this end, I've added the implementation of a `Z3Legacy` solver (to be renamed, as discussed with @baierd), which builds on the 4.5.0 release of z3, rebuilt to use `com.microsoft.z3legacy` as its package name to avoid name collisions with new z3. 

Pre-built binaries with the modifications are available at https://github.com/leventeBajczi/z3/releases/tag/legacy-4.3.0-15644597379, which [Theta](https://github.com/ftsrg/theta/) already uses like this. This should be somehow integrated into the solver management of JavaSMT, which I have no experience with.

To demonstrate how good this version of Z3 is, I ran some preliminary tests in CPAchecker, and found that an NLA encoding of programs using Z3 as the interpolation solver in the predicate analysis configuration results in 510 successfully solved tasks, which are otherwise unsolved by the bitprecise encoding of the same tasks with mathsat5. 

Also, I found that  https://github.com/sosy-lab/java-smt/issues/381 (which I opened) is referenced in 3 of the tests, that now failed on this implementation. I removed the overly constraining line which expected the constraint IDs to be unique, which is not necessary IMO if the interpolation problem reported in that issue is otherwise handled - which it seems to be.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
@@ -50,9 +50,9 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
   private final ShutdownRequestListener interruptListener;
 
   private final UniqueIdGenerator trackId = new UniqueIdGenerator();
-  @Nullable private final Deque<PersistentMap<String, BooleanFormula>> storedConstraints;
+  private final Optional<Deque<PersistentMap<String, BooleanFormula>>> storedConstraints;
 
-  private final @Nullable PathCounterTemplate logfile;
+  private final Optional<PathCounterTemplate> logfile;
 
   Z3LegacyAbstractProver(
       Z3LegacyFormulaCreator pCreator,
@@ -71,13 +71,13 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
     shutdownNotifier.register(interruptListener);
 
     if (pOptions.contains(ProverOptions.GENERATE_UNSAT_CORE)) {
-      storedConstraints = new ArrayDeque<>();
-      storedConstraints.push(PathCopyingPersistentTreeMap.of());
+      storedConstraints = Optional.of(new ArrayDeque<>());
+      storedConstraints.get().push(PathCopyingPersistentTreeMap.of());
     } else {
-      storedConstraints = null; // we use NULL as flag for "no unsat-core"
+      storedConstraints = Optional.empty();
     }
 
-    logfile = pLogfile;
+    logfile = Optional.ofNullable(pLogfile);
     mgr = pMgr;
   }
 
@@ -102,10 +102,10 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
 
   /** dump the current solver stack into a new SMTLIB file. */
   protected void logSolverStack() throws SolverException {
-    if (logfile != null) { // if logging is not disabled
+    if (logfile.isPresent()) { // if logging is not disabled
       try {
         // write stack content to logfile
-        Path filename = logfile.getFreshPath();
+        Path filename = logfile.get().getFreshPath();
         MoreFiles.createParentDirectories(filename);
         Files.writeString(filename, this + "(check-sat)\n");
       } catch (IOException e) {
@@ -165,11 +165,11 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
     Preconditions.checkState(!closed);
     long e = creator.extractInfo(f);
     try {
-      if (storedConstraints != null) { // Unsat core generation is on.
+      if (storedConstraints.isPresent()) { // Unsat core generation is on.
         String varName = String.format("Z3_UNSAT_CORE_%d", trackId.getFreshId());
         BooleanFormula t = mgr.getBooleanFormulaManager().makeVariable(varName);
         assertContraintAndTrack(e, creator.extractInfo(t));
-        storedConstraints.push(storedConstraints.pop().putAndCopy(varName, f));
+        storedConstraints.get().push(storedConstraints.get().pop().putAndCopy(varName, f));
       } else {
         assertContraint(e);
       }
@@ -181,16 +181,12 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
 
   protected void push0() {
     Preconditions.checkState(!closed);
-    if (storedConstraints != null) {
-      storedConstraints.push(storedConstraints.peek());
-    }
+    storedConstraints.ifPresent(pPersistentMaps -> pPersistentMaps.push(pPersistentMaps.peek()));
   }
 
   protected void pop0() {
     Preconditions.checkState(!closed);
-    if (storedConstraints != null) {
-      storedConstraints.pop();
-    }
+    storedConstraints.ifPresent(Deque::pop);
   }
 
   @Override
@@ -282,7 +278,7 @@ abstract class Z3LegacyAbstractProver<T> extends AbstractProverWithAllSat<T> {
   public List<BooleanFormula> getUnsatCore() {
     Preconditions.checkState(!closed);
     checkGenerateUnsatCores();
-    if (storedConstraints == null) {
+    if (storedConstraints.isEmpty()) {
       throw new UnsupportedOperationException(
           "Option to generate the UNSAT core wasn't enabled when creating the prover environment.");
     }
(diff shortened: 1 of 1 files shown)


baierd:
Thank you very much for the PR! It seems like the old Z3 version is beneficial to have. Could you also post the tables and/or quantile-plots from your experiments (for documentation purposes)?

We need to make sure that there are no issues when running both Z3 instances at the same time! I'll try to start some tests in CPAchecker and report back.

leventeBajczi:
Sure! Here are some:
[z3-vs-baseline.table.html](https://github.com/user-attachments/files/23231677/z3-vs-baseline.table.html)
In particular, the `#/table?filter=0(0*status*(status(notIn()),category(in(correct)))),1(0*status*(status(notIn()),category(in(error))))` filter shows the real advantage of using Z3 with nonlinear integers complementing the best bitvector config (can solve 510 unsolved tasks).

baierd:
Thank you very much!

Btw. could you try using Z3 version 4.5 from our Ivy repo?
Maybe the old lib still works?

leventeBajczi:
I haven't used it, but I'd be cautious about using it. The libraries I built use the `com.microsoft.z3legacy` package, as to not mistakenly use a mismatched java/native library.

baierd:
> I haven't used it, but I'd be cautious about using it. The libraries I built use the `com.microsoft.z3legacy` package, as to not mistakenly use a mismatched java/native library.

Good to know! So we need to compile and release it again. Thanks for the clarification!

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
```diff
@@ -0,0 +1,396 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2024 Dirk Beyer <https://www.sosy-lab.org>
```
baierd: 2025 :wink:

src/org/sosy_lab/java_smt/SolverContextFactory.java
```diff
@@ -55,6 +56,7 @@ public enum Solvers {
     MATHSAT5,
     SMTINTERPOL,
     Z3,
+    Z3LEGACY,
```
baierd: @kfriedberger @PhilippWendler @leventeBajczi  i kinda feel like Z3 legacy is a little unspecific. Since the version is fixed, should we use the actual version? I.e. `Z3-3.5`?

leventeBajczi:
Well, I'd suggest just using the [ones I linked to](https://github.com/leventeBajczi/z3/releases/tag/legacy-4.3.0-15644597379), but sure, I can also provide a patch on top of the 4.5.0 Z3 so that you can rebuild it if you'd prefer.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
```diff
@@ -0,0 +1,396 @@
...
+    Native.solverIncRef(z3context, z3solver);
+
+    interruptListener = reason -> Native.interrupt(z3context);
+    shutdownNotifier.register(interruptListener);
+
+    if (pOptions.contains(ProverOptions.GENERATE_UNSAT_CORE)) {
+      storedConstraints = new ArrayDeque<>();
+      storedConstraints.push(PathCopyingPersistentTreeMap.of());
+    } else {
+      storedConstraints = null; // we use NULL as flag for "no unsat-core"
```
baierd: Optional would reduce the risk of accidental null usage.

baierd: Using `Optional` would reduce the chance of accidental `null` access.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyFormulaCreator.java
```diff
@@ -0,0 +1,1124 @@
...
+            args.build(),
+            FunctionDeclarationImpl.of(
+                getAppName(f),
+                getDeclarationKind(f),
+                argTypes.build(),
+                getFormulaType(f),
+                Native.getAppDecl(environment, f)));
+      case Z3_VAR_AST:
+        int deBruijnIdx = Native.getIndexValue(environment, f);
+        return visitor.visitBoundVariable(formula, deBruijnIdx);
```
baierd, line 566: Please note that `visitBoundVariable()` has recently been deprecated.
`visitQuantifier()` now returns the free variable equivalents for all bound variables and the body with all the bound variables re-substituted.

baierd (changes requested):
I can't find more issues than some small suggestions and quantifier visitation being outdated. Good job and thank you for all your work!

I suggest to import this into a branch in JavaSMT first, then add the binaries to Ivy, run a more complex test-setup in CPAchecker before merging.

src/org/sosy_lab/java_smt/SolverContextFactory.java
kfriedberger: The solver is able to provide its version string anyway. As it is a version of Z3 with interpolation support, we might name it `Z3_INTERPOLATION` or simply `Z3_3_5` . Special chars like dots or hyphens are not allowed in Java enums.

The JavaSMT enum it independent of the native library, so the library can be called `z3Legacy` or similar.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
kfriedberger: We could go through the whole code once and update to latest coding conventions, and include latest bugfixes from Z3.

kfriedberger: ... i mean, including latest bugfixes for JavaSMT only, no bugfixes for an outdated Z3. :smile:

baierd: I agree. This should include Z3 as well, since this for example is still present in our Z3. But this should be done independently of Levis work and help, as he/we want this for SVCOMP this year actually ;D

baierd:
> Well, I'd suggest just using the [ones I linked to](https://github.com/leventeBajczi/z3/releases/tag/legacy-4.3.0-15644597379), but sure, I can also provide a patch on top of the 4.5.0 Z3 so that you can rebuild it if you'd prefer.

We want to use a uniform build-process and build-system, as the dependencies used when building may depend on the system used. Also, the scripts do everything more or less automatically anyway, including the upload. @kfriedberger knows more about the publishing, but it might be more work to re-use your binaries compared to executing a command.

You have to remember that even for a fixed Z3 version, the libraries behind it might be updated from time to time!

leventeBajczi:
Thanks for the reviews and the answers! I'll get to fixing the code soon, but in the meantime, I prepared a patch:
[z3legacy.patch](https://github.com/user-attachments/files/23283326/z3legacy.patch)
This can be applied to [the 4.5.0 tag](https://github.com/Z3Prover/z3/tree/d57a2a6dce9291acf9c71a561252f3e133f0c894) and then built with 
```
mkdir release
python2.7 scripts/mk_make.py --prefix=$PWD/release --java
cd build/ && make -j $(nproc) && make install
```
(i.e., the usual build command back in 4.5.0 times :sweat_smile: )

Besides the package rename, this also fixes some weird issue in the `hwf.cpp` file. I just took a later version's equivalent code, and for us, this has been working for the past ~1 year, so I think it's safe to say it didn't break something trivial, and fixed that weird behavior.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
leventeBajczi: Nice catch, fixed!

src/org/sosy_lab/java_smt/SolverContextFactory.java
leventeBajczi: I like encoding the version (`Z3_INTERPOLATION` could I think be confused with `iZ3`), so I changed it to `Z3_4_5_0`.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyAbstractProver.java
leventeBajczi: I fixed it in this file; but this came directly from the Z3 implementation, and I did not touch that now. I agree this should probably be fixed globally in the project.

src/org/sosy_lab/java_smt/solvers/z3legacy/Z3LegacyFormulaCreator.java
leventeBajczi: I think this should probably be fixed together with the up-to-date `z3` formula creator, and I'd prefer to leave this part as close in the two implementations as possible. LMK if you don't agree!

PR #539, opened 2025-10-28
Fully Evaluate Nested Arrays for Model Generation in CVC5

baierd:
This MR adds CVC5 multi-dimensional array model evaluation based on @daniel-raffler s implementation in [commit 0d463c83410a332ed81ff4f7628d25ae729b3782](https://github.com/sosy-lab/java-smt/commit/0d463c83410a332ed81ff4f7628d25ae729b3782#diff-e5607c039d73f792b5719608419f6f0b09ccba3200598fbc492cc20b377d42c3), with minor modifications.

It also adds 2 tests for multi-dimensional arrays not based on SMTLIB2 parsing (we did not have any) as well.

PR #540, opened 2025-10-28
Add CVC5 Parsing Support

baierd:
This PR adds support for parsing to CVC5.

Should only be merged after [PR539](https://github.com/sosy-lab/java-smt/pull/539).

### DIFF
--- a/src/org/sosy_lab/java_smt/test/ModelTest.java
+++ b/src/org/sosy_lab/java_smt/test/ModelTest.java
@@ -2485,16 +2485,16 @@ public class ModelTest extends SolverBasedTest0.ParameterizedSolverBasedTest0 {
     requireArrays();
     requireBitvectors();
 
-    assume()
-        .withMessage("Solver is quite slow for this example")
-        .that(solverToUse())
-        .isNoneOf(Solvers.MATHSAT5, Solvers.PRINCESS, Solvers.BITWUZLA);
-
     BooleanFormula formula =
         context
             .getFormulaManager()
             .parse(
                 Files.readString(Path.of("src/org/sosy_lab/java_smt/test/SMT2_UF_and_Array.smt2")));
+
+    assume()
+        .withMessage("Solver is quite slow for this example")
+        .that(solverToUse())
+        .isNoneOf(Solvers.CVC5, Solvers.MATHSAT5, Solvers.PRINCESS, Solvers.BITWUZLA);
 
     checkModelIteration(formula, false);
   }


baierd:
The last remaining problem is that MathSAT5 output can't be parsed in general, as MathSAT5 splits up the query and uses a `.` in the beginning of its intermediate variables, e.g.:

```
(set-info :source |printed by MathSAT|)
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(declare-fun q () Bool)
(declare-fun u () Bool)
(define-fun .def_15 () Int (* (- 1) c))
(define-fun .def_16 () Int (+ b .def_15))
(define-fun .def_17 () Int (+ a .def_16))
(define-fun .def_19 () Bool (= .def_17 0))
(define-fun .def_27 () Bool (= .def_19 q))
(define-fun .def_28 () Bool (not .def_27))
(define-fun .def_23 () Bool (<= b a))
(define-fun .def_29 () Bool (and .def_23 .def_28))
(define-fun .def_11 () Bool (= a b))
(define-fun .def_34 () Bool (and .def_11 .def_29))
(define-fun .def_30 () Bool (or u .def_29))
(define-fun .def_31 () Bool (and q .def_30))
(define-fun .def_35 () Bool (and .def_31 .def_34))
(assert .def_35)
```

We should ask the CVC5 devs to allow this, and or add a re-substitution procedure.

baierd:
> The last remaining problem is that MathSAT5 output can't be parsed in general, as MathSAT5 splits up the query and uses a `.` in the beginning of its intermediate variables, e.g.:
> 
> ```
> (set-info :source |printed by MathSAT|)
> (declare-fun a () Int)
> (declare-fun b () Int)
> (declare-fun c () Int)
> (declare-fun q () Bool)
> (declare-fun u () Bool)
> (define-fun .def_15 () Int (* (- 1) c))
> (define-fun .def_16 () Int (+ b .def_15))
> (define-fun .def_17 () Int (+ a .def_16))
> (define-fun .def_19 () Bool (= .def_17 0))
> (define-fun .def_27 () Bool (= .def_19 q))
> (define-fun .def_28 () Bool (not .def_27))
> (define-fun .def_23 () Bool (<= b a))
> (define-fun .def_29 () Bool (and .def_23 .def_28))
> (define-fun .def_11 () Bool (= a b))
> (define-fun .def_34 () Bool (and .def_11 .def_29))
> (define-fun .def_30 () Bool (or u .def_29))
> (define-fun .def_31 () Bool (and q .def_30))
> (define-fun .def_35 () Bool (and .def_31 .def_34))
> (assert .def_35)
> ```
> 
> We should ask the CVC5 devs to allow this, and or add a re-substitution procedure.

I've now added a re-substitution for MathSAT5s output. Still, we should look into either modifying the output of MathSAT, ask the MathSAT devs to change this, use the alternative dump function of MathSAT, or ask the CVC5 devs to allow this.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaCreator.java
```diff
@@ -427,7 +441,7 @@ public <R> R visit(FormulaVisitor<R> visitor, Formula formula, final Term f) {
         List<Formula> freeVars = new ArrayList<>();
         for (Term boundVar : f.getChild(0)) { // unpack grand-children of f.
           String name = getName(boundVar);
-          Term freeVar = Preconditions.checkNotNull(accessVariablesCache(name, boundVar.getSort()));
+          Term freeVar = getFreeVariableFromCache(name, boundVar.getSort(), visitor);
```
daniel-raffler, line 448: Can't we just use `makeVariable` here? Then we don't have to use `registerBoundVariablesWithVisitor` later

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaCreator.java
```diff
@@ -417,7 +431,7 @@ public <R> R visit(FormulaVisitor<R> visitor, Formula formula, final Term f) {
         // BOUND vars are used for all vars that are bound to a quantifier in CVC5.
         // We resubstitute them back to the original free.
         // CVC5 doesn't give you the de-brujin index
-        Term originalVar = accessVariablesCache(formula.toString(), sort);
+        Term originalVar = getFreeVariableFromCache(formula.toString(), sort, visitor);
```
daniel-raffler, line 438: We could just remove the entire case as it is dead code. Bound variables are replaced in the visitor before we ever get here

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+    if (!parseSolver.isLogicSet()) {
+      try {
+        parseSolver.setLogic("ALL");
+      } catch (CVC5ApiException e) {
+        // Should never happen in this configuration
+        throw new AssertionError("Unexpected exception", e);
+      }
+    }
+    // Expected success string: "success\n"
+    parseSolver.setOption("print-success", "true");
```
daniel-raffler: Won't this print to `stdout`?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+    Set<String> declarationsForAllVarsAndUfs =
+        getSMTLIB2DeclarationsFor(
+            ImmutableMap.<String, Term>builder()
+                .putAll(creator.getAllCachedVariablesAndUFs(true, true))
+                .buildOrThrow());
+
+    StringBuilder extraDefs = new StringBuilder();
+    for (String declaration : declarationsForAllVarsAndUfs) {
+      String declWoNewline = declaration.replace("\n", "");
+      if (!formulaStr.contains(declWoNewline)) {
```
daniel-raffler: This may add a duplicate declaration if the SMTLIB input doesn't match the generated declaration exactly, f.ex if there is additional whitespace or if the name has unnecessary quotes

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+    // (double parsing is a problem for UFs it seems, and the internal UFs cause problems)
+    Set<String> declarationsForAllVarsAndUfs =
+        getSMTLIB2DeclarationsFor(
+            ImmutableMap.<String, Term>builder()
+                .putAll(creator.getAllCachedVariablesAndUFs(true, true))
+                .buildOrThrow());
+
+    StringBuilder extraDefs = new StringBuilder();
+    for (String declaration : declarationsForAllVarsAndUfs) {
+      String declWoNewline = declaration.replace("\n", "");
```
daniel-raffler: Linebreaks can also occur in quoted names

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+    }
+    return extraDefs.append(formulaStr).toString();
+  }
+
+  private String sanitizeInputForMathsat(String inputString) {
+    // MathSAT5 output, resubstitute the illegal input.
+    // Since no quotes are used, we can assume that there are no spaces in the defined name.
+    ImmutableMap.Builder<String, String> replacementMapBuilder = ImmutableMap.builder();
+    String sanitizedInputFormula = inputString;
+    String pattern = "(define-fun \\.)[^\\s]+";
```
daniel-raffler: What about `(define-fun |.def| ...)` ? The symbol should still be illegal when there are quotes around it. The regexp will also match parts of the input that are inside a String literal (`"(define-const .abc)"`) or SMTLIB quotes (`|(define-const .abc)|`)

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+
+    // Add all already known (cached) variables and UFs
+    // (double parsing is a problem for UFs it seems, and the internal UFs cause problems)
+    String sanitizedInputFormula = addCachedDeclarationsTo(formulaStr);
+    boolean mathsatReplacement = false;
+    // Sanitize input String for CVC5 (it disallows . and @ without quotes)
+    if (sanitizedInputFormula.contains("define-fun .def_")) {
+      mathsatReplacement = true;
+      sanitizedInputFormula = sanitizeInputForMathsat(sanitizedInputFormula);
+    }
```
daniel-raffler: I think this is too error prone and should be handled somewhere else. Either we ask the CVC5 developers to allow identifiers that start with `.` or we ask the MathSAT team to change their output. The SMTLIB standard is quite vague here, so I'm nor sure which is the better option:
```
Members of the ⟨symbol ⟩ category starting with the character @ or . are reserved for solver
use. Solvers can use them respectively as identifiers for abstract values and solver generated
function symbols other than abstract values.
```
The easiest solution for now may be to just fix the tests and remove the `.` from there

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+      }
+
+      command = getNextCommand(parser);
+    }
+
+    // Register new terms in our caches or throw for errors
+    registerNewTermsInCache(sm, substituteFromBuilder, substituteToBuilder);
+
+    // TODO: Which terms can end up here? Seems like this is always empty
+    checkState(sm.getNamedTerms().isEmpty());
```
daniel-raffler: This is for named terms, f.ex `p1` and `p2` in `(=> (! (> x y) : named p1) (! (= x z) : named p2))`. We can probably just ignore it and don't need the exception.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+
+    // If the symbols used in the term were already declared before parsing, the term uses new
+    // ones with the same name, so we need to substitute them!
+    parsedTerm =
+        resubstituteCachedVariables(
+            substituteFromBuilder.build(), substituteToBuilder.build(), parsedTerm);
+
+    // Quantified formulas do not give us the bound variables in getDeclaredTerms() above.
+    // Find them and register a free equivalent
+    creator.registerBoundVariablesWithVisitor(parsedTerm);
```
daniel-raffler: This shouldn't be needed if we just create the variables in the visitor (see above)

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +83,269 @@ static Term getCVC5Term(Formula pT) {
...
+    ImmutableSet.Builder<Term> substituteToBuilder = ImmutableSet.builder();
+
+    int numberOfAssertions = 0;
+    Command command = getNextCommand(parser);
+    while (!command.isNull()) {
+      if (command.toString().contains("push") || command.toString().contains("pop")) {
+        // TODO: push and pop?
+        throw new IllegalArgumentException(
+            "Parsing SMTLIB2 with CVC5 in JavaSMT does not support push or pop currently.");
+      }
```
daniel-raffler: We already throw an exception if the input contains `pop` or `push` in [AbstractFormulaManager.parse](https://github.com/sosy-lab/java-smt/blob/7d89b87329ba34e955b0ca95a91816128ba9ae26/src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java#L329)

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: True! I completely forgot. Thanks for the reminder! fixed.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Ignoring something without a full understanding as to why it could happen is error-prone. Hence the exception is fine, since we don't handle the cases.

My TODO is however worded very poorly. I improved it with your help!

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: True, lets get rid of this check then! Thanks.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: SMTLIB2 does not quote its internal commands/symbols etc.
Hence why the quotes in `(define-fun |.def| ...)` are OK.

However, `|(define-const .abc)|` would break this, yes! Very good catch.
Quotes are rather hard to handle though, maybe we just disallow the combination for now?

But to be fair, i don't like this solution either. Its slow and error-prone. As written above, i'll try to contact the devs and sort this out.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Or mismatched types... all good points!

We should extract the name and type properly, including and respecting possible quotes and other special handling or symbols, and check for the symbol and the symbol + type combination.

Do we have some code for this somewhere?

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: As far as i understood it no. It just prints the SMTLIB2 success message for parsing a command as a response to parsing it in . I think what you remember is the last arguments of `setStringInput()`, which can set a stream for error messages, but we seem to get all of them through the command invoking as Strings or exceptions (at least in all tests until now). `invoke()`, can set the output stream in other languages, but just returns Strings in Java. The later is what returns the success message we activate with the marked option.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: I fully agree.

I'll simply block combinations that break the current solution for now.
Maybe [this](https://github.com/sosy-lab/java-smt/pull/540#discussion_r2482561583) could also help here.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Actually, my proposal is also not a full solution. We really would need to extract all existing symbols from the input. But we can actually do that rather safely by simply using the parser and start to read in all symbols and extract their names (and types?), then failing once we are missing symbols and re-starting with the symbol adding process.

We would need to test how the parser reacts to let-bindings, but we could try.

daniel-raffler: > We should extract the name and type properly, including and respecting possible quotes and other special handling or symbols, and check for the symbol and the symbol + type combination.
> 
> Do we have some code for this somewhere?

I think it's difficult to get this right without effectively writing our own parsing code. In https://github.com/sosy-lab/java-smt/pull/474 I used a separate run to filter out duplicate definitions from the input and check that there is only one type for every variable. The code can be found [here](https://github.com/sosy-lab/java-smt/pull/474) and could (mostly) be copied over

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
daniel-raffler: Ah, so this is about getting `"sucess"` as return String when calling `invoke`? We can keep it then, but I think CVC5 would throw an exception anyway if there is a parse error

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaCreator.java
baierd: It is now removed.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: According to the documentation is can also return 2 errors here, but in general you are correct.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Turns out the solution is really easy, or as a wise person once said: hours of aimless programming can save you minutes of reading the documentation.

I enabled `lenient` parsing and all problems in regard to disallowed input (MathSAT5 input with the dot in front of the symbol) are now solved.
Also, re-using the SymbolManager and some other options allow it to find most existing symbols or already parsed symbols. There are some small exceptions (mainly if only used in 1 assert in the end without any other input) that needs special handling, but there we can get the EXACT symbol name easily.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Solved. See [here](https://github.com/sosy-lab/java-smt/pull/540#discussion_r2483716968).

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Solved. See [here](https://github.com/sosy-lab/java-smt/pull/540#discussion_r2483716968).

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaCreator.java
baierd: No. We don't want to create new ones in all cases, just in controlled and known cases like this one.

Suppose we change the code, or the behavior of CVC5 changes in an update; suddenly we create NEW variables without knowing where they come from. If this happens in a complex program, you will have difficulty finding it. So we just want to handle this very specific case and let it run into an exception otherwise.

baierd:
I started some experiments to see how well MathSAT and Bitwuzla output works with CVC5. I can also start Z3 once we know how that went.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +85,266 @@ static Term getCVC5Term(Formula pT) {
...
+    parseSolver.setOption("print-success", "true");
+    // more tolerant of non-conforming inputs, default: default
+    parseSolver.setOption("parsing-mode", "lenient");
+    // force all declarations and definitions to be global when parsing, default: false
+    parseSolver.setOption("global-declarations", "true");
+    // use API interface for fresh constants when parsing declarations and definitions, default:
+    // true
+    parseSolver.setOption("fresh-declarations", "false");
+    // allows overloading of terms and sorts if true, default: true
+    parseSolver.setOption("term-sort-overload", "false");
```
daniel-raffler: Could you add some more information on what these options do? I had to go through the changelogs to find some of them 😅

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
```diff
@@ -66,47 +85,276 @@ static Term getCVC5Term(Formula pT) {
...
+    StringBuilder declaration =
+        getSMTLIB2DeclarationsFor(ImmutableMap.of(symbolNotDeclared, knownTerm));
+    // TODO: insert after options if options present?
+    return parseCVC5(declaration.append(formulaStr).toString());
+  }
+
+  /**
+   * Builds the parser from the given {@link Solver} and the {@link SymbolManager}. The {@link
+   * SymbolManager} needs to be persistent to remember terms already parsed/created.
+   */
```
daniel-raffler: I think this comment is outdated

baierd:
> I started some experiments to see how well MathSAT and Bitwuzla output works with CVC5. I can also start Z3 once we know how that went.

Results show only a few parsing related errors (but not all that much has been parsed): 
[results.2025-11-02_13-29-47.table.html](https://github.com/user-attachments/files/23290331/results.2025-11-02_13-29-47.table.html)
[cvc5Parser.integration-imc.2025-11-02_13-00-50.logfiles.zip](https://github.com/user-attachments/files/23290332/cvc5Parser.integration-imc.2025-11-02_13-00-50.logfiles.zip)


The main problem is the missing `FP to IEEE-BV` support for CVC5 in general. (See [PR 514](https://github.com/sosy-lab/java-smt/pull/514)). With an additional complication in the parser for this operation (e.g. for MathSAT5 to CVC5):
```
Exception in thread "main" java.lang.IllegalArgumentException: Error parsing with CVC5: Symbol 'fp.as_ieee_bv' not declared as a variable
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5FormulaManager.parseCVC5(CVC5FormulaManager.java:164)
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5FormulaManager.parseImpl(CVC5FormulaManager.java:88)
	at org.sosy_lab.java_smt.solvers.cvc5.CVC5FormulaManager.parseImpl(CVC5FormulaManager.java:38)
	at org.sosy_lab.java_smt.basicimpl.AbstractFormulaManager.parse(AbstractFormulaManager.java:330)
	at org.sosy_lab.java_smt.basicimpl.AbstractFormulaManager.translateFrom(AbstractFormulaManager.java:488)
	at org.sosy_lab.cpachecker.util.predicates.smt.FormulaManagerView.translateFrom(FormulaManagerView.java:1907)
	at org.sosy_lab.cpachecker.util.predicates.interpolation.SeparateInterpolatingProverEnvironment.convertToItp(SeparateInterpolatingProverEnvironment.java:122)
```

Commandline to trigger: `bin/cpachecker --no-output-files --heap 2000M --bmc-interpolation --option imc.fallBack=false --option solver.solver=MATHSAT5 --option solver.interpolationSolver=CVC5 --option cpa.predicate.memoryAllocationsAlwaysSucceed=true --timelimit 60s --stats --spec test/programs/benchmarks/properties/unreach-call.prp --32 test/programs/benchmarks/float-benchs/inv_Newton.c.p+cfa-reducer.c`

Triggering SMTLIB2:
[cvc5.parse.fp_to_bv.smt2.txt](https://github.com/user-attachments/files/23290452/cvc5.parse.fp_to_bv.smt2.txt)
(Remove the `.txt` at the end!)

Another problem is that some symbols are still parsed multiple times:
```

Exception in thread "main" java.lang.IllegalArgumentException: <string>:11:14: symbol '|main::nondet@3|' already defined at line 2 column 14
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.BitwuzlaNativeJNI.Parser_parse__SWIG_0(Native Method)
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.Parser.parse(Unknown Source)
	at org.sosy_lab.java_smt.solvers.bitwuzla.BitwuzlaFormulaManager.parseImpl(BitwuzlaFormulaManager.java:126)
	at org.sosy_lab.java_smt.solvers.bitwuzla.BitwuzlaFormulaManager.parseImpl(BitwuzlaFormulaManager.java:28)
	at org.sosy_lab.java_smt.basicimpl.AbstractFormulaManager.parse(AbstractFormulaManager.java:330)
	at org.sosy_lab.java_smt.basicimpl.AbstractFormulaManager.translateFrom(AbstractFormulaManager.java:488)
	at org.sosy_lab.cpachecker.util.predicates.smt.FormulaManagerView.translateFrom(FormulaManagerView.java:1907)
```

SMTLIB2:
[cvc5.parse.extra.variable.smt2.txt](https://github.com/user-attachments/files/23290453/cvc5.parse.extra.variable.smt2.txt)
(Remove the `.txt` at the end!)

baierd:
Note: the second parsing problem is a Bitwuzla problem ;D
I opened a issue.

Seems like the parser works besides the FP to IEEE BV issue. We can tackle that later, as we might have to contact the devs for this. I opened a issue for this.

@kfriedberger this is now ready for review.

src/org/sosy_lab/java_smt/solvers/cvc5/CVC5FormulaManager.java
baierd: Fixed

kfriedberger (approved):
fine for me

PR #541, opened 2025-10-28
Quickfix for array model generation in Bitwuzla

daniel-raffler:
This MR adds Bitwuzla multi-dimensional array model evaluation.
(Similar to https://github.com/sosy-lab/java-smt/pull/539 for CVC5)

Also, more tests for this not based on SMTLIb2 parsing are added.

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaModel.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaModel.java
@@ -10,10 +10,11 @@ package org.sosy_lab.java_smt.solvers.bitwuzla;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static org.sosy_lab.common.collect.Collections3.transformedImmutableListCopy;
 
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -23,17 +24,13 @@ import org.sosy_lab.java_smt.solvers.bitwuzla.api.Bitwuzla;
 import org.sosy_lab.java_smt.solvers.bitwuzla.api.Kind;
 import org.sosy_lab.java_smt.solvers.bitwuzla.api.Sort;
 import org.sosy_lab.java_smt.solvers.bitwuzla.api.Term;
-import org.sosy_lab.java_smt.solvers.bitwuzla.api.TermManager;
 import org.sosy_lab.java_smt.solvers.bitwuzla.api.Vector_Term;
 
 class BitwuzlaModel extends AbstractModel<Term, Sort, Void> {
 
   // The prover env, not the creator env!
   private final Bitwuzla bitwuzlaEnv;
-
   private final BitwuzlaTheoremProver prover;
-
-  private final BitwuzlaFormulaCreator bitwuzlaCreator;
   private final ImmutableList<ValueAssignment> model;
 
   protected BitwuzlaModel(
@@ -44,7 +41,6 @@ class BitwuzlaModel extends AbstractModel<Term, Sort, Void> {
     super(prover, bitwuzlaCreator);
     this.bitwuzlaEnv = bitwuzlaEnv;
     this.prover = prover;
-    this.bitwuzlaCreator = bitwuzlaCreator;
 
     // We need to generate and save this at construction time as Bitwuzla has no functionality to
     // give a persistent reference to the model. If the SMT engine is used somewhere else, the
@@ -92,14 +88,15 @@ class BitwuzlaModel extends AbstractModel<Term, Sort, Void> {
    * @return A list of value assignments for all elements in the array, including nested arrays
    */
   private List<ValueAssignment> buildArrayAssignments(Term expr, Term value) {
-    TermManager termManager = bitwuzlaCreator.getTermManager();
-
     // Iterate down the Store-chain: (Store tail index element)
     List<ValueAssignment> result = new ArrayList<>();
     while (value.kind().equals(Kind.ARRAY_STORE)) {
       Term index = value.get(1);
       Term element = value.get(2);
-      Term select = termManager.mk_term(Kind.ARRAY_SELECT, expr, index);
+      Term select =
+          ((BitwuzlaFormulaCreator) creator)
+              .getTermManager()
+              .mk_term(Kind.ARRAY_SELECT, expr, index);
 
       // CASE 1: nested array dimension, let's recurse deeper
       if (expr.sort().array_element().is_array()) {
@@ -111,12 +108,10 @@ class BitwuzlaModel extends AbstractModel<Term, Sort, Void> {
             new ValueAssignment(
                 creator.encapsulate(creator.getFormulaType(element), select),
                 creator.encapsulate(creator.getFormulaType(element), element),
-                creator.encapsulateBoolean(termManager.mk_term(Kind.EQUAL, select, element)),
+                creator.encapsulateBoolean(buildEqForTwoTerms(select, element)),
                 getVar(expr),
                 creator.convertValue(element, element),
-                FluentIterable.from(getArrayIndices(select))
-                    .transform(creator::convertValue)
-                    .toList()));
+                transformedImmutableListCopy(getArrayIndices(select), this::evaluateImpl)));
       }
 
       // Move to the next Store in the chain
@@ -136,7 +131,7 @@ class BitwuzlaModel extends AbstractModel<Term, Sort, Void> {
     if (left.sort().is_fp() || right.sort().is_fp()) {
       kind = Kind.FP_EQUAL;
     }
-    return bitwuzlaCreator.getTermManager().mk_term(kind, left, right);
+    return ((BitwuzlaFormulaCreator) creator).getTermManager().mk_term(kind, left, right);
   }
 
   private ValueAssignment getAssignmentForUfInstantiation(Term pTerm) {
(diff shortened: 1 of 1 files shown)


baierd:
@daniel-raffler could you maybe add a test that (for Integer and BV if possible) that uses multi-dimensional arrays, but with at least 1 store instruction?

Bonus points if the store itself stores a multi-dimensional array.

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,453 @@ public void arrayTest5()
...
+            } else {
+              // Assume its an UF value
+              return args;
+            }
+          }
+        });
+  }
+
+  /** Evaluate the UF or array constraint and check that the model contains the expected values. */
+  private <T> void checkModelContains(
```
baierd: Methods like this already exist AFAIK. Could you check and reuse the existing ones?

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,453 @@ public void arrayTest5()
...
+
+    var idxA = bvmgr.makeBitvector(8, 1);
+    var valA = bvmgr.makeBitvector(8, 10);
+
+    var idxB = bvmgr.makeBitvector(8, 2);
+    var valB = bvmgr.makeBitvector(8, 5);
+
+    checkModelContains(
+        bmgr.and(
+            bvmgr.equal(amgr.select(array, idxA), valA),
```
baierd, line 2595: Is this not covered by existing tests?
(It might be that there are only Integer based tests for this)

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,453 @@ public void arrayTest5()
...
+            bvmgr.equal(amgr.select(array, idxB), valB)),
+        "array",
+        ImmutableMap.of(ImmutableList.of(idxA), valA, ImmutableList.of(idxB), valB));
+  }
+
+  @Test
+  public void testArray1BvBV() {
```
baierd, line 2602: A description, e.g. SMTLIB2, of whats tested should be added to all new tests.

daniel-raffler:
> @daniel-raffler could you maybe add a test that (for Integer and BV if possible) that uses multi-dimensional arrays, but with at least 1 store instruction?
> 
> Bonus points if the store itself stores a multi-dimensional array.

I've added https://github.com/sosy-lab/java-smt/pull/541/commits/68cb59c25f0da60312cea0c0f70bf88e9bf89274 from the [modelArrayGeneration](https://github.com/sosy-lab/java-smt/tree/modelArrayGeneration) branch and two more tests for 3d arrays that use stores. Is that what you had in mind?

> LGTM. But i would like to test this with at least the 2 tests from the related CVC5 branch, if possible more.

The tests from the CVC5 branch also seem to be working

daniel-raffler:
> Thanks for adding the tests! I think many can be combined to test the behavior in 2 tests (per theory). Something like:
> 
>     1. Nested arrays with UFs (you can combine store and select)
> 
>     2. Nested arrays with UFs and non UFs (store and select combined)
> 
> 
> (The CVC5 branch already has tests for only nested arrays w select, so no need for that)

I've removed the UF tests again as they were meant to test assignments for uninterpreted functions and aren't needed here. Other than that there is some overlap with the tests from the CVC5 branch and I'm not if we can't just remove all of the tests here.

What do you think still needs testing that isn't already covered by the CVC5 branch?

I'm getting different results for `testArray1BvBV` and `testArray2BvBV` on Z3 and Yices (EDIT: See [here](https://sosy-lab.gitlab.io/-/software/java-smt/-/jobs/11899950688/artifacts/JUnit.html#ModelTest) for the JUnit results). In the first test we get an incomplete model while the second test crashes, so maybe these two should be kept? Since `testArray1BvBV` is essentially the same as `testGetArrays3BitvectorNoParsing` from the CVC5 branch we would then only need to add `testArray2BvBV` (and maybe `testArray2IntInt` for completeness sake)

src/org/sosy_lab/java_smt/test/ModelTest.java
daniel-raffler: Sure, I can rewrite the tests to use `testModelGetters` just as on the CVC5 branch

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaModel.java
```diff
@@ -108,94 +108,91 @@ private ValueAssignment getSimpleAssignment(Term pTerm) {
...
 
-  private String getArrayName(Term array) {
-    String name = array.symbol();
-    if (name != null) {
-      return name;
+    } else {
+      // Should be unreachable
+      // We assume that array values are made up of "const" and "store" nodes with non-array
+      // constants as leaves
+      throw new AssertionError();
```
baierd: A more expressive exception and message could be added here.

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,312 @@ public void arrayTest5()
     checkModelIteration(formula, false);
   }
 
+  /** Returns the arguments of an UF or array definition from the model. */
+  private Iterable<Formula> getFunctionArgs(Formula pValue) {
```
baierd, line 2501: Actually why don't we have a visitor method for retrieving all UFs, including all applications? @kfriedberger do you know? Seems like this might be helpful in general.

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,312 @@ public void arrayTest5()
...
+    requireVisitor();
+    return mgr.visit(
+        pValue,
+        new DefaultFormulaVisitor<>() {
+          @Override
+          protected Iterable<Formula> visitDefault(Formula f) {
+            throw new IllegalArgumentException();
```
baierd, line 2508: Are you sure this is correct?
Since you don't override all methods of the super-class, this can be called!

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,312 @@ public void arrayTest5()
...
+          }
+
+          @Override
+          public Iterable<Formula> visitFunction(
+              Formula f, List<Formula> args, FunctionDeclaration<?> functionDeclaration) {
+            if (functionDeclaration.getKind().equals(FunctionDeclarationKind.SELECT)) {
+              return FluentIterable.concat(
+                  getFunctionArgs(args.get(0)), ImmutableList.of(args.get(1)));
+            } else {
+              // Assume its an UF value
```
baierd, line 2524: `visitFunction()` is defined to be called on any function, including UFs. I would recommend explicitly checking for UFs. You could take a look at the `VariableAndUFExtractor` for inspiration.

baierd (changes requested):

src/org/sosy_lab/java_smt/test/ModelTest.java
daniel-raffler: Yeah, that's fine. The argument to this function is a "key" formula from the model. So it's either `(f 1 2 3)` for a UF mapping `f(1,2,3)` or `(Select (Select (Select 1 arr) 2) 3)` for an array mapping `arr(1,2,3)`. We then get the arguments `1`, `2` and `3` by simply returning them for the UF mapping: this is the `else` branch in `visitFunction`. Or descending the term recursively if we have an array: that's the `if` part of `visitFunction` and `visitFreeVariable` for the base case

The whole function could be removed if we rewrite the tests to use `testModelGetters`

src/org/sosy_lab/java_smt/test/ModelTest.java
kfriedberger: This might be related to [FormulaManager#extractVariablesAndUFs](https://sosy-lab.github.io/java-smt/api/org/sosy_lab/java_smt/api/FormulaManager.html#extractVariablesAndUFs(org.sosy_lab.java_smt.api.Formula)), but would provide more insights into the formula. If required, we could add it into the API.

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,312 @@ public void arrayTest5()
...
+
+    checkModelContains(
+        bmgr.and(
+            imgr.equal(amgr.select(amgr.select(array, idxA1), idxA2), valA),
+            imgr.equal(amgr.select(amgr.select(array, idxB1), idxB2), valB)),
+        "array",
+        ImmutableMap.of(
+            ImmutableList.of(idxA1, idxA2), valA, ImmutableList.of(idxB1, idxB2), valB));
+  }
+
```
kfriedberger, line 2822: The new tests are good.

src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2254,6 +2259,312 @@ public void arrayTest5()
...
+            } else {
+              // Assume its an UF value
+              return args;
+            }
+          }
+        });
+  }
+
+  /** Evaluate the UF or array constraint and check that the model contains the expected values. */
+  private void checkModelContains(
```
kfriedberger, line 2532: Please note that the model-listing should only provide a fast overview of assigned variables. In several cases the solver will not be able to provide a useful interpretation, such as in cases like:
- partial model (if a part of the formula is irrelevant for satisfiability)
- quantifiers (e.g., infinite number of assignments for `FORALL x : array[x]=x`)
- arrays and UFs (the solver is allowed to assume a default value that is not provided as model listing)

The user always can query an evaluation directly, without explicitly depending on the model listing.

daniel-raffler: Hi Karlheinz,
I've made the tests a bit more permissive to cover you last case, and solvers may now drop one of the assignments if it is the default value. This branch should be largely the same as https://github.com/sosy-lab/java-smt/pull/539 and your changes from there could easily be copied over.

PR #542, opened 2025-10-31
Allow Mathsat to use its extended dumping method (allowing let-expressions and quantifier dumping)

baierd:
Adds 2 options to allow Mathsat to use its extended dumping method, which allows let-expressions and quantifier dumping.

Tests pass for all.

Let-expressions seem to be an alternative that allows SMTLib2 output to not use the `.` in front of additional bindings names (e.g. `(define-fun .def_1 ...)`. (They are still present, but no longer used in a critical way it seems. This needs to be tested more though.)

Also, this prepares support for the quantifier-manager in Mathsat. (Mathsat can't solve quantified formulas, but create and dump them!)

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5SolverContext.java
+++ b/src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5SolverContext.java
@@ -60,6 +60,26 @@ public final class Mathsat5SolverContext extends AbstractSolverContext {
                 + "Format is 'key1=value1,key2=value2'")
     private String furtherOptions = "";
 
+    @Option(
+        secure = true,
+        description =
+            "If enabled, SMTLIB2 output of MathSAT5 will be generated using the extended dumping "
+                + "function msat_to_smtlib2_ext(), instead of the regular dumping "
+                + "function msat_to_smtlib2(). This extended dump might allow the exporting of "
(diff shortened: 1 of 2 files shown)


src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5SolverContext.java
```diff
@@ -60,6 +60,26 @@ private static final class Mathsat5Settings {
...
+                + "let-expressions per default. If you want to use let bindings, enable option "
+                + "solver.mathsat5.dumpSMTLIB2LetExpressions as well.")
+    private boolean useExtendedSMTLIB2Output = false;
+
+    @Option(
+        secure = true,
+        description =
+            "If used together with solver.mathsat5.useExtendedSMTLIB2Output, SMTLIB2 output will "
+                + "contain let-bindings instead of define-fun based bindings.")
+    private boolean dumpSMTLIB2LetExpressions = false;
```
daniel-raffler, line 81: I think we may not even need these options and could just use always `msat_to_smtlib2_ext` with `let` bindings by default

baierd: Maybe we will switch in the future. We know that the extended dump can dump things the normal one can't. But we don't know whether or not there are limits/problems. Hence i want to test this first, and the options allow me to test it in other applications (CPAchecker) and users can also benefit.

Issue #503, opened 2025-08-06
NoSuchElementException for MathSAT5 Model Despite Its Iterator Claiming That There Should Be an Element


baierd:
MathSAT5 throws a `NoSuchElementException` in the model in the following CPAchecker run `--symbolicExecution --stats --spec test/programs/benchmarks/properties/unreach-call.prp --32 --no-output-files test/programs/benchmarks/hardness-nfm22/hardness_fillercode_fillercodesize_ps-cn-25_file-4.i` in this [branch/MR](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/255) latest state. This looks like a bug, either from us or MathSAT5. We should investigate and either fix this, or tell the MathSAT5 devs about it.

Stacktrace (excerpt):
```
...
Starting analysis ... (CPAchecker.runAlgorithm, INFO)

(= |var_1_35#45@1| #b00000000)
(= |var_1_35#45@1| #b00000010)
Exception in thread "main" java.util.NoSuchElementException
	at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5Model.asList(Mathsat5Model.java:61)
	at org.sosy_lab.java_smt.basicimpl.CachingModel.asList(CachingModel.java:40)
	at org.sosy_lab.java_smt.api.BasicProverEnvironment.getModelAssignments(BasicProverEnvironment.java:112)
	at org.sosy_lab.cpachecker.util.predicates.smt.BasicProverEnvironmentView.getModelAssignments(BasicProverEnvironmentView.java:71)
	at org.sosy_lab.cpachecker.cpa.constraints.domain.ConstraintsSolver.handleSolverResult(ConstraintsSolver.java:391)
	at org.sosy_lab.cpachecker.cpa.constraints.domain.ConstraintsSolver.checkUnsat(ConstraintsSolver.java:359)
	at org.sosy_lab.cpachecker.cpa.constraints.domain.ConstraintsSolver.checkUnsatWithOptionDefinedSolverReuse(ConstraintsSolver.java:269)
	at org.sosy_lab.cpachecker.cpa.constraints.ConstraintsTransferRelation.getIfSatisfiable(ConstraintsTransferRelation.java:337)
	at org.sosy_lab.cpachecker.cpa.constraints.ConstraintsTransferRelation$ValueAnalysisStrengthenOperator.strengthen(ConstraintsTransferRelation.java:388)
[...]

PR #544, opened 2025-11-01
#503: update MathSAT5 to version 5.6.12

kfriedberger:
MathSAT5: update MathSAT5 to v5.6.12., which comes with updated dependencies:

- On Windows systems, MPIR is replaced with GMP.
  MPIR was no longer developed since 2017, as there was GMP as alternative anyway.
  Now, MPIR is officially dead, and it is removed from MathSAT and also from JavaSMT.

- The Linux version of MathSAT5 requires GLIBC 2.38 or newer,
  available at least in Ubuntu 24.04 or newer.
  In this PR, we also updated our development Podman/Docker image to Ubuntu 24.04.

### DIFF
--- a/doc/Developers-How-to-Release-into-Ivy.md
+++ b/doc/Developers-How-to-Release-into-Ivy.md
@@ -271,8 +271,9 @@ but in the normal system environment, where some testing can be applied by the d
 We publish MathSAT for Linux (x64 and arm64) and Windows (x64) systems at once.
 The build process can fully be done on a Linux system, 
 and requires several dependencies, such as gcc, x86_64-w64-mingw32-gcc, and aarch64-linux-gnu-gcc.
-We prefer to use the Docker container based on Ubuntu 22.04 (better use Ubuntu 18.04 for older GLIBC!)
+We prefer to use the Docker container based on Ubuntu 24.04
 for compiling the dependencies and assembling the libraries.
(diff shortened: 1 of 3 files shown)


baierd:
Looks good to me! Nice work!

Do you want me to test it on Windows (i have a proper Windows machine at home)?

baierd (approved):

kfriedberger:
Thanks for the review.

The changes are already tested twice on Windows: 
- CI with AppVeyor: https://ci.appveyor.com/project/sosy-lab/java-smt/builds/53006348
- locally on a Win11 machine

Whats still missing, is a complete evaluation with a CPAchecker benchmark whether the model issues are solved.

PhilippWendler:
Ubuntu 24.04 seems really new for me to require as a baseline. This would mean most CPAchecker configs for example do not even run on Ubuntu 22.04 anymore, which still has almost 2 years of support.

I think I remember that in the past the MathSAT developers have produced binaries for us with greater backwards compatibility after we asked. Maybe it is worth it asking again?

kfriedberger:
We can consider the request for supporting older versions of GLIBC with the next version of MathSAT5, maybe when solving #547. Please note that this is an exceptional case for this solver because of the long history of MathSAT5 being the default solver in CPAchecker. We will not actively go after other pre-built solvers and their dependendencies.

PR #550, opened 2025-11-15
CVC5: extend publishing scripts for CVC5 on Windows with ARM64.

kfriedberger:
I have no hardware or software available for testing the solver natively on Windows on ARM64, and we do not have any CI or configuration for that. The same issue arises with other solvers as well, such as Z3. Lets just assume, the binary library works as expected.


daniel-raffler (approved):
Thanks for the quick fix!

LGTM

PR #552, opened 2025-11-16
MathSAT: Disable enumerations when OptiMathSat is used

daniel-raffler:
Hello,
this PR fixes the MathSAT visitor by skipping the  `msat_is_enum_type` call when OptiMathSAT is used. OptiMathSAT does not appear to support enumerations and JavaSMT will crash when we try to call this method in the visitor code


kfriedberger (approved):
OptiMathSAT is quote old anyway, without any recent development.
PR looks good. simple plain solution, can be merged. Thanks.

PR #553, opened 2025-11-17
Model usage: fix usage of model values and model evaluation after changing the underlying context.

kfriedberger:
Several solvers referenced a live prover context in the model, allowing the user to insert further queries in the model or remove queries from the context, potentially making the model inconsistent and unreliable for further iteration or evaluation.

The solution consists of several steps, for some solvers:
- compute the value-assignments upfront when constructing the model.
- invalidate the model once the context is changed.

Solvers that already provide a persistent model, such as MathSAT, SMTInterpol, and Z3, will not change their behaviour.
However, other solvers will throw an exception on invalid usage.

While this change is backward-compatible in the API, it is a serious bugfix that changes the behaviour of some solver bindings in JavaSMT, because we need to loosen a basic guarantee for models.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
@@ -166,7 +166,20 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
 
   protected abstract boolean isUnsatImpl() throws SolverException, InterruptedException;
 
-  /** Override for OptimizationProver */
+  /**
+   * Performs an optimization-aware check and returns the optimization status.
+   *
+   * <p>This method is the public entry point for optimization checks. It validates that the prover
+   * is open, resets internal change-tracking state, and delegates solver-specific work to {@link
+   * #checkImpl()}. Subclasses that implement optimization support must provide the actual checking
+   * logic in {@code checkImpl()}.
+   *
+   * <p>The signature of this method matches that of {@link OptimizationProverEnvironment#check()},
+   * to allow overrides in subclasses that implement both {@link BasicProverEnvironment} and {@link
+   * OptimizationProverEnvironment}.
+   *
+   * @return the optimization status; {@link OptStatus#OPT} indicates a satisfiable/optimal result
+   */
   public final OptStatus check() throws InterruptedException, SolverException {
     checkState(!closed);
     wasLastSatCheckSatisfiable = false;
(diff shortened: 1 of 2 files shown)


src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -29,12 +29,17 @@
...
   protected final boolean generateModels;
   protected final boolean generateAllSat;
   protected final boolean generateUnsatCores;
   private final boolean generateUnsatCoresOverAssumptions;
   protected final boolean enableSL;
+
+  // flags for status
   protected boolean closed = false;
+  protected boolean wasLastSatCheckSatisfiable = true; // assume SAT for an empty prover
+  protected boolean changedSinceLastSatQuery = false; // assume not-changed for an empty prover
```
daniel-raffler: I think `changedSinceLastSatQuery` should be `true` initially. Otherwise we could get the model (or an unsat core) without having called `(check-sat)` first

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
```diff
@@ -196,9 +192,7 @@ private List<BooleanFormula> getUnsatCore0() {
    */
   @Override
   public List<BooleanFormula> getUnsatCore() {
```
baierd, line 188: UnsatCores can also affect solver states (e.g. in Z3). It might be a good idea to generally assume that they do?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
```diff
@@ -84,10 +86,11 @@ public void close() {
   @Override
   public boolean isUnsat() throws SolverException, InterruptedException {
     Preconditions.checkState(!closed);
-    wasLastSatCheckSat.set(false);
+    changedSinceLastSatQuery = false;
```
baierd: We can add a default implementation for isUnsat() and simplify these flags as well. The same goes for other calls like allSAT.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaTheoremProver.java
kfriedberger: Can you give an example? Why would Z3 change its state when calling `getUnsatCore`?

src/org/sosy_lab/java_smt/solvers/boolector/BoolectorAbstractProver.java
kfriedberger: I thought about that, too.

kfriedberger: fixed

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
kfriedberger: thanks for noticing. fixed.

PR #555, opened 2025-11-26
MathSAT: update to version 5.6.14

kfriedberger:


daniel-raffler:
This is a minor issue, but MathSAT no longer seems to return its version number when `SolverContext.getVersion()` is called. Instead it will now print a git hash, which isn't terribly useful as we don't have the source. Is there a way to override this in our bindings? Or should we report this to the MathSAT developers?

kfriedberger:
Maybe the MathSAT developers forgot to update their version-file. :smile: :cloud: 
Let's just keep it that way. If MathSAT5 reports it's own version as Git hash, just forward it as-is. :heavy_check_mark:

PhilippWendler:
For users a git hash of a non-open-source tools is useless. I would highly appreciate it if I could still find out from the logs of CPAchecker which MathSAT version was used. Please make it report `5.6.14` if that is what you are using.

kfriedberger:
The missing version numbers of MathSAT might be related to the change of their build-system. We should inform Alberto about it.
In the mean time, we can modify the version string directly, at least for public released major versions.

```
$ ./mathsat-5.6.11-win64-msvc/bin/mathsat.exe -version
MathSAT5 version 5.6.11 (968f05bc5839) (Aug  9 2024 10:46:25, gmp 6.0.0/mpir 3.0.0, msvc 19.12, 64-bit)

$ ./mathsat-5.6.12-win64/bin/mathsat.exe -version
MathSAT5 version 031d838a61a5 (Sep 17 2025 11:59:25, gmp 6.3.0, gcc 13.0.0, 64-bit, reentrant)

$ ./mathsat-5.6.13-win64/bin/mathsat.exe -version
MathSAT5 version 9f6eb52a7aea (Nov 13 2025 12:06:00, gmp 6.3.0, gcc 13.0.0, 64-bit, reentrant)

$ ./mathsat-5.6.14-win64/bin/mathsat.exe -version
MathSAT5 version 4b32db6b0e84 (Nov 14 2025 13:39:41, gmp 6.3.0, gcc 13.0.0, 64-bit, reentrant)
```

PR #559, opened 2025-12-20
Add generic equality and distinct to the API

daniel-raffler:
Hello everyone,
in SMTLIB `=` and `distinct` are defined for all sorts and with a variable number of argument:
```
(par (A) (= A A Bool :chainable))
(par (A) (distinct A A Bool :pairwise))
```
We already support equality, but it's a different method for each `FormulaManager`, and only two arguments can be used. Distinct is only supported for bitvectors and numeral formulas, but missing from the other formula managers. While it's possible to simulate `distinct` by rewriting the formula, this leads to quadratically many terms, which may not always be desirable.

This PR tries to fix the issue by adding two generic methods for `equality` and `distinct` to the `FormulaManager`. The changes to the API are backwards compatible, and I've included a default implementation of `distinct` and `=` with multiple arguments for solvers that don't support these operations natively

### DIFF
--- a/src/org/sosy_lab/java_smt/test/FormulaManagerTest.java
+++ b/src/org/sosy_lab/java_smt/test/FormulaManagerTest.java
@@ -10,7 +10,6 @@ package org.sosy_lab.java_smt.test;
 
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.TruthJUnit.assume;
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 import static org.sosy_lab.java_smt.api.FormulaType.BooleanType;
 import static org.sosy_lab.java_smt.api.FormulaType.IntegerType;
@@ -77,11 +76,11 @@ public class FormulaManagerTest extends SolverBasedTest0.ParameterizedSolverBase
     var formulaType = imgr != null ? IntegerType : getBitvectorTypeWithSize(8);
     var var1 = mgr.makeVariable(formulaType, "var1");
 
-    assertEquals(bmgr.makeTrue(), mgr.equal());
-    assertEquals(bmgr.makeTrue(), mgr.distinct());
+    assertThat(mgr.equal()).isEqualTo(bmgr.makeTrue());
+    assertThat(mgr.distinct()).isEqualTo(bmgr.makeTrue());
 
-    assertEquals(bmgr.makeTrue(), mgr.equal(var1));
-    assertEquals(bmgr.makeTrue(), mgr.distinct(var1));
+    assertThat(mgr.equal(var1)).isEqualTo(bmgr.makeTrue());
+    assertThat(mgr.distinct(var1)).isEqualTo(bmgr.makeTrue());
   }
 
   @Test


src/org/sosy_lab/java_smt/test/FormulaManagerTest.java
```diff
@@ -31,6 +35,144 @@
...
+            + String.format("(define-const %s %s)", var2, formulaSort)
+            + String.format("(define-const %s %s)", var3, formulaSort)
+            + String.format("(assert (distinct %s %s %s))", var1, var2, var3);
+
+    var f = mgr.distinct(var1, var2, var3);
+    var g = mgr.parse(str);
+
+    assertThat(f).isEqualTo(g);
+  }
+
```
kfriedberger, line 243: Please add futher tests, e.g. for:
- zero items as parameters
- one item as parameters
- visitor of the resulting expression for several theories.

src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -131,6 +132,18 @@ <T extends Formula> T makeApplication(
    */
   <T extends Formula> T makeApplication(FunctionDeclaration<T> declaration, Formula... args);
 
+  default BooleanFormula equal(Formula... pArgs) {
+    return equal(ImmutableList.copyOf(pArgs));
+  }
+
+  BooleanFormula equal(List<Formula> pArgs);
```
kfriedberger: Someone implementing this interface will need to adjust their implementation. We could provide a default implementation throwing UnsupportedOperationException if not overridden.

I am unsure if we ever did it that way in JavaSMT. Maybe we just added new methods as it is done here. Then we can continue with that strategy, i.e., keep the PR as-is. 😄

src/org/sosy_lab/java_smt/test/FormulaManagerTest.java
daniel-raffler: I've now added a few more tests:
- zero and one parameters are covered by `testWrongNumberArgs`
- visitors are tested by `testEqualityVisitor` and `testDistinctVisitor`

src/org/sosy_lab/java_smt/api/FormulaManager.java
daniel-raffler: Hmm.. I think I'd prefer to keep it unimplemented. The compiler errors were quite helpful in figuring out what still needs to be updated. If we add a default implementation that throws `UnsupportedOperationException` users would have to read the changelog to find out that new methods were added

kfriedberger: Sure. Then lets keep as-is.

daniel-raffler:
> I changed the code to accept zero or one elements, too. This allows the user to skip special cases.

Good idea! Thanks

src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -131,6 +133,46 @@ <T extends Formula> T makeApplication(
...
   <T extends Formula> T makeApplication(FunctionDeclaration<T> declaration, Formula... args);
 
+  /**
+   * Create an equality formula between the given arguments. We return "true" if all arguments are
+   * equal, even if there are less than two arguments.
+   *
+   * @param pArgs Arguments to be compared for equality, ordering does not matter.
+   * @return Equality formula
+   */
+  default BooleanFormula equal(Formula... pArgs) {
```
PhilippWendler, line 143: Is this really a good naming scheme for methods on the `FormulaManager` interface?

Contrary to the other formula-manager interfaces, this interface contains a lot of other methods that are not for creating formulas. A method named `equal` does not seem to be immediately clear about what it is used for.

Furthermore, all other methods for creating formulas in the `FormulaManager` interface are named `make*`. I would suggest to use this naming scheme consistently.

src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -131,6 +133,46 @@ <T extends Formula> T makeApplication(
...
+    return equal(ImmutableList.copyOf(pArgs));
```
PhilippWendler, line 144: Why do an eager copy here and not just provide a view of the array?

src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -131,6 +133,46 @@ <T extends Formula> T makeApplication(
...
+  }
+
+  /**
+   * Create an equality formula between the given arguments. We return "true" if all arguments are
+   * equal, even if there are less than two arguments.
+   *
+   * @param pArgs Arguments to be compared for equality, ordering does not matter.
+   * @return Equality formula
+   */
+  BooleanFormula equal(Collection<Formula> pArgs);
```
PhilippWendler, line 154: Why `Collection` and not `Iterable`?

baierd:
@daniel-raffler we should incorporate the changes requested by @PhilippWendler. I think he makes very good points!

PR #560, opened 2025-12-20
Solver independent interpolation

juliusbrehme:
Implemented independent interpolation strategies to support solvers lacking native interpolation features.

Features added:
- Uniform Backward/Forward Interpolation.
- Projection-based Interpolation

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingSolverDelegate.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingSolverDelegate.java
@@ -10,11 +10,9 @@
 
 package org.sosy_lab.java_smt.basicimpl;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
-import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 import java.util.Collection;
 import java.util.List;
@@ -89,10 +87,6 @@ public class IndependentInterpolatingSolverDelegate<T> extends AbstractProver<T>
   @Override
   public BooleanFormula getInterpolant(Collection<T> identifiersForA)
       throws SolverException, InterruptedException {
-    Preconditions.checkState(!closed);
-    checkArgument(
-        getAssertedConstraintIds().containsAll(identifiersForA),
-        "Interpolation can only be performed over previously asserted formulas.");
 
     if (identifiersForA.isEmpty()) {
       return bmgr.makeTrue();


src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -197,6 +209,15 @@ public <T> void binaryInterpolationWithConstantFalse()
 
   @Test
   public <T> void binaryBVInterpolation1() throws SolverException, InterruptedException {
+    assume()
+        .withMessage(
+            "Solver %s with strategy %s is not supported or times out",
+            solverToUse(), itpStrategyToUse())
+        .that(
```
baierd, line 279: It might be better to split this assume statement into 2 for clarity.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
```diff
@@ -1191,6 +1217,8 @@ public <T> void issue381InterpolationTest1() throws InterruptedException, Solver
    */
   @Test
   public <T> void issue381InterpolationTest2() throws InterruptedException, SolverException {
+    requireIntegers();
+    requireSeqItp(itpStrategyToUse());
```
baierd: Can we just override `requireSeqItp()` in `ParameterizedInterpolatingSolverBasedTest0` to always include the strategy for only this class?

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
```diff
@@ -414,6 +489,54 @@ protected final BooleanFormulaSubject assertThatFormula(BooleanFormula formula)
     return assertUsing(context).that(formula);
   }
 
+  /** Check that the interpolant in the subject is valid fot the formulas A and B. */
+  public void isValidInterpolant(BooleanFormula interpolant, BooleanFormula a, BooleanFormula b)
+      throws SolverException, InterruptedException {
+    // TODO: move into assertion framework
+    isValidInterpolant(interpolant, ImmutableList.of(a), ImmutableList.of(b));
```
baierd: Could you look into adding this as an assertion function? Similar to the `isUnsatisfiable()` method (example: `assertThatFormula(query).isUnsatisfiable();`)

The new call would then need to interpolate as well of course.

src/org/sosy_lab/java_smt/example/IndependentInterpolation.java
```diff
@@ -0,0 +1,140 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2024 Dirk Beyer <https://www.sosy-lab.org>
```
baierd: 2026 ;D

src/org/sosy_lab/java_smt/example/IndependentInterpolation.java
```diff
@@ -0,0 +1,140 @@
...
+    // instead of 'named' formulas, we return a 'handle' (of generic type T)
+    prover.addConstraint(bmgr.and(imgr.equal(y, z), imgr.equal(z, two)));
+    T ip1 = prover.addConstraint(bmgr.and(imgr.equal(x, one), imgr.equal(y, x)));
+
+    // check for satisfiability
+    boolean unsat = prover.isUnsat();
+    Preconditions.checkState(unsat, "The example for interpolation should be UNSAT");
+
+    BooleanFormula itp = prover.getInterpolant(ImmutableList.of(ip1));
+    logger.log(Level.INFO, "Interpolants are:", itp);
```
baierd: We could add the interpolation technique used, as well as the interpolant (and the groups A and B if not too large) to the output.

src/org/sosy_lab/java_smt/basicimpl/InterpolatingSolverDelegate.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Set;
+import org.sosy_lab.java_smt.api.BasicProverEnvironment;
+import org.sosy_lab.java_smt.api.BooleanFormula;
+import org.sosy_lab.java_smt.api.InterpolatingProverEnvironment;
+import org.sosy_lab.java_smt.api.Model;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
+import org.sosy_lab.java_smt.api.SolverException;
+
+/** Delegate that wraps non-interpolating provers, allowing them to return itp tracking points. */
+public class InterpolatingSolverDelegate extends AbstractProver<String>
```
baierd, line 29: Maybe we need tests that involve the additional delegates as well, to check for problems when multiple are stacked.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingSolverDelegate.java
```diff
@@ -0,0 +1,432 @@
...
+    if (bmgr.isFalse(conjugatedFormulasOfA)) {
+      return bmgr.makeFalse();
+    } else if (bmgr.isFalse(conjugatedFormulasOfB)) {
+      return bmgr.makeTrue();
+    }
+
+    if (interpolationStrategy == null) {
+      interpolant = delegate.getInterpolant(identifiersForA);
+
+    } else if (interpolationStrategy.equals(ProverOptions.GENERATE_PROJECTION_BASED_INTERPOLANTS)) {
```
baierd: This could be improved by making it a switch-case.

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingSolverDelegate.java
```diff
@@ -0,0 +1,432 @@
...
+      checkState(
+          validationSolver.isUnsat(),
+          "Invalid Craig interpolation: interpolant does not contradict formula group B.");
+      validationSolver.pop();
+    }
+    return true;
+  }
+
+  /** interpolate(A(x,y),B(y,z))=∀z.¬B(y,z). */
+  private BooleanFormula getUniformBackwardInterpolant(
```
baierd: Maybe it would be beneficial to separate the distinct techniques into their own package and classes, as if we add more techniques, especially techniques with more code, this class will end up very large and hard to read. E.g. a new package `package org.sosy_lab.java_smt.basicimpl.interpolation_techniques;` with a class for each technique?

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -53,7 +53,37 @@ enum ProverOptions {
...
+    ENABLE_SEPARATION_LOGIC,
+
+    /**
+     * Enables Craig interpolation, using the model-based interpolation strategy. This strategy
+     * constructs interpolants based on the model provided by a solver, i.e. model generation must
+     * be enabled. This interpolation strategy is only usable for solvers supporting quantified
+     * solving over the theories interpolated upon. The solver does not need to support
+     * interpolation itself.
+     */
+    GENERATE_PROJECTION_BASED_INTERPOLANTS,
```
baierd, line 70: @kfriedberger @PhilippWendler do you guys have feedback/ideas/proposals for how we want the user to be able to switch in between native and solver independent interpolation techniques? Currently we just `ProverOptions`, which is not ideal, as they can't be switched once a prover has been created.

PhilippWendler, line 70: It might not be a good idea to add options that are only relevant for interpolating provers to the general `ProverOptions` class? What happens if I create a regular prover environment and tell it to generate interpolants?

PhilippWendler: Why would I want to switch the interpolation implementation for a prover instance and why would it be a problem to create a new prover instance instead? I don't really see a use case.

What I am interested in is to have the following 3 possibilities:
1. create an interpolating prover that guarantees that native solver interpolation is used (and creation fails if this is not possible)
2. create an interpolating prover that always uses the specific strategy implemented here
3. create an interpolating prover that prefers the solver native interpolation but falls back to a general implementation if the solver does not support interpolation

The third method could be done by catching the exception from the first case, but why not let JavaSMT provide it as well for the benefit of all its users?

baierd: I added this as a quick dummy. Hence the question about how we want to do it properly ;D

baierd: OK that seems reasonable.

I thought it might be good to decide about a interpolation technique when asking for interpolants, as users can implement point 3 themselves (potentially even with more than one fallback. 

Also interpolation is (usually) only possible after UNSAT has been returned. If a user wants to generate interpolants with multiple techniques in sequence, that would not be possible. E.g.: start with a technique like projection based interpolants that tends to return interpolants fast, but they also tend to be weak -> then use a stronger technique if that first interpolant is not useful. In your approach, a user would need to build a second (or n) new provers and then check satisfiability again.

(Just some thoughts. I want to provide the best possible API. I don't have a preference at the moment.)

@leventeBajczi you also use interpolation, maybe you have some thoughts?

leventeBajczi: I don't really have any further input to this -- as long as it's clear how I'm getting the interpolants, I don't mind if I need to create separate solver instances. We would probably always use point 3 if it was available, especially if it also handled the case when a solver returns an error for the interpolation query (although I'm not sure if the resulting solver state always stays usable). Failing that, we already have some support for using a portfolio of solvers, so we could easily implement the switching ourselves.

src/org/sosy_lab/java_smt/test/InterpolatingProverTest.java
juliusbrehme: I did this for requireTreeItp  as well

src/org/sosy_lab/java_smt/basicimpl/IndependentInterpolatingSolverDelegate.java
juliusbrehme: Added switch case, but the null can not be in a switch case

src/org/sosy_lab/java_smt/example/IndependentInterpolation.java
juliusbrehme: I added this if the formulas are not longer than 500 chars

src/org/sosy_lab/java_smt/test/SolverBasedTest0.java
juliusbrehme: Added this to BooleanFormulaSubject

baierd:
Note: i merged `master` into this branch. We need to merge the implementation of the 2 interpolating prover delegates! This seems to be causing issues. I will do it, but don't have the time right now and want to document it.

PR #568, opened 2025-12-27
Add more files to remove for 'ant clean'

daniel-raffler:
Remove autogenerated code (AutoValue, `.apt-generated`), JavaDoc files and JUnit test results when `ant clean` is called


kfriedberger (approved):
thanks. looks good. can be merged.

Issue #570, opened 2025-12-27
Segfaults in Bitwuzla


daniel-raffler:
Hello,
there are some intermittent segfaults when using Bitwuzla in CPAchecker. Here is an example stack trace:
```
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libbitwuzlaj.so+0x41af38]  bzla::type::TypeManager::garbage_collect(bzla::type::TypeData*)+0x28
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 5363  org.sosy_lab.java_smt.solvers.bitwuzla.api.BitwuzlaNativeJNI.delete_Sort(J)V (0 bytes) @ 0x00007ccadc65ebe9 [0x00007ccadc65eba0+0x0000000000000049]
J 7441 c2 org.sosy_lab.java_smt.solvers.bitwuzla.api.Sort.finalize()V (5 bytes) @ 0x00007ccadc5233f4 [0x00007ccadc523320+0x00000000000000d4]
J 8473% c2 java.lang.ref.Finalizer$FinalizerThread.run()V java.base@23-ea (39 bytes) @ 0x00007ccadc584e30 [0x00007ccadc584b20+0x0000000000000310]
v  ~StubRoutines::call_stub 0x00007ccadbf38cc6
```
The problem is non-deterministic and hard to reproduce. It may be linked to floating-point `Sort` objects that get garbage collected to early. We should try disabling the `Sort` finalizer to see if it makes any difference

Issue #545, opened 2025-11-02
Bitwuzla Parsing Symbols Twice With Error


baierd:
Bitwuzla parses a symbol twice and throws an error when parsing the following SMTLIB2 queries in order:
```
(declare-const |main::b@3| (_ BitVec 32))
(declare-const |main::a@4| (_ BitVec 32))
(declare-const |main::c@2| (_ BitVec 32))
(define-fun @def0 () Bool (= |main::a@4| #b00000000000000000000000000001010))
(assert (and (and (and @def0 (not @def0)) (bvsle |main::b@3| #b00000000000000000000000000000101)) (not (= |main::c@2| #b00000000000000000000000000000000))))
```
```
(declare-const |main::a@3| (_ BitVec 32))
(declare-const |main::a@4| (_ BitVec 32))
(assert (and (not (= |main::a@3| #b00000000000000000000000000001010)) (= |main::a@4| (bvadd |main::a@3| #b00000000000000000000000000000001))))
```
```
(declare-const nondet_int!2@ (_ BitVec 32))
(declare-const |main::nondet@3| (_ BitVec 32))
(declare-const |main::a@3| (_ BitVec 32))
(declare-const nondet_int!3@ (_ BitVec 32))
(declare-const |main::b@3| (_ BitVec 32))
(assert (and (and (and (= |main::nondet@3| nondet_int!2@) (= |main::a@3| #b00000000000000000000000000000000)) (= |main::b@3| nondet_int!3@)) (not (= |main::nondet@3| #b00000000000000000000000000000000))))
```
```
(assert false)
```
```
(assert false)
```

Exception in CPAchecker:
```
java.lang.IllegalArgumentException: <string>:11:14: symbol '|main::nondet@3|' already defined at line 2 column 14
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.BitwuzlaNativeJNI.Parser_parse__SWIG_0(Native Method)
	at org.sosy_lab.java_smt.solvers.bitwuzla.api.Parser.parse(Unknown Source)
	at org.sosy_lab.java_smt.solvers.bitwuzla.BitwuzlaFormulaManager.parseImpl(BitwuzlaFormulaManager.java:126)
	at org.sosy_lab.java_smt.solvers.bitwuzla.BitwuzlaFormulaManager.parseImpl(BitwuzlaFormulaManager.java:28)
[...]

Issue #529, opened 2025-09-26
Temporary input file "cadical"


kfriedberger:
I noticed a kind of temporary file named "cadical" in the project directory after running JUnit tests. This file seems to appear in JavaSMT as well as in CPAchecker. We should avoid this file or redirect it into a valid temp-directory. The file seems to contain a few chars, maybe encoding some SAT problem for Cadical which is used internally in some SMT solvers, such as Boolector, CVC4, CVC5, Yices2, and Bitwuzla.

**First steps:**
- find out which test or solver causes this file.
- then fix it.

daniel-raffler:
The file seems to be created by Bitwuzla, and I can get it when running `ModelTest`. It is much larger while the test is still running (~ 1.5MB) and the first line reads `aig 265631 9885 0 1595 255746`. Could this be related to the `aiger` option we enabled [some](https://github.com/sosy-lab/java-smt/commit/279eec14ddcf5991d8dfcb1aa21b7403ee8bdb36) time ago?

PR #574, opened 2026-01-02
Update Bitwuzla to 0.82 / bugfixes

daniel-raffler:
Hello,
this is an update to the Bitwuzla binaries that will close three issues:
- #570
The issue was that the Swig generated finalizer deletes the C++ object for `Sorts` when the Java proxy object is garbage colleced. This is an issue if the solver returns the same `Sort` object multiple times and it gets wrapped in several Java proxy objects: Once the first of these proxies gets garbage collected, the C++ object is also deleted and the other Java proxies now point to freed memory, which was causing the Segfaults. The solution for now is to simply not delete the C++ objects in the finalizer
- #529 
I believe this was caused by the `--aiger` flag, so I just recompiled without it and the files no longer seem to appear
- #545
Smtlib-style quotes (f.ex `|var|`) are now removed from symbol names. This is safe as JavaSmt can't create quoted symbols, and the Bitwuzla parser will always identify a quoted symbol with its simple symbol, if it exists. The quotes were causing issues when using `FormulaManager.translateFrom` to move formulas from one solver to another

The PR also updates Bitwuzla to version 0.8.2

@kfriedberger
Could you update the binaries, please? Bitwuzla now has dependencies on GMP and MPFR and can no longer fetch them as subprojects. I therefore had to update the Docker images to include those dependencies, and these images should probably also be uploaded. There is still a minor problem when building for `aarch64`, see the issue [here](https://github.com/bitwuzla/bitwuzla/issues/198). To work around it, you will have to add the line `pkg-config = 'aarch64-linux-gnu-pkg-config'` to the toolchain file `x86_64-linux-aarch64.txt` in the Bitwuzla main directory before running `ant publish-bitwuzla`

### DIFF
--- a/src/org/sosy_lab/java_smt/test/SolverContextTest.java
+++ b/src/org/sosy_lab/java_smt/test/SolverContextTest.java
@@ -200,7 +200,7 @@ public class SolverContextTest extends SolverBasedTest0.ParameterizedSolverBased
             bvmgr2.equal(bvmgr2.makeVariable(8, "x"), bvmgr2.makeBitvector(8, 17)));
         assertThat(prover2.isUnsat()).isFalse();
         try (Model model = prover2.getModel()) {
-          assertThat(model.asList().size()).isEqualTo(1);
+          assertThat(model.asList()).hasSize(1);
         }
       }
     }


lib/native/source/libbitwuzla/bitwuzla.i
```diff
@@ -110,7 +113,12 @@ namespace bitwuzla {
 }
 %extend Term {
   std::string symbol() {
-    return $self->symbol().value();
+    std::string sym = $self->symbol().value();
+    if (sym.front() == '|' && sym.back() == '|') {
+      return sym.substr(1, sym.size() - 2);
```
kfriedberger, line 118: This does not yet un-escape the symbol `|` (from escaped `\|`) within a symbol name.

daniel-raffler: Smtlib does not support escaping for `|` in symbol names:
```
A quoted symbol is any sequence of whitespace characters and printable characters that
starts and ends with | and does not contain | or \
```
The only way that we may see a `|` in the symbol name, is if it's a quoted symbol and the `|` surround the symbol name (as in `|main::a|`). In that I would argue that the `|`s aren't really part of the symbol name, but Bitwuzla still includes them, and that's why they have to be removed here

kfriedberger: Thanks for the hint. The reference is here: 
https://smt-lib.org/papers/smt-lib-reference-v2.7-r2025-02-05.pdf at `3.1. Lexicon > Symbols`

docker/ubuntu2404.Dockerfile
```diff
@@ -54,30 +55,71 @@ RUN apt-get update \
...
 # Install GMP for linux on x64 and arm64
-# We could add another build for windows when needed
-RUN wget https://gmplib.org/download/gmp/gmp-6.2.1.tar.lz \
- && tar xf gmp-6.2.1.tar.lz \
- && rm gmp-6.2.1.tar.lz \
- && cd gmp-6.2.1 \
+RUN wget https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz \
+ && tar xf gmp-6.3.0.tar.xz \
+ && rm gmp-6.3.0.tar.xz \
+ && cd gmp-6.3.0 \
```
kfriedberger, line 61: Are we sure, no other solver requeires that specific version of GMP?
--> ´The API of GMP is compatible in most cases. We will find out soon enough. This is not a blocker.

kfriedberger:
**Progress:**
- Podman images are updated. ✔️ 
- Bitwuzla cloned and patched: https://github.com/kfriedberger/bitwuzla/commit/7c96f469a86e492f2de53ac963c072594bbcd078 ✔️ 
- Library compiled and published to Ivy ✔️ 
- Dependency updated: see 64e4572df83bce464f663a1682edb00d19314bec ✔️ 

Further steps coming soon.

kfriedberger:
@daniel-raffler Thanks for the preparation of this PR. Well done.

daniel-raffler:
Thanks for the quick update!

Issue #171, opened 2019-11-02
Define the difference between Plain Symbols and Nullary UFs.


kfriedberger:
All solvers allow to define nullary UFs, i.e., UFs without arguments. For some solvers, the returned formula can be used just like a plain symbol/variable and is even comparable, i.e. identical. For other solvers, a nullary UF is different from a plain symbol and appears to be a separate object.

We should document, which solvers allow this and provide more tests for this (see fb0de870f61047f92ce872d3718bce6c4673872f).
Or we could return plain variables when requesting a nullary UF (or vice versa).

PR #580, opened 2026-01-14
Add support for Ufs with 0 arguments for all solver / bugfixes

daniel-raffler:
Hello,

this PR adds support for Ufs with no arguments to Bitwuzla and Boolector, where it was still missing. It also fixes various bugs on Princess, CVC4 and CVC5 where 0-ary Ufs and variables where treated as different symbols: Since all of these solvers are using a variable cache, it is our job to make sure that no duplicate symbols are created and `(uf)` and `uf` are treated the same


daniel-raffler:
Hello Karlheinz

> Is this PR a solution to the old issue #171 ? If yes, please document the new rules for nullary UFs and variables.

Yes, I think it closes it and I'll add some documentation. Most solvers expect at least one argument when defining a UF and will throw an exception otherwise. In that case we simply return a variable as a workaround. It's often hard to say if the solver considers variables and 0-ary UFs the same as most solvers allows multiple variables with the same name. It's then up to the variable cache (= us) to decide if they are identical

> Is there a difference in plain SMTLIB2 between nullary UFs and variables for parsing? Abd how do solvers handle each case?

In Smtlib nullary Ufs and variables are the same:
```
(declare-const f τ) has the same effect as the command (declare-fun f () τ)
```
(page 67, [here](https://smt-lib.org/papers/smt-lib-reference-v2.7-r2025-02-05.pdf#subsection.4.2.3))

In JavaSmt it's a bit more complicated as `declareUF` returns a `FunctionDeclaration` and not a `Formula`. However, we can guarantee that after applying the nullary UF the result is the same as when just declaring a variable

baierd:
Good step towards uniform behavior! Thank you!

> In Smtlib nullary Ufs and variables are the same
> However, we can guarantee that after applying the nullary UF the result is the same as when just declaring a variable

The important part here is that all solvers need the same behavior. So we need tests that check the cases:
- Create a 0-ary UF with some name and a variable with the same name (both directions) with equal and not equal types. Behavior here needs to be consistent.
- Are 0-ary UFs treated the same as variables for (basically) all operations? Especially quantifiers and visitor/rebuilding needs to be checked.

PhilippWendler:
What weighs higher?

1. If nullary UFs are supported: the inconsistency and potential surprise / risk of errors that when users create a nullary UF and then visit the formula, they do not see a UF but a variable.
2. If nullary UFs are forbidden: the inconsistency and potential hassle if users have code that does not care about argument count and simply want to use the same code/API for all arities of UFs (but note that even if nullary UFs are allowed to be created, these users would not be able to achieve their goal when visiting formulas), plus the backwards incompatibility for those solvers where JavaSMT already allowed nullary UFs.

To me the answer is not obvious, but I think this question is important and should be discussed deliberately before committing to one of the options.

kfriedberger:
There are several ways to go further:

#### 1. Visitor-Centric Approach
Modify the FormulaVisitor interface (by flag/option), such that the user gets one of both, either variable or nullary UF.
* How it would work: When `visit` is called on a variable, we provide a way to return it as a function application.
* Pros: Keeps the internal representation simple (just variables) but provides the necessary hook for the user.
* Cons: SMTLIB output does not use that flag/option, but uses the user-given kind.

#### 2. API-Enforcement Approach (Silent Conversion)
As implemented in this PR here: We convert nullary UFs to variables silently.
* How it would work: If a user calls declareUF with zero arguments, the API simply converts it into a variable.
* Pros: Good clarity, the user will see that a nullary UF is a variable in SMTLIB.
* Cons: Breaks backward compatibility (grey area with a long history) and might be bad for potential users that generate UFs dynamically with zero or more arguments (however, a fix on their side is small: check for number of arguments and create a variable, e.g., we even do that in JavaSMT automatically). And this way will ignore SMTLIB parser input.

#### 3. API-Enforcement Approach (Error on Nullary UF)
Nearly the same as approach 2, but with exceptions, i.e, we forbit `declareUF` to accept any empty list of arguments, forcing the use of makeVariable.
* How it would work: If a user calls declareUF with zero arguments, the API simply throws an exception, directing them to the variable API.
* Pros: Better clarity, the user will see that a nullary UF is a variable in SMTLIB.
* Cons: Breaks backward compatibility (grey area with a long history) and might be bad for potential users that generate UFs dynamically with zero or more arguments (however, a fix on their side is small: check for number of arguments and create a variable). And this way will ignore SMTLIB parser input.

#### 4. Solver-Level Normalization
Keep the frontend objects strictly separate (a UFApplication object vs. a Variable object) but unify them only at the moment they are passed to the native solver.
* How it would work: JavaSMT maintains two distinct mappings. When `f()` (UF) and `f` (Variable) are sent to the solver, we recognize them as identical and map to the same native object. When giving objects back to the user, convert them according to the mapping into variable or nullary UF.
* Pros: Best for the user; the Java-side structure remains exactly as defined.
* Cons: This might cause duplicate symbols/cache inconsistencies in solvers. Implementing this correctly requires a complex mapping. And this way will ignore SMTLIB parser input and SMTLIB output.

I am fine with approach 2. And I have not yet seen users that use nullary UFs in larger projects.

daniel-raffler:
Thanks for the summary! I'm merging this for now. We can still switch to a different approach if there are any problems

PR #581, opened 2026-01-14
Add better error messages for duplicate symbols

daniel-raffler:
Hello

this PR fixes issues with name collisions on CVC4/5 and Princess. None of these solvers support variable/Uf names to be overloaded, however, checks for name collisions were missing or broken. Because of this, duplicate symbols could be declared, but the solver would still crash when trying to use these new symbols. This PR fixes the checks and makes sure that a more user-friendly error message is printed when trying to redefine an existing symbol

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/princess/PrincessFunctionDeclaration.java
+++ b/src/org/sosy_lab/java_smt/solvers/princess/PrincessFunctionDeclaration.java
@@ -95,7 +95,7 @@ abstract class PrincessFunctionDeclaration {
       super(pApp.fun());
 
       ImmutableList.Builder<FormulaType<?>> builder = ImmutableList.builder();
-      for (int i = 0; i < pApp.args().length(); i++) {
+      for (int i = 0; i < pApp.fun().arity(); i++) {
         builder.add(PrincessEnvironment.getFormulaType(pApp.apply(i)));
       }
       argSorts = builder.build();


src/org/sosy_lab/java_smt/solvers/princess/PrincessFunctionDeclaration.java
```diff
@@ -79,27 +77,60 @@ public String toString() {
...
+        final FormulaType<?> actualType = PrincessEnvironment.getFormulaType(arg);
+        final FormulaType<?> expectedType = argSorts.get(i);
+
+        if (actualType.isBooleanType()) {
+          // boolean term -> build ITE(t,0,1)
           termArg =
               new ITermITE(
                   (IFormula) arg, new IIntLit(IdealInt.ZERO()), new IIntLit(IdealInt.ONE()));
-        } else if (!exprIsRational(arg) && functionTakesRational(i)) {
+        } else if (actualType.isIntegerType() && expectedType.isRationalType()) {
```
kfriedberger, line 133: interesting. This looks simpler than before. Maybe Princess improved its API in the last years. 😄 Does this really hold for all inputs?

daniel-raffler: No, the API didn't change. Princess uses `IFunction`s for function declarations and it's not always possible to get a type signature from them as they may be polymorphic. When declaring a UF symbol it's generally possible to cast the `IFunction` to a `MonoSortedIFunction` with a fixed type that can then be accessed. However, the cast doesn't always seem to work and sometimes the signature just isn't available. Because of that we're now storing the signature inside `PrincessIFunctionDeclaration`, which wraps around the actual `IFunction`. The function cache now also maps to `PrincessIFunctionDeclaration`, instead of `IFuntion`. This allows us to check for collisions when declaring a new UF symbol.. and it also has the side-effect of simplifying the code above as we can just compare the type of the arguments to the stored function type

kfriedberger: Thanks for the explanation. Question is resolved. PR can be merged.

Issue #558, opened 2025-12-11
Add Proof Support to API


baierd:
JavaSMT is lacking a common API for proofs of unsatisfiablity. This is partially due to basically no two solvers using the same proof format.

@gcarpio21 already worked quite a lot on supporting proofs in the formats of the solvers in JavaSMT, but also started working on a common proof format based on SMTInterpols format.
Since this work is currently squashed into a single branch, we should split the core-framework for proofs with a well working solver implementation for proofs into a new branch and focus on merging it to master before continuing. @gcarpio21 please choose an appropriate SMT solver for this and commit your work to branch `558-add-general-proof-api`.
Subsequently, we should add more solvers in their own branches.

To keep track of the progress:

- [ ] General proof API
- [ ] Test suite for proofs

Solvers with proof support that have been added to JavaSMT:
- [ ] Z3
- [ ] CVC5
- [ ] CVC4
- [ ] MathSAT5
- [ ] SMTInterpol
- [ ] Princess
- [ ] OpenSMT

PR #583, opened 2026-01-16
558 add general proof api

gcarpio21:
## Objective  
This PR aims to extend support for retrieving proofs of unsatisfiability from the CVC5 solver according to a general proof interface that would be used to do the same for other capable solvers. It is a step in closing the issue #558 .

## Key Changes  
- Introduced `Proof`, and `ProofRule` as the main components for handling proofs.  
- Implemented proof retrieval for the **CVC solver**.  
- Added multiple tests for retrieving proofs in edges cases and different theories. 

It is now possible to enable proof generating capabilities from CVC5 by passing  `ProverOptions.GENERATE_PROOFS` when creating a new `ProverEnvironment` instance. Then when a query returns UNSAT one can call `getProof()` to retrieve the proof of unsatisfiability.

### DIFF
--- a/src/org/sosy_lab/java_smt/api/proofs/Proof.java
+++ b/src/org/sosy_lab/java_smt/api/proofs/Proof.java
@@ -22,5 +22,5 @@ public interface Proof {
    * @return an object of type {@link ProofNode} the root proof node.
    * @see ProofNode
    */
-  public ProofNode getProofRoot();
+  ProofNode getProofRoot();
 }
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractProof.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractProof.java
@@ -10,14 +10,15 @@
 
 package org.sosy_lab.java_smt.basicimpl;
 
+import java.util.Objects;
 import org.sosy_lab.java_smt.api.proofs.Proof;
 import org.sosy_lab.java_smt.api.proofs.ProofNode;
 
-public class AbstractProof implements Proof {
-  public final ProofNode root;
+public abstract class AbstractProof implements Proof {
+  private final ProofNode root;
 
   public AbstractProof(ProofNode pRoot) {
-    root = pRoot;
+    root = Objects.requireNonNull(pRoot);
   }
 
   @Override


src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
```diff
@@ -160,6 +161,12 @@ default ImmutableMap<String, String> getStatistics() {
     return ImmutableMap.of();
   }
 
+  /**
+   * Get proof of unsatisfiability of the conjuction of the current satck of all formulas. Should
+   * only be called after {@link #isUnsat()} returned <code>true</code>.
```
baierd: I suggest the wording: ` This method should be called only immediately after an {@link #isUnsat()} call that returned <code>true</code>.`. We know that some solvers tend to throw information as the model or proofs away if you perform other operations in between `isUnsat()` and calling the model/proof API.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -52,6 +52,14 @@ enum ProverOptions {
      */
     GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS,
 
+    /**
+     * Whether the solver should generate proofs for unsatisfiable formulas. For Z3, this parameter
+     * is context(not solver)-based, so for setting it, the extra option in the {@link
+     * org.sosy_lab.common.configuration.Configuration} instance: "requireProofs" should be set to
```
baierd: We currently are not concerned with Z3. So we should (for now) remove the information about it.

src/org/sosy_lab/java_smt/api/proofs/Proof.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2024 Dirk Beyer <https://www.sosy-lab.org>
```
baierd: 2026 ;D
In all new files.

src/org/sosy_lab/java_smt/api/proofs/ProofFrame.java
```diff
@@ -0,0 +1,53 @@
...
+
+package org.sosy_lab.java_smt.api.proofs;
+
+/**
+ * Stores proof related information and gets stored in a stack when processing proofs. Each frame
+ * contains a proof object and the number of arguments it has.
+ *
+ * @param <T> The type of the proof object.
+ */
+public abstract class ProofFrame<T> {
```
baierd: Do i see it correctly that this class is only needed to generate the proof?

src/org/sosy_lab/java_smt/api/proofs/Proof.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.sosy_lab.java_smt.api.proofs;
+
+import java.util.Optional;
+import java.util.Set;
+import org.sosy_lab.java_smt.api.Formula;
+
+/** A proof node in the proof DAG of a proof. */
+public interface Proof {
```
baierd, line 17: It seems like the name of this class should be `ProofNode`.

src/org/sosy_lab/java_smt/basicimpl/AbstractProof.java
```diff
@@ -0,0 +1,110 @@
...
+  private final Set<Proof> children = new LinkedHashSet<>();
+  private ProofRule rule;
+  protected Optional<Formula> formula = Optional.empty();
+
+  protected AbstractProof(ProofRule rule, @Nullable Formula formula) {
+    this.rule = rule;
+    this.formula = Optional.ofNullable(formula);
+  }
+
+  // TODO: Use Optional instead of nullable
```
baierd: This TODO seems fulfilled.

src/org/sosy_lab/java_smt/basicimpl/AbstractProof.java
```diff
@@ -0,0 +1,110 @@
...
+public abstract class AbstractProof implements Proof {
+
+  // protected abstract class Transformation {
+  //  protected <TFormulaInfo, TType, TEnv, TFuncDecl, T> Transformation(
+  //      FormulaCreator<TFormulaInfo, TType, TEnv, TFuncDecl> formulaCreator, T proof) {}
+
+  //    protected abstract Proof generateProof();
+  // }
+
+  private final Set<Proof> children = new LinkedHashSet<>();
```
baierd: Please try using immutable data-structures as far as possible.
You can find a good guide as to why and when to use mutable and immutable data-structures in CPAcheckers style-guide [here](https://gitlab.com/sosy-lab/software/cpachecker/-/blob/main/doc/StyleGuide.md?ref_type=heads).

src/org/sosy_lab/java_smt/basicimpl/AbstractProof.java
```diff
@@ -0,0 +1,110 @@
...
+import org.sosy_lab.java_smt.api.proofs.ProofRule;
+
+/**
+ * A proof DAG of a proof.
+ *
+ * @author Gabriel Carpio
+ */
+public abstract class AbstractProof implements Proof {
+
+  // protected abstract class Transformation {
```
baierd: Is this commented out code still needed?

baierd (changes requested):
Partial review. I will continue tomorrow.

src/org/sosy_lab/java_smt/api/proofs/ProofFrame.java
gcarpio21: Yes, this is wrapper that helps process the proofs, I did not want to repeat code over all solvers so I put the class in this directory.

gcarpio21: I moved this class inside `AbstractProof` I think this better fits the structure of the project. I was thinking of taking the proof interfaces out of the `proofs` package into the `api` one for the same reason.

src/org/sosy_lab/java_smt/api/proofs/Proof.java
baierd: Note: this was named `ProofNode` in the past and was changed due to me requesting it.

In general: the `Proof` interface/object should hold all proofs themselves, and provide API to access them (similar to the model). The problem with this being called `Proof` can be seen in the API, e.g. `  ImmutableSet<Proof> getChildren();`. If `getChildren()` returns a `Proof`, it is clear that there is a problem, as the `Proof` should only exist as one central element. Hence i was wrong in wanting to rename this interface to `Proof`, and it should be called `ProofNode` as @gcarpio21 named it correctly initially! 

We should then add a new interface `Proof`, that has the API to return the root (and only ever the root) of the proof DAG (and potentially more in the future).

src/org/sosy_lab/java_smt/api/proofs/ProofRule.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.sosy_lab.java_smt.api.proofs;
+
+/** A proof rule from a given proof format. */
+public interface ProofRule {
+
+  /** Get the name of the proof rule. */
```
baierd, line 29: An example would be helpful for users to understand. Especially since distinct solvers can return distinct rules.

src/org/sosy_lab/java_smt/api/SolverContext.java
```diff
@@ -52,6 +52,9 @@ enum ProverOptions {
      */
     GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS,
 
+    /** Whether the solver should generate proofs for unsatisfiable formulas. */
```
baierd, line 55: Could you add a note about Z3 here please? The Z3 proof situation might be confusing for users.

src/org/sosy_lab/java_smt/test/ProverEnvironmentTest.java
```diff
@@ -189,4 +203,340 @@ private void checkSimpleQuery(ProverEnvironment pProver)
...
+      prover.addConstraint(query);
+      assertThat(prover.isUnsat()).isTrue();
+
+      // Test getProof()
+      Proof proof = prover.getProof();
+      checkProof(proof);
+    }
+  }
+
+  // Traverses a proof and asserts that certain values are not null, instances, etc.
```
baierd: Looks like the name could be improved into something like `varifyProofObjectValidity`. (`checkProof` might be misunderstood such that the proof itself is checked for validity, but you are only checking the objects inside.) Also JavaDoc with more information would be helpful here.

src/org/sosy_lab/java_smt/api/SolverContext.java
gcarpio21: You mean about it needing proof generation enabled on the context object? 

I was thinking to leave that out for this, as the only solver with proof generation should be CVC5 for now. Once it is turn for Z3 then I would add the corresponding note.

src/org/sosy_lab/java_smt/api/proofs/Proof.java
gcarpio21: This has been renamed and now there is a new `Proof` interface, which allows retrieval of the root node

PR #584, opened 2026-01-19
Changes to the FormulaManager.equal and FormulaManager.distinct API

daniel-raffler:
As suggested by @PhilippWendler:
- We now allow `Iterables` as arguments for `FormulaManager.equal` and `FormulaManager.distinct`
- The methods have been renamed to `makeEqual` and `makeDistinct`
- We use `FluentIterable` to avoid unnecessary copying

See #559 for the discussion

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
@@ -267,6 +267,7 @@ public abstract class AbstractFormulaManager<TFormulaInfo, TType, TEnv, TFuncDec
   }
 
   /** Override if the solver API only supports binary equality. */
+  @SuppressWarnings("unused")
   protected TFormulaInfo equalImpl(TFormulaInfo pArg1, TFormulaInfo pArgs) {
     throw new UnsupportedOperationException(
         "equalImpl(x, y) must be implemented in a subclass, if required.");
@@ -307,6 +308,7 @@ public abstract class AbstractFormulaManager<TFormulaInfo, TType, TEnv, TFuncDec
     return formulaCreator.encapsulateBoolean(distinctImpl(args));
   }
 
+  @SuppressWarnings("unchecked")
   private FluentIterable<TFormulaInfo> getTFormulaInfosForComparison(Iterable<Formula> pArgs) {
     final ImmutableSet<FormulaType<Formula>> types =
         FluentIterable.from(pArgs).transform(formulaCreator::getFormulaType).toSet();


src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -140,8 +139,8 @@ <T extends Formula> T makeApplication(
    * @param pArgs Arguments to be compared for equality, ordering does not matter.
    * @return Equality formula
    */
-  default BooleanFormula equal(Formula... pArgs) {
-    return equal(ImmutableList.copyOf(pArgs));
+  default BooleanFormula makeEqual(Formula... pArgs) {
+    return makeEqual(FluentIterable.from(pArgs));
```
PhilippWendler: I would simply use `Arrays.asList` here. `FluentIterable` has no benefit, and just creates one more object for wrapping the array.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -273,19 +273,19 @@ protected TFormulaInfo equalImpl(TFormulaInfo pArg1, TFormulaInfo pArgs) {
   }
 
   @Override
-  public BooleanFormula equal(Collection<Formula> pArgs) {
-    if (pArgs.size() < 2) {
+  public BooleanFormula makeEqual(Iterable<Formula> pArgs) {
+    final Collection<Formula> args = ImmutableList.copyOf(pArgs);
+    if (args.size() < 2) {
```
PhilippWendler: This change doesn't make sense. If you really need a size, it is better to require `Collection`. Right now you force an unnecessary eager copy on every caller, even if they already have a `Collection` (as long as it does not happen to be an `ImmutableCollection`). If you would require `Collection`, only callers that do not have one need to copy.

I was thinking that you do not need the size and can accept `Iterable` at no cost. Checking whether the input is empty or has one element does not need the size. And it could anyway be done after the copy that is already made later on in the method.

src/org/sosy_lab/java_smt/api/FormulaManager.java
daniel-raffler: Done

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
daniel-raffler: It's possible to remove the check entirely, see https://github.com/sosy-lab/java-smt/pull/584/commits/7009c867489aafce3f04945380ab0442999e4ecb. However, we then have to handle the special case for each solver separately

Should I keep it this way, or revert to `Collection` for the argument type?

PhilippWendler: That is up to the maintainers to decide. It just wanted to say that it does not make sense to force an eager copy of an `Iterable` if using `Collection` would get rid of the copy, in particular given that there was yet another eager copy made.

Either requiring a `Collection` with no eager copy, or allowing `Iterable` with just one eager copy, or allowing `Iterable` with no eager copy (and decentralized length checks) might be meaningful options. And there is also the alternative to check whether the `Iterable` has more than one element by using the iterator manually.

baierd: Do we really need this check? Can't we delegated this check to the equals methods of the specialized formula managers?

Edit: meaning we could use the existing type check for every solver at every equals. Maybe even use a new `equalImplWithTypeCheck()`that uses this type check on every pair, so that it is not done for the safe, normal use of `makeEqual()`.

daniel-raffler: > Do we really need this check? Can't we delegated this check to the equals methods of the specialized formula managers?
>
> Edit: meaning we could use the existing type check for every solver at every equals. Maybe even use a new `equalImplWithTypeCheck()`that uses this type check on every pair, so that it is not done for the safe, normal use of `makeEqual()`.

Do you mean the check down here?
```
 Preconditions.checkArgument(
        ImmutableSet.copyOf(types).size() == 1,
        "All arguments to `equal` must have the same type, but found %s different types: %s",
        types.size(),
        types);
```

We could delegate for `makeEqual`, but some of the `FormulaManagers` only allow two arguments. Delegating also won't work for `makeDistinct`, as most `FormulaManagers` don't implement the operation at all. It's possible to use the default implementation and rewrite `distinct` into a formula that only uses `and`, `not` and (binary) equality. However, this leads to a quadratic number of terms, which is why it's generally better to use `distinct` directly

The problem with the `args.size() < 2` check is that `Iterable` doesn't give us any easy way to check that `makeEqual`/`makeDistinct` was called with at least two arguments. That means we can't decide if we're in the special case, or not. We could of course use the iterator to count ourselves:
```Java
  private <T> boolean hasAtLeastTwo(Iterable<T> pArgs) {
    var it = pArgs.iterator();
    if (it.hasNext()) {
      it.next();
      if (it.hasNext()) {
        return true;
      }
    }
    return false;
  }
```
However, that seems a little awkward, and I believe it would be simpler to make the argument a `Collection` again, and then call `size()` on it

daniel-raffler: On a side-node: The type check may be a little too strict as we allow mixed integer-real arguments in `RationalFormulaManager.equal`, but not here. That shouldn't be too hard to fix, though

kfriedberger: fixed.

kfriedberger (approved):
looks valid now.

PR #585, opened 2026-01-20
Close memory leaks on CVC5 and Bitwuzla

daniel-raffler:
Hello,

this PR aims to fix memory leaks on CVC5 and Bitwuzla
* **CVC5** We now use `Context.deletePointers` to make sure all solver memory is freed. `Context` is a utility class in the CVC5 JNI bindings that tracks all solver objects. The `deletePointers` method can then be called at the end of the program to free native memory. It is not tied to a `TermManager` and will delete Objects from all remaining CVC5 SolverContexts, which is why it can only be used when closing the last solver instance.
We used to only close `Solver` and `TermManager` when a CVC5 SolverContext is closed. While this approach is more fine-grained, it also leads to memory leaks as Terms and Sorts are never freed, even after the `SolverContext` is closed. Our new solution is to check in `SolverContext.close()` if we're closing the last CVC5 instance and then delete all Objects with `deletePointers`. Otherwise only the `Solver` object is deleted right away, and all other Objects will be collected at the end of the Program. (Thanks to @baierd for the suggestion)
* **Bitwuzla** We adopt a very similar approach to CVC5. Unlike CVC5, Bitwuzla does not come with its own Java bindings, and we use SWIG to generate our own JNI headers. SWIG uses finalizers to release native memory when it is no longer needed. This allows for more fine-grained control, but is also error-prone, and we had to disable several of the finalizers to avoid random crashes in Bitwuzla. This lead to memory leaks, which have now been closed by tracking all solver objects, similar to how it is done in CVC5

Here are some benchmarks to show that the leaks have indeed been closed. The program executes the same JavaSMT trace in a loop while monitoring total memory usage of the process:

**CVC5** (before)
cvc5-0

**CVC5** (after)
cvc5-1

**Bitwuzla** (before)
bitwuzla-0

**Bitwuzla** (after)
bitwuzla-1

@kfriedberger
We will need to recompile the Bitwuzla bindings for this PR. Could you handle this, please?

### DIFF
--- a/lib/ivy.xml
+++ b/lib/ivy.xml
@@ -5,7 +5,7 @@ This file is part of JavaSMT,
 an API wrapper for a collection of SMT solvers:
 https://github.com/sosy-lab/java-smt
 
-SPDX-FileCopyrightText: 2023 Dirk Beyer <https://www.sosy-lab.org>
+SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
 
 SPDX-License-Identifier: Apache-2.0
 -->
@@ -196,7 +196,7 @@ SPDX-License-Identifier: Apache-2.0
         <dependency org="org.sosy_lab" name="javasmt-solver-opensmt" rev="2.9.0-gef441e1c" conf="runtime-opensmt-x64->solver-opensmt-x64; runtime-opensmt-arm64->solver-opensmt-arm64; contrib->sources,javadoc"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-optimathsat" rev="1.7.3-sosy1" conf="runtime-optimathsat->solver-optimathsat" />
         <dependency org="org.sosy_lab" name="javasmt-solver-cvc4" rev="1.8-prerelease-2020-06-24-g7825d8f28" conf="runtime-cvc4->solver-cvc4" />
-        <dependency org="org.sosy_lab" name="javasmt-solver-cvc5" rev="2025-11-13-ef82847" conf="runtime-cvc5-x64->solver-cvc5-x64; runtime-cvc5-arm64->solver-cvc5-arm64"/>
+        <dependency org="org.sosy_lab" name="javasmt-solver-cvc5" rev="2026-01-21-f621457" conf="runtime-cvc5-x64->solver-cvc5-x64; runtime-cvc5-arm64->solver-cvc5-arm64"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-boolector" rev="3.2.2-g1a89c229" conf="runtime-boolector->solver-boolector" />
         <dependency org="org.sosy_lab" name="javasmt-solver-bitwuzla" rev="0.8.2-108-g86432ccd" conf="runtime-bitwuzla-x64->solver-bitwuzla-x64; runtime-bitwuzla-arm64->solver-bitwuzla-arm64; contrib->sources,javadoc"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-z3-legacy" rev="4.5.0-gd57a2a6dc" conf="runtime-z3-legacy-x64->solver-z3-legacy-x64; runtime-z3-legacy-arm64->solver-z3-legacy-arm64" />
--- a/src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/api/FloatingPointFormulaManager.java
@@ -2,14 +2,14 @@
 // an API wrapper for a collection of SMT solvers:
 // https://github.com/sosy-lab/java-smt
 //
-// SPDX-FileCopyrightText: 2025 Dirk Beyer <https://www.sosy-lab.org>
+// SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
 //
 // SPDX-License-Identifier: Apache-2.0
 
 package org.sosy_lab.java_smt.api;
 
 import static org.sosy_lab.java_smt.api.FormulaManager.API_METHOD_NOT_IMPLEMENTED;
-import static org.sosy_lab.java_smt.api.FormulaType.getFloatingPointType;
+import static org.sosy_lab.java_smt.api.FormulaType.getFloatingPointTypeFromSizesWithoutHiddenBit;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -113,7 +113,8 @@ public interface FloatingPointFormulaManager {
         number.getExponent(),
         number.getMantissa(),
         number.getMathSign(),
-        getFloatingPointType(number.getExponentSize(), number.getMantissaSize()));
+        getFloatingPointTypeFromSizesWithoutHiddenBit(
+            number.getExponentSize(), number.getMantissaSizeWithoutHiddenBit()));
   }
 
   /**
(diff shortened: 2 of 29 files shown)


lib/native/source/libbitwuzla/src/org/sosy_lab/java_smt/solvers/bitwuzla/api/TermManager.java
```diff
@@ -16,44 +16,48 @@
...
     }
-    return ptr;
   }
 
-  @SuppressWarnings({"deprecation", "removal"})
-  protected void finalize() {
-    delete();
+  static void removeReference(Reference pointer) {
+    synchronized (TermManager.class) {
+      references.removeLastOccurrence(pointer);
```
kfriedberger, line 47: This does a partial iteration of the list. Can we reduce this to O(1) by using a Set?

I see that in combination with the method "deleteReferences()" from below, this is not expensive, because both methods iterate in same order (aka reverse order) and the elements are removed from the deque directly after access. If this is the only way of accessing the method "removeReference(ptr)", we can keep it, but documentation is required.

lib/native/source/libbitwuzla/src/org/sosy_lab/java_smt/solvers/bitwuzla/api/TermManager.java
```diff
@@ -16,44 +16,48 @@
...
-      ptr = obj.swigCPtr;
-      obj.swigCMemOwn = false;
-      obj.delete();
+  @Override
+  void deleteCPtr() {
+    BitwuzlaNativeJNI.delete_TermManager(swigCPtr);
+  }
+
+  static void addReference(Reference pointer) {
+    synchronized (TermManager.class) {
```
kfriedberger, line 40: Why synchronise on "TermMgr.class" instead of the field "references"?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaNativeApiTest.java
```diff
@@ -92,21 +90,6 @@ public void boolType() {
...
-    assertThat(tru12.is_true()).isTrue();
-
-    new Thread(
-            () -> {
-              assertThat(tru1.is_true()).isTrue();
-              assertThat(tru12.is_true()).isTrue();
-            })
-        .start();
-  }
-
```
kfriedberger, line 109: Why was this test removed?
- Was it successful? -> keep it.
- Did it crash? -> Add negative test.
- Did it crash too hard? -> Where is the documentation?

kfriedberger (approved):
Well done. Thanks for the analysis and solution.

I will try to find some time soon for updating the solver library.

Side note: Which tool was used to generate the graphs for measurements? JProfiler or similar?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaNativeApiTest.java
daniel-raffler: There is a short comment in the commit message:
```
We would have to wait for the thread to make sure it's not still running when the TermManager is destroyed by the @After method. Since the test seems to be about an old bug that has since been fixed, I'm simply removing it
```
The problem is that `@After` method destroys all solver objects while the thread from the test might still be running. This then leads to random segfaults. I removed the test mostly because I'm not sure what it's meant to be testing. However, it's not too hard to fix it and I can add it again later today

baierd: We could just as the Bitwuzla devs about this.

lib/native/source/libbitwuzla/src/org/sosy_lab/java_smt/solvers/bitwuzla/api/TermManager.java
daniel-raffler: > This does a partial iteration of the list. Can we reduce this to O(1) by using a Set?

Yes, but the Set would have to preserve the insertion order. Most of the code here is taken from CVC5, which uses a linked Map to keep track of the references. I've only switched to a List recently, and this seems to have improved performance, while also saving some memory. Note that the list can get very large (> 400000) quickly, so it's important to not waste too much additional memory for each reference

> I see that in combination with the method "deleteReferences()" from below, this is not expensive, because both methods iterate in same order (aka reverse order) and the elements are removed from the deque directly after access. If this is the only way of accessing the method "removeReference(ptr)", we can keep it, but documentation is required.

Yes, that's the idea! Normally `removeReference` will only be called through `deleteReferences`, and this makes the operation effectively O(1). There is only one exception, which is closing a `ProverEnvironment`. In this case the reference is deleted directly and we have to go though the entire list looking for it. However, `ProverEnvironments` are only opened/closed very rarely, so this shouldn't cause any performance issues

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaNativeApiTest.java
daniel-raffler: The test was originally added in 1b5167b04d10781fb81fd9744c5fac6a48e1fd79 and seems to be about garbage management in the SWIG bindings. It's set to `@Ignore` , so I assume that it didn't work at the time. In the test there is a list to store old Bitwuzla environments and keep them from being garbage collected. It may still crash, though, as the thread that is started may run after the test has already completed, and the list itself was garbage collected. Here is a fixed version of the test that simply waits for the thread:
```java
  @Test
  public void repeatedTermCreationInMultipleSolversTest() throws InterruptedException {
    Term tru1 = termManager.mk_true();
    Term tru12 = termManager.mk_true();
    assertThat(tru1.is_true()).isTrue();
    assertThat(tru12.is_true()).isTrue();

    Thread thread = new Thread(
            () -> {
              assertThat(tru1.is_true()).isTrue();
              assertThat(tru12.is_true()).isTrue();
            });
    thread.start();
    thread.join();
  }
```
Should I re-add this test, or can we just drop it?

daniel-raffler:
> Well done. Thanks for the analysis and solution.

Thanks!

> Side note: Which tool was used to generate the graphs for measurements? JProfiler or similar?

I used [procpath](https://procpath.readthedocs.io/en/latest/) for the graphs and wrote a small program to repeatably run the same JavaSMT trace so that we can monitor its memory usage. [Here](https://github.com/user-attachments/files/24764073/MemoryLeak.zip) is the source for the tool. If needed, this could also be integrated into JavaSMT for benchmarking. The other approach is to use CPA-Daemon to look for memory leaks, but then it's not entirely clear if the problem is with JavaSMT or CPAchecker. Profilers weren't too useful as they can only see the Java heap, and not what is going on in native memory when using JNI

daniel-raffler:
> Our new solution is to check in `SolverContext.close()` if we're closing the last CVC5 instance and then delete all Objects with `deletePointers`. Otherwise only the `Solver` object is deleted right away, and all other Objects will be collected at the end of the Program

One more note: This solution works well if there is only a single `SolverContext`, or when only using `SolverContexts` sequentially (i.e closing the last `SolverContext` before opening a new one). It might work less well in a parallel setting when `SolverContexts` keep being opened and closed in parallel. While the code should be thread-safe, memory will only be freed whenever we get to a point where all `SolverContexts` are closed. If there are multiple busy threads this might not happen until the end of the program, and memory will still pile up in the mean time

We could fix this by managing references for each `TermManager`, rather than using a global queue. However, this adds additional overhead, and for CVC5 we would have to ask the developers for a fix as we don't use our own JNI bindings

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaNativeApiTest.java
kfriedberger: ok, lets just drop it.

lib/native/source/libbitwuzla/src/org/sosy_lab/java_smt/solvers/bitwuzla/api/TermManager.java
kfriedberger: both would work equally, just that "references" is a basic type and cannot be used for synchronization.

PR #513, opened 2025-09-02
Fix Floating-Point Mantissa Not Including Hidden Bit

baierd:
While implementing [PR 512](https://github.com/sosy-lab/java-smt/pull/512) i found a problem in our implementation of FloatingPointType and number. Our returned significant precision is off by one, see [SMTLib2 standard page for Floating-Points](https://smt-lib.org/theories-FloatingPoint.shtml). E.g. for single precision FP, we return significant size 23. But the standard defines exponent and significant as:
```
eb defines the number of bits in the exponent;
sb defines the number of bits in the significand, *including* the hidden bit.
```
Hence it should be 24.
As a consequence, we also return this wrongly for `FormulaType` method `toSMTLIBString()`.
This is fixed in this PR. 
We also encode the significant without the hidden bit in  `toString()`, and subsequently in `fromString`.
While this is in general not as bad, i changed it to include the hidden bit now (in both), so that it is in-line with the standard.
These behavior changes should be communicated clearly to users!

The solvers generally expect the hidden bit to be part of the significant, hence why we had 4 solvers with the code `type.getMantissaSize() + 1`, and only one solver without the `+ 1` (MathSAT5).

The general behavior of FP creation/solving is correct IFF you know that the mantissa does not include the hidden bit.

Also, we don't communicate any information about our interpretation of FP precisions, hence the users might think that by building `FormulaType.getFloatingPointType(8, 24)` (first input is exponent, second is significant) they get a single precision FP-type, but in reality they get a total size of 33 bits currently. Our static implementation of the single/double precisions is fine however.

This PR adds/changes the following:
- Add new constructors for FP type that specify the hidden bit inclusion in the method name. Also, update the API for getMantissaSize() to getMantissaSizeWithHiddenBit() and getMantissaSizeWithoutHiddenBit().
- Use the less ambiguous API in our internal implementation, reducing confusion about magic `+-1`
- Including the hidden bit in the documentation of relevant API, as well as the parameter names.
- Deprecate the old FP type constructor and mantissa size getters.
- Return the correct SMTLIB2 interpretation for `FloatingPointType.toSMTLIBString()`.
- Change behavior of `FloatingPointType.toString()` and `FloatingPointType.fromString()` to include the hidden bit in the significant.
- Deprecated public constants for FP single/double precisions, as they do not specify the in/exclusion of the hidden bit and will be moved to internal API. Also added new constants that are not public for mantissas that specify the hidden bit exclusion.
[...]

PR #588, opened 2026-01-23
Fix Z3 Legacy Floating-Point Theory

baierd:
The old Z3 version does not support FPs currently. Some problems are solved by #513, all remaining problems are found in the retranslation of SMT model values to Java values. There are 3 main problems that are solved, 2 of wich are not optimally solved for now (1 & 2) below:
1. NaN could not be detected, but crashed the normal evaluation. This is now detected based on the unique String representation of NaN.
2. +/-Infinity could not be detected, but crashed the normal evaluation. This is now detected based on the unique String representation of +/- inifity.
3. Normal FP values need to be built by their exponent and significand String representations pulled from the AST string. None other representation seems to be accurate somehow. Zeros need to be detected before this however, but thats easy.

This branch is based on #513 and should only be merged after #513.

### DIFF
--- a/lib/ivy.xml
+++ b/lib/ivy.xml
@@ -5,7 +5,7 @@ This file is part of JavaSMT,
 an API wrapper for a collection of SMT solvers:
 https://github.com/sosy-lab/java-smt
 
-SPDX-FileCopyrightText: 2023 Dirk Beyer <https://www.sosy-lab.org>
+SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
 
 SPDX-License-Identifier: Apache-2.0
 -->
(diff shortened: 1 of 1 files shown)


baierd:
Examples for calling the FP managers `makeNumber()` with single precisions and output of the model value of the constant based on multiple API calls. The exponents string representation is called: exp string, while the mantissas string representation is: mant string.

`astToString()` seems to only return the type with the names of special values or the bit values of sign, exponent, and mantissa. 

`getNumeralString()` seems to be only the names of special values or the bit values of sign, exponent, and mantissa.

```
fpmgr.makeNumber(0.0, singlePrecType)
astToString:      (_ +zero 8 24)
getNumeralString: +zero
exp string:       0   mant string:  1

fpmgr.makeNumber(1.0, singlePrecType)
astToString:      (fp #b0 #x7f #b00000000000000000000000)
getNumeralString: 1.0 0
exp string:       0   mant string:  1

fpmgr.makeNumber(2.0, singlePrecType)
astToString:      (fp #b0 #x80 #b00000000000000000000000)
getNumeralString: 1.0 1
exp string:       1   mant string:  1

fpmgr.makeNumber(3.0, singlePrecType)
astToString:      (fp #b0 #x80 #b10000000000000000000000)
getNumeralString: 1.5 1
exp string:       1   mant string:  1.5
```

We might be able to detect zero based on the string representation.

Edit: just using the hex/binary representation of the AST after filtering out zeros works fine.

PR #590, opened 2026-01-26
Intelij: Update copyright template

daniel-raffler:
Updates the copyright template to always use the current year when creating a new file. The old template was still using 2024 in the copyright header

### DIFF
--- a/.idea/copyright/JavaSMT.xml
+++ b/.idea/copyright/JavaSMT.xml
@@ -10,7 +10,7 @@ SPDX-License-Identifier: Apache-2.0
 
 <component name="CopyrightManager">
   <copyright>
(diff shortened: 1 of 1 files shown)


daniel-raffler:
Hello Karlheinz

> Thanks for the fix. Could you also check the Eclipse-template?

I just tried, and Eclipse already seems to use the right date for the header when creating a new class. The template appears to come from `.settings/org.eclipse.jdt.ui.prefs` and looks like this:

Screenshot From 2026-01-28 10-37-14

It says `${year}` in the copyright header, so this should update automatically

> We should (from time to time) synchronize these templates from [SoSy-Lab project-template repository](https://gitlab.com/sosy-lab/software/java-project-template). I do not know whether we already documented that step anywhere in JavaSMT.

The Intelij project files appear to be specific to JavaSMT, and are not included in `java-project-template`. This might just be an omission as the template does include an Eclipse project. However, I'm not sure how easy it would be to add a generic Intelij project to `java-project-template`

I also ran `syncTo.sh`, but didn't see too many changes since the last update

PR #594, opened 2026-02-10
Refactor `FloatingPointNumber`

daniel-raffler:
Hello, 

after https://github.com/sosy-lab/java-smt/pull/513 the only way to create a `FloatingPointNumber` is to use `FloatingPointNumber.of(String, FloatingPointType)` or `FloatingPointNumber.of(Sign, BigInteger, BigInteger, FloatingPointType)`. Both  methods accept a `FloatingPointType` for the exponent and mantissa size.  Versions of `FloatingPointNumber.of(...)` that accept mantissa and exponent size as separate integer arguments have been deprecated as it was not clear if the hidden bit should be included in the mantissa size, or not

However, when reading out the mantissa/exponent size of there is no way to get the `FloatingPointType` from an existing `FoatingPointNumber`. Instead, the sizes have to be accessed separately though `FloatingPointNumber.getExponentSize`and `FloatingPointNumber.getMantissaSizeWith(out)HiddenBit`, which can be  error prone:
```java
public FloatingPointType getFloatingPointType(FloatingPointNumber pFloat) {
    return FormulaType.getFloatingPointTypeFromSizesWithoutHiddenBit(
        pFloat.getExponentSize(), pFloat.getMantissaSizeWithHiddenBit()); // there is a bug here
  }
```

In this PR we try to fix the issue by adding a new method `FloatingPointNumber.getFloatingPointType()` to get the `FloatingPointType` of a `FloatingPointNumber`. We also change the implementation, so that the `FloatingPointType` is stored directly when creating a new `FloatingPointNumber`, and add some additional sanity checks for the sizes

### DIFF
--- a/src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
+++ b/src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
@@ -269,6 +269,8 @@ public abstract class FloatingPointNumber {
   public static FloatingPointNumber of(String bits, FloatingPointType floatingPointType) {
     // Note: sign bit + exponent size + mantissa size without hidden bit == exponent size + mantissa
     // size with hidden bit
+    checkNotNull(bits);
+    checkNotNull(floatingPointType);
     var exponentSize = floatingPointType.getExponentSize();
     var mantissaSizeWithoutHiddenBit = floatingPointType.getMantissaSizeWithoutHiddenBit();
     Preconditions.checkArgument(


src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -158,24 +165,15 @@ public int getTotalSize() {
...
-              + " getFloatingPointTypeFromSizesWithoutHiddenBit(exponentSize,"
-              + " mantissaSizeWithoutHiddenBit))",
-      imports = {
-        "org.sosy_lab.java_smt.api.FloatingPointNumber",
-        "org.sosy_lab.java_smt.api.FloatingPointNumber.Sign",
-      },
-      staticImports =
-          "org.sosy_lab.java_smt.api.FormulaType"
-              + ".getFloatingPointTypeFromSizesWithoutHiddenBit")
+  @SuppressWarnings("InlineMeSuggester")
```
baierd: Why is inline me not working anymore? Sometimes its a little tricky to find the correct expression for it.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -199,18 +197,21 @@ public static FloatingPointNumber of(
   @Deprecated(
       since = "2026.01, because mantissa arguments with/without sign bits can be misleading",
       forRemoval = true)
+  @SuppressWarnings("InlineMeSuggester")
```
baierd: Why is inline me not working anymore? Sometimes its a little tricky to find the correct expression for it.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -226,15 +227,17 @@ public static FloatingPointNumber of(
...
-    Preconditions.checkArgument(exponent.compareTo(BigInteger.ZERO) >= 0);
-    Preconditions.checkArgument(mantissa.compareTo(BigInteger.ZERO) >= 0);
-    return new AutoValue_FloatingPointNumber(
-        sign,
-        exponent,
-        mantissa,
-        floatingPointType.getExponentSize(),
-        floatingPointType.getMantissaSizeWithoutHiddenBit());
+    Preconditions.checkArgument(
+        exponent.compareTo(BigInteger.ZERO) >= 0, "Exponent must not be negative");
```
baierd: You can use the BigInteger signum for this.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -249,10 +252,10 @@ public static FloatingPointNumber of(
   @Deprecated(
       since = "2026.01, because mantissa arguments with/without sign bits can be misleading",
       forRemoval = true)
+  @SuppressWarnings("InlineMeSuggester")
```
baierd: Why is inline me not working anymore? Sometimes its a little tricky to find the correct expression for it.

src/org/sosy_lab/java_smt/api/PackageSanityTest.java
```diff
@@ -17,6 +17,10 @@ public class PackageSanityTest extends AbstractPackageSanityTests {
 
   {
     setDistinctValues(FormulaType.class, FormulaType.BooleanType, FormulaType.IntegerType);
+    setDistinctValues(
+        FloatingPointType.class,
+        FormulaType.getSinglePrecisionFloatingPointType(),
+        FormulaType.getDoublePrecisionFloatingPointType());
     setDefault(FloatingPointType.class, getSinglePrecisionFloatingPointType());
```
baierd: Seems like we set this twice now?

src/org/sosy_lab/java_smt/api/FormulaType.java
```diff
@@ -270,6 +275,9 @@ public static FloatingPointType getFloatingPointTypeFromSizesWithoutHiddenBit(
    */
   public static FloatingPointType getFloatingPointTypeFromSizesWithHiddenBit(
       int exponentSize, int mantissaSizeWithHiddenBit) {
+    checkArgument(
```
baierd, line 278: Maybe we should stick to the [standards](https://smt-lib.org/theories-FloatingPoint.shtml) wording here, in that it should be "greather than 1".

From:
```
 :sorts-description "All nullary sort symbols of the form 
 
    (_ FloatingPoint eb sb),

  where eb and sb are numerals greater than 1."
```

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -267,11 +270,13 @@ public static FloatingPointNumber of(
    *     based on this type.
    */
   public static FloatingPointNumber of(String bits, FloatingPointType floatingPointType) {
-    checkNotNull(floatingPointType);
     // Note: sign bit + exponent size + mantissa size without hidden bit == exponent size + mantissa
     // size with hidden bit
     var exponentSize = floatingPointType.getExponentSize();
     var mantissaSizeWithoutHiddenBit = floatingPointType.getMantissaSizeWithoutHiddenBit();
+    Preconditions.checkArgument(exponentSize >= 0, "Exponent size must not be negative");
```
baierd: We should check this only once, and that location is clearly the type (so that we are failing early). This also applies to the other duplicate checks in this class.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -226,15 +227,17 @@ public static FloatingPointNumber of(
...
-        exponent,
-        mantissa,
-        floatingPointType.getExponentSize(),
-        floatingPointType.getMantissaSizeWithoutHiddenBit());
+    Preconditions.checkArgument(
+        exponent.compareTo(BigInteger.ZERO) >= 0, "Exponent must not be negative");
+    Preconditions.checkArgument(
+        mantissa.compareTo(BigInteger.ZERO) >= 0, "Mantissa must not be negative");
+    Preconditions.checkArgument(
+        exponent.bitLength() <= floatingPointType.getExponentSize(),
```
baierd, line 233: Shouldn't this be a `==`?

baierd:
@kfriedberger is this change/idea OK for you?

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: I removed this one because the user is supposed to choose between `getFloatingPointTypeFromSizesWithHiddenBit` and `getFloatingPointTypeFromSizesWithoutHiddenBit` for the replacement

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: It wasn't working before 😉 These warnings showed up after I removed some redundant checks and the method was just a single expression. Now that some null checks have been added again the `@SuppressWarnings` isn't actually needed, and I removed it again

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: Thanks for the hint!

src/org/sosy_lab/java_smt/api/FormulaType.java
daniel-raffler: Thanks, I've updated the error messages

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: Thanks, the extra check has been removed

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: No, I think this should be fine. The exponent may have leading zeroes that are not counted by `BigInteger.bitLength`

src/org/sosy_lab/java_smt/api/PackageSanityTest.java
daniel-raffler: Yes, you're right. I've removed the `setDefault`

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: We could comment this so that we remember.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
```diff
@@ -267,7 +265,6 @@ public static FloatingPointNumber of(
    *     based on this type.
    */
   public static FloatingPointNumber of(String bits, FloatingPointType floatingPointType) {
-    checkNotNull(floatingPointType);
```
baierd, line 270: Seems like this could be useful still.

src/org/sosy_lab/java_smt/api/FormulaType.java
baierd: Thank you!

baierd: I just noticed, we are now doing this twice with your addition in the constructor.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: I've added a short comment

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
daniel-raffler: Both `bits` and `floatingPointType` are accessed in the following lines, so the NPE is thrown either way. However, I added the check back in for better documentation

src/org/sosy_lab/java_smt/api/FormulaType.java
daniel-raffler: > I just noticed, we are now doing this twice with your addition in the constructor.

Hmm... I think they're checking different things:
- In `FloatingPointNumber.of(Sign, BigInteger, BigInteger, FloatingPointType)` we check the values. That is `exponent` and `mantissa` must be "unsigned", and their bit width must not exceed the size of the type
- In `FloatingPointType.getFloatingPointTypeFromSizes...` we check the type. That is the mantissa and the exponent field must be at least 2 bits wide

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: By definition the chosen method must be `getFloatingPointTypeFromSizesWithoutHiddenBit` here. As that is the direct replacement of the old code.

But it seems like InlineMe does not allow nested function calls with multiple arguments. So it is not possible.

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: See: https://github.com/sosy-lab/java-smt/pull/594#discussion_r2867936962

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: See: https://github.com/sosy-lab/java-smt/pull/594#discussion_r2867936962

src/org/sosy_lab/java_smt/api/PackageSanityTest.java
baierd: Thanks!

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: Thank you!

src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
baierd: Thanks!

src/org/sosy_lab/java_smt/api/FormulaType.java
baierd: We check `checkArgument(mantissaSizeWithHiddenBit > 1, "Mantissa size (including the 'hidden bit') must be greater than 1");` followed by `checkArgument(mantissaSizeWithoutHiddenBit > 0, "Mantissa size (without 'hidden bit') must be greater than 0");`. The only difference is the error message. But to be fair, the cost is constant here and does not really matter. We can keep it like it is for the better error msg.

baierd (approved):

PR #597, opened 2026-02-25
Enforce correct options for `unsatCoreOverAssumptions` and update documentation for `unsatCoreOverAssumptions` with needed `ProverOptions`

daniel-raffler:
Hello,

this PR clarifies that the prover option `ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` is required for `ProverEnvironment.unsatCoreOverAssumptions` by adding the missing links to the documentation. It also adds additional tests for `ProverEnvironment.unsatCoreOverAssumptions` and `ProverEnvironment.getUnsatCore` and improves the error messages. In addition, two solver issues are fixed:
- MathSat is no longer given the option "unsat_core_generation" for `unsatCoreOverAssumptions`. The option isn't needed for `unsatCoreOverAssumptions` and was causing some issues when combined with interpolation over floating-point formulas. However, it still needs to be specified when `getUnsatCore` is (also) used
- Bitwuzla now uses `Bitwuzla.get_unsat_assumptions` for `unsatCoreOverAssumptions`

### DIFF
--- a/src/org/sosy_lab/java_smt/test/UnsatCoreTest.java
+++ b/src/org/sosy_lab/java_smt/test/UnsatCoreTest.java
@@ -46,7 +46,7 @@ public class UnsatCoreTest extends SolverBasedTest0.ParameterizedSolverBasedTest
   }
 
   @Test
-  public void test_getUnsatCore() throws InterruptedException, SolverException {
+  public void simpleGetUnsatCoreTest() throws InterruptedException, SolverException {
     requireUnsatCore();
     try (var prover = context.newProverEnvironment(ProverOptions.GENERATE_UNSAT_CORE)) {
       var a = bmgr.makeVariable("a");
@@ -60,7 +60,7 @@ public class UnsatCoreTest extends SolverBasedTest0.ParameterizedSolverBasedTest
   }
 
   @Test
-  public void test_getUnsatCore_noOption() throws InterruptedException, SolverException {
+  public void getUnsatCoreThrowsWithoutOptionTest() throws InterruptedException, SolverException {
     requireUnsatCore();
     try (var prover = context.newProverEnvironment()) {
       var a = bmgr.makeVariable("a");
@@ -74,7 +74,7 @@ public class UnsatCoreTest extends SolverBasedTest0.ParameterizedSolverBasedTest
   }
 
   @Test
-  public void test_unsatCoreOverAssumptions() throws InterruptedException, SolverException {
+  public void simpleUnsatCoreOverAssumptionsTest() throws InterruptedException, SolverException {
     requireUnsatCoreOverAssumptions();
     // MathSat and old Z3 only allow literals as assumptions:
     assume().that(solver).isNoneOf(Solvers.MATHSAT5, Solvers.Z3_WITH_INTERPOLATION);
@@ -91,7 +91,7 @@ public class UnsatCoreTest extends SolverBasedTest0.ParameterizedSolverBasedTest
   }
 
   @Test
-  public void test_unsatCoreOverAssumptions_onlyLiterals()
+  public void unsatCoreOverAssumptionsOnlyLiteralsTest()
       throws InterruptedException, SolverException {
     requireUnsatCoreOverAssumptions();
     try (var prover =
@@ -107,7 +107,7 @@ public class UnsatCoreTest extends SolverBasedTest0.ParameterizedSolverBasedTest
   }
 
   @Test
-  public void test_unsatCoreOverAssumptions_noOption()
+  public void unsatCoreOverAssumptionsThrowsForMissingOptionTest()
       throws InterruptedException, SolverException {
     requireUnsatCoreOverAssumptions();
     try (var prover = context.newProverEnvironment()) {


src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
```diff
@@ -196,7 +194,7 @@ public List<BooleanFormula> getUnsatCore() {
   public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
       Collection<BooleanFormula> assumptions) throws SolverException, InterruptedException {
     Preconditions.checkNotNull(assumptions);
-    closeAllEvaluators();
```
kfriedberger, line 199: does unsat-core-computation potentially change the stack? then we need to close evaluators.

kfriedberger (dismissed):
see comment. rest looks good to me

src/org/sosy_lab/java_smt/solvers/mathsat5/Mathsat5AbstractProver.java
daniel-raffler: Thanks! I've added the line back in

baierd:
@daniel-raffler could you be so kind and sum up the consequences of these changes?
E.g. is there combinations that worked before, that do not work now. And which ones do now work?

src/org/sosy_lab/java_smt/test/ProverEnvironmentTest.java
```diff
@@ -98,30 +89,31 @@ public void unsatCoreTestForOptimizationProver() throws SolverException, Interru
     }
   }
 
+  private Formula makeNumber(int number) {
+    return imgr == null ? bvmgr.makeBitvector(8, number) : imgr.makeNumber(number);
+  }
+
+  private Formula makeVariable(String name) {
```
baierd: Please reflect the bit size of the BV in the name or documentation, so that it is clear what can be expected.

baierd (changes requested):

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -94,7 +94,6 @@ protected final void checkGenerateUnsatCoresOverAssumptions() {
         generateUnsatCoresOverAssumptions,
         TEMPLATE,
         ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS);
-    Preconditions.checkState(!wasLastSatCheckSatisfiable);
```
baierd, line 97: What are the per-requisites for unsat-core over assumption generation? We currently do not check whether the last SAT check was UNSAT, or whether the stack was modified since then. Are those needed?

Also, there is no tests for this, but we should add some.

baierd:
Looks good to me besides my 2 points. Good job!

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
daniel-raffler: > What are the per-requisites for unsat-core over assumption generation? We currently do not check whether the last SAT check was UNSAT, or whether the stack was modified since then. Are those needed?

We just need the two conditions that are checked here: the solver isn't closed, and the `ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` option was used

The is no need to check the solver state as `unsatCoreOverAssumptions(assumptions)` will always run `isUnsatWithAssumptions` before getting the unsat core

daniel-raffler:
> @daniel-raffler could you be so kind and sum up the consequences of these changes? E.g. is there combinations that worked before, that do not work now. And which ones do now work?

- MathSat now requires `ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` for `ProverEnvironment.unsatCoreOverAssumptions`. This is the same as for all other solvers, we were simply missing the check. However, the "unsat_core_generation" option is now no longer passed to MathSat when `ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` is used. This means that Philipp's interpolation [fallback](https://github.com/sosy-lab/cpachecker/commit/5a60b3886ae034c3f7248662c670e26aaea61a30) will still work on MathSat after adding the `ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS` option. With the option added it should then also work on other solvers
- It is no longer necessary to call `isUnsat` before `unsatCoreOverAssumptions`. (The check was always redundant as `unsatCoreOverAssumptions` will have to make another SAT check anyways. I believe that it might have been a leftover from some earlier refactoring)

You can also go back to 50ac52c273348b3e07e2239a046cef9bf80a687e and run the tests in `UnsatCoreTest` to see the difference

src/org/sosy_lab/java_smt/test/ProverEnvironmentTest.java
daniel-raffler: I've added these so that tests can be written over "numbers", which may be either integers or bitvectors, depending on what the solver supports. This avoids having to write duplicate test to make sure all solvers are covert. The only assumption for "numbers" should be that they support addition and multiplication, and that they can be created from integer constants

I've included similar methods in [598](https://github.com/sosy-lab/java-smt/pull/598) and was thinking of turning this into a more general "NumberManager" just for testing. Can we add the documentation at a later point?

baierd: > I've added these so that tests can be written over "numbers", which may be either integers or bitvectors, depending on what the solver supports. This avoids having to write duplicate test to make sure all solvers are covert. The only assumption for "numbers" should be that they support addition and multiplication, and that they can be created from integer constants

Yes and i really like the idea!
What i want is that either `makeVariable()` or its JavaDoc tell developers what kind of BV they can expect. Using `makeVariable()` only works if the BVs used are equally sized to the other BVs in a test. 

Proposal:
- rename `makeVariable()` to `makeIntOrBvVariable()`
- add JavaDoc explaining when Integer or BV is returned and the bit-size of the BV.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: > ... unsatCoreOverAssumptions(assumptions) will always run `isUnsatWithAssumptions()` before getting the unsat core

OK that solves that. Thanks!

src/org/sosy_lab/java_smt/test/ProverEnvironmentTest.java
daniel-raffler: I've renamed the method and added some documentation

baierd:
Thanks for the summary!

> It is no longer necessary to call `isUnsat` before `unsatCoreOverAssumptions`. (The check was always redundant as `unsatCoreOverAssumptions` will have to make another SAT check anyways. I believe that it might have been a leftover from some earlier refactoring)

We should communicate this in the documentation! I'll add it.

baierd:
>> It is no longer necessary to call isUnsat before unsatCoreOverAssumptions. (The check was always redundant as unsatCoreOverAssumptions will have to make another SAT check anyways. I believe that it might have been a leftover from some earlier refactoring)

> We should communicate this in the documentation! I'll add it.

I now added it to the JavaDoc.

BUT, i wonder whether we named the method wrongly. It is called `unsatCoreOverAssumptions()`, diverging from our schema of methods establishing SAT (UNSAT) e.g. `isUnsat()`, but also from the getters. A name like `checkAndGetUnsatCoreOverAssumptions()` could reflect this better. But this would be a breaking change.
@kfriedberger thoughts?

src/org/sosy_lab/java_smt/test/ProverEnvironmentTest.java
baierd: The new name does not make sense in our naming schema. Either you make a number or a variable. Also, it does not reflect the actual type directly (i.e. i don't know whether i can also get a float by its name). There is also no notion of a 32bit integer in SMT. I fixed it.

baierd:
Thank you for your work! Approved!

PR #598, opened 2026-02-25
Add interpolation to Bitwuzla

daniel-raffler:
Hello everyone

there has been some recent effort to add interpolation support to Bitwuzla, see [here](https://github.com/bitwuzla/bitwuzla/commit/4f9768ad1fda21ba19af9085f3046fffb457fd76). This PR updates our JNI bindings, so that the new interpolation API can be used in JavaSMT. While it works quite well so far, I'm not sure if their implementation is quite finished yet, and it might be better to wait for an official release before merging

### DIFF
--- a/lib/ivy.xml
+++ b/lib/ivy.xml
@@ -196,7 +196,7 @@ SPDX-License-Identifier: Apache-2.0
         <dependency org="org.sosy_lab" name="javasmt-solver-opensmt" rev="2.9.0-gef441e1c" conf="runtime-opensmt-x64->solver-opensmt-x64; runtime-opensmt-arm64->solver-opensmt-arm64; contrib->sources,javadoc"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-optimathsat" rev="1.7.3-sosy1" conf="runtime-optimathsat->solver-optimathsat" />
         <dependency org="org.sosy_lab" name="javasmt-solver-cvc4" rev="1.8-prerelease-2020-06-24-g7825d8f28" conf="runtime-cvc4->solver-cvc4" />
-        <dependency org="org.sosy_lab" name="javasmt-solver-cvc5" rev="2026-01-21-f621457" conf="runtime-cvc5-x64->solver-cvc5-x64; runtime-cvc5-arm64->solver-cvc5-arm64"/>
+        <dependency org="org.sosy_lab" name="javasmt-solver-cvc5" rev="2026-02-26-d22638a" conf="runtime-cvc5-x64->solver-cvc5-x64; runtime-cvc5-arm64->solver-cvc5-arm64"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-boolector" rev="3.2.2-g1a89c229" conf="runtime-boolector->solver-boolector" />
         <dependency org="org.sosy_lab" name="javasmt-solver-bitwuzla" rev="0.8.2-339-g550ed911" conf="runtime-bitwuzla-x64->solver-bitwuzla-x64; runtime-bitwuzla-arm64->solver-bitwuzla-arm64; contrib->sources,javadoc"/>
         <dependency org="org.sosy_lab" name="javasmt-solver-z3-legacy" rev="4.5.0-gd57a2a6dc" conf="runtime-z3-legacy-x64->solver-z3-legacy-x64; runtime-z3-legacy-arm64->solver-z3-legacy-arm64" />
--- a/src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
+++ b/src/org/sosy_lab/java_smt/api/BasicProverEnvironment.java
@@ -26,6 +26,8 @@ public interface BasicProverEnvironment<T> extends AutoCloseable {
   String NO_MODEL_HELP = "Model computation failed. Are the pushed formulae satisfiable?";
   String ASSUMPTION_SOLVING_NOT_SUPPORTED = "Solving with assumptions is not supported.";
   String UNSAT_CORE_NOT_SUPPORTED = "Unsat core extraction is not supported.";
+  String UNSAT_CORE_WITH_ASSUMPTIONS_NOT_SUPPORTED =
+      "Unsat core with assumptions is not supported.";
 
   /**
    * Push a backtracking point and add a formula to the current stack, asserting it. The return
@@ -125,16 +127,24 @@ public interface BasicProverEnvironment<T> extends AutoCloseable {
   }
 
   /**
-   * Get an unsat core. This should be called only immediately after an {@link #isUnsat()} call that
-   * returned <code>false</code>.
+   * Get an unsat core.
+   *
+   * <p>This should be called only immediately after an {@link #isUnsat()} call that returned <code>
+   * false</code>. Requires the {@link ProverOptions#GENERATE_UNSAT_CORE} option to work.
    */
   List<BooleanFormula> getUnsatCore();
 
   /**
    * Returns an UNSAT core (if it exists, otherwise {@code Optional.empty()}), over the chosen
-   * assumptions. Does NOT require the {@link ProverOptions#GENERATE_UNSAT_CORE} option to work.
+   * assumptions. Neither {@link #isUnsat()}, nor {@link #isUnsatWithAssumptions(Collection)} needs
+   * to be called before using this method, as {@link #isUnsatWithAssumptions(Collection)} is
+   * automatically called with the given assumptions.
    *
-   * @param assumptions Selected assumptions
+   * <p>Requires the {@link ProverOptions#GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS} option to work.
+   * {@link ProverOptions#GENERATE_UNSAT_CORE} is not needed, but both options can be safely
+   * combined.
+   *
+   * @param assumptions Selected assumptions.
    * @return Empty optional if the constraints with assumptions are satisfiable, subset of
    *     assumptions which is unsatisfiable with the original constraints otherwise.
    */
--- a/src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
+++ b/src/org/sosy_lab/java_smt/api/FloatingPointNumber.java
@@ -102,7 +102,12 @@ public abstract class FloatingPointNumber {
    */
   public abstract BigInteger getMantissa();
 
-  public abstract int getExponentSize();
+  /** Returns the {@link FloatingPointType} of this {@link FloatingPointNumber}. */
+  public abstract FloatingPointType getFloatingPointType();
+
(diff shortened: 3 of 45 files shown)


src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaModel.java
```diff
@@ -152,12 +151,7 @@ private List<ValueAssignment> buildArrayAssignments(Term expr, Term value) {
   }
 
   private Term buildEqForTwoTerms(Term left, Term right) {
-    assert left.sort().equals(right.sort());
-    Kind kind = Kind.EQUAL;
-    if (left.sort().is_fp() || right.sort().is_fp()) {
-      kind = Kind.FP_EQUAL;
-    }
-    return ((BitwuzlaFormulaCreator) creator).getEnv().mk_term(kind, left, right);
+    return creator.getEnv().mk_term(Kind.EQUAL, left, right);
```
kfriedberger, line 154: This needs to be checked:
- is NAN unequal to NAN? --> we have a test for that in FPFormulaMgrTest.nanAssignedNanIsTrue for formula, but not yet for model assignments.
- is FP with unequal precision considered equal or not? Is that case even supported?

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaInterpolatingProver.java
```diff
@@ -0,0 +1,93 @@
...
+      BitwuzlaFormulaManager pManager,
+      BitwuzlaFormulaCreator pCreator,
+      ShutdownNotifier pShutdownNotifier,
+      Set<ProverOptions> pOptions,
+      Options pSolverOptions) {
+    super(pManager, pCreator, pShutdownNotifier, pOptions, enableInterpolation(pSolverOptions));
+  }
+
+  private static Options enableInterpolation(Options pSolverOptions) {
+    pSolverOptions.set(Option.PRODUCE_INTERPOLANTS, 1);
```
kfriedberger: This call has a side-effect: it changes the given Set. Please avoid such changes and try to use immutable structures.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaInterpolatingProver.java
```diff
@@ -0,0 +1,93 @@
...
+    checkArgument(stack.keySet().containsAll(formulasOfA));
+    return creator.encapsulateBoolean(
+        formulasOfA.isEmpty()
+            ? creator.getEnv().mk_true()
+            : env.get_interpolant(
+                new Vector_Term(FluentIterable.from(formulasOfA).transform(stack::get))));
+  }
+
+  private List<BooleanFormula> init(List<BooleanFormula> formulas) {
+    return formulas.subList(0, formulas.size() - 1);
```
kfriedberger: Why do we need this method? Does Bitwuzla return one formula too much? Maybe inline or rename the method. "init" is a strange name here.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
```diff
@@ -0,0 +1,236 @@
...
+  private final Terminator terminator =
+      new Terminator() {
+        @Override
+        public boolean terminate() {
+          return shutdownNotifier.shouldShutdown(); // shutdownNotifer is defined in the superclass
+        }
+      };
+  private static final UniqueIdGenerator ID_GENERATOR = new UniqueIdGenerator();
+
+  protected final Map<Integer, Term> stack = new HashMap<>();
```
kfriedberger: This data-structure is a potential memory-leak for both keays and values, as it is kept forever until the prover is deleted.
We might want to use a structure like `Deque<PersistentMap<Integer, T>>` as used in other ItpProvers, like PrincessItpProver.
It provides:
- possibility for cleanup on POP
- fast updates on PUSH and ADD_ASSERTION
- instant access to all ids in the top level when queried

kfriedberger (commented):
Overall, this looks quite good.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
kfriedberger: fixed

kfriedberger:
We should document the steps for updating the swig-patchfile. :-) Quite some manual work to get right paths.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaInterpolatingProver.java
kfriedberger: fixed

kfriedberger (dismissed):
can be merged if wanted

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaInterpolatingProver.java
daniel-raffler: Yes, they're using a slightly different definition for `get_interpolants`:

>`std::vector<Term> get_interpolants(const std::vector<std::vector<Term>> &A)`
>
>Get an inductive sequence of interpolants $\langle I_1, \ldots, I_n \rangle$ given the current set of assertions $F$ and a sequence of partitions.
>
>The sequence of partitions is given as a list of set increments $\{A_1, \ldots, A_n\}$ of asserted formulas $F = \{F_1, F_2, \ldots, F_n\}$, which expands into sets of partitions
>
>$$\{(A_1, B_1), (A_2, B_2), \ldots, (A_n, B_n)\}$$
>
>$$\begin{split}A_1 &= F_1 \\
>A_2 &= F_1\,\cup\,F_2 \\
>\ldots & \\
>A_n & = F_1\,\cup\,F_2\,\cup\,\ldots\,\cup\,F_n \\
>\end{split}$$
>
>$$B_i = F \setminus A_i \text{ with } A_i \wedge B_i \models \bot.$$
>
>The resulting sequence of interpolants is inductive, i.e., it holds that $(I_i \wedge F_{i+1}) \rightarrow I_{i+1}$.
>
>Requires that the last check_sat() query returned Result::UNSAT.

This means that N partitions will produce N interpolants. However, in our tests we're expecting only N-1 interpolants. I believe the last interpolant that Bitwuzla returns will always be 'false', so it should be save to just drop it. We probably also could just drop the last partition, assuming that it is just "the rest of the assertions" anyway

kfriedberger: I have added the assertion for `lastInterpolant === FALSE`. Approved.

baierd:
Thank you very much for your work!

Note: Bitwuzlas interpolation is NOT in a released version yet. Merging this would mean we need a new Bitwuzla version based on a non-release version. I would advise to wait for their release.

kfriedberger:
I see no issue with releasing an intermediate version for Bitwuzla. We have intermediate development versions for several solvers. I already uploaded the built version into Ivy, such that this branch here can use it (--> tests are green). There is no blocker for merging.

baierd:
I mean i don't have a problem with intermediate versions in general. I want to avoid accidentally introducing regressions.

I'll run some tests for Bitwuzla.

baierd:
I started BMC (incremental) for the old Bitwuzla and the new, as well as IMC and Predicate Abstraction for this version to test the interpolation. Should be done in ~2 days i guess.

kfriedberger (approved):

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaModel.java
daniel-raffler: > is NAN unequal to NAN? --> we have a test for that in FPFormulaMgrTest.nanAssignedNanIsTrue for formula, but not yet for model assignments.

I've added two more tests in FpFormulaMgrTest that test model assignments. They fail for Bitwuzla before the change, but are working now

> is FP with unequal precision considered equal or not? Is that case even supported?

No, the sorts need to match exactly. Bitwuzla will throw an exception otherwise, and I've added a native test to demonstrate this

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaInterpolatingProver.java
daniel-raffler: Thanks, I completely missed that 😅

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
daniel-raffler: Thanks for the fix!

daniel-raffler:
> We should document the steps for updating the swig-patchfile. :-) Quite some manual work to get right paths.

Yes, it's gotten a bit complicated 😅. I'll add a HOWO, but this can probably be done as a separate PR

baierd:
Some benchmark results for this update:
- [results for BMC](https://github.com/user-attachments/files/25665610/results.2026-03-01_21-37-39.table.html) (incremental) on the SV-COMP26 Reachsafety set in current CPAchecker main seems to be roughly the same. New Bitwuzla solves 4272 tasks correctly while the old solved 4277 correctly. Tthe new version solves a few tasks (11) that the old could not, but also fails to solve 16 that the old could.
- [results for IMC](https://github.com/user-attachments/files/25665636/results.2026-03-01_21-37-04.table.html) on the SV-COMP26 Reachsafety set in current CPAchecker main seems to show that the interpolation is really really good! This compares against MathSAT5 and CVC5, and shows the best results of all 3. 

Quantile-Plot for IMC:
image

PhilippWendler:
Thanks. The improvement over MathSAT would look nice, but I suspect that it is due to just one cause: There are several thousands of tasks in SV-Benchmarks that only need trivial interpolants but where MathSAT fails on interpolation because the tasks have floats. If either MathSAT improves this or we adopt [this workaround](https://gitlab.com/sosy-lab/software/cpachecker/-/issues/1584), CPAchecker with MathSAT can solve at least 3600 tasks more across ReachSafety+SoftwareSystems. Only after this it makes sense to compare solvers.

baierd:
Good to know. From a perspective of comparing solvers and their capabilities it is an OK comparison. It is just a snapshot of "what can a solver verify for this algorithm (and tool) using X".

(Better would be more data from different sources. Usually we just compare what we use the solvers for; verification. But we know from the SMT competition that this is not the most tested/evaluated use-case.)

PR #600, opened 2026-02-27
Fix: 599 can't get model after successfull sat check with assumptions

### LINKED ISSUE #599: Can't get model after successfull SAT check with assumptions
Hello,

several of the solvers (MathSat, Z3, ...) don't update `wasLastSatCheckSatisfiable` in `ProverEnvironment.isUnsatWithAssumptions`. Because of that any subsequent call to `ProverEnvironement.getModel` will fail, even if the SAT check was successfull:
```
Model computation failed. Are the pushed formulae satisfiable?

java.lang.IllegalStateException: Model computation failed. Are the pushed formulae satisfiable?
at com.google.common.base.Preconditions.checkState(Preconditions.java:513)
at org.sosy_lab.java_smt.basicimpl.AbstractProver.checkGenerateModels(AbstractProver.java:76)
at org.sosy_lab.java_smt.solvers.mathsat5.Mathsat5AbstractProver.getModel(Mathsat5AbstractProver.java:147)
at org.sosy_lab.cpachecker.util.predicates.smt.BasicProverEnvironmentView.getModel(BasicProverEnvironmentView.java:66)
at org.sosy_lab.cpachecker.util.predicates.weakening.CEXWeakeningManager.performWeakening(CEXWeakeningManager.java:109)
at org.sosy_lab.cpachecker.util.predicates.weakening.InductiveWeakeningManager.findSelectorsToAbstract(InductiveWeakeningManager.java:229)
at org.sosy_lab.cpachecker.util.predicates.weakening.InductiveWeakeningManager.findInductiveWeakeningForRCNF(InductiveWeakeningManager.java:171)
at org.sosy_lab.cpachecker.util.predicates.weakening.InductiveWeakeningManagerTest.testSlicingVerySimple(InductiveWeakeningManagerTest.java:78)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) 
```

This is causing several tests in CPAchecker to fail with newer JavaSMT versions, and would likely also cause issues in the analysis. Solvers that use `BasicProverWithAssumptionsWrapper` for assumption solving are not affected by the issue

We should fix the issue and add some additional tests for `ProverEnvironment.isUnsatWithAssumptions` to catch issues like this more easily in the future

baierd:
This PR adds updates to SAT status and stack change to all solver provers that implement `isUnsatWithAssumptions()` on their own (without the wrapper that already updates the status), as there is no easy way to commonly do it for this method currently. Tests for expected behavior are also added.

Closes #599.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
@@ -213,7 +213,6 @@ public abstract class AbstractProver<T> implements BasicProverEnvironment<T> {
    * @throws UnsupportedOperationException if optimization is not supported by this prover
    */
   protected OptStatus checkImpl() throws InterruptedException, SolverException {
-    // TODO: Should this also change wasLastSatCheckSatisfiable etc.?
     if (this instanceof OptimizationProverEnvironment) {
       throw new UnsupportedOperationException("checkImpl() must be implemented in a subclass.");
     }


daniel-raffler:
Could we maybe move the checks into `AbstractProver`, similiar to how it's already done for `isUnsat`?
```java
  @Override
  public boolean isUnsatWithAssumptions(Collection<BooleanFormula> assumptions)
      throws SolverException, InterruptedException {
    checkState(!closed);
    changedSinceLastSatQuery = false;
    wasLastSatCheckSatisfiable = false;
    final boolean isUnsat = isUnsatWithAssumptionsImpl(assumptions);
    if (!isUnsat) {
      wasLastSatCheckSatisfiable = true;
    }
    return isUnsat;
  }

  protected abstract boolean isUnsatWithAssumptionsImpl(Collection<BooleanFormula> assumptions)
      throws SolverException, InterruptedException;
```
Solvers can then simply implement `isUnsatWithAssumptionsImpl` and wouldn't have to update any flags themselves

baierd:
While i would also prefer such a solution, the answer is; not without re-working many components of our prover architecture. I tried....


We should rework this after the big release and merge the assumption solving components into the `AbstractProver` in all cases if possible.

baierd:
Note: AppVeyor test failed the `SolverConcurrencyTest` and i can't restart it.

PR #601, opened 2026-02-27
Unify optimization solving getModel() behavior

baierd:
Requesting a model should fail if it is not done immediately after SAT has been returned. This was not tested for optimization solving. This PR adds tests and unifies the thrown exception (our MathSAT5 impl threw an incorrect exception).


baierd:
@kfriedberger I don't know much about optimization solving; do we need the model checks/tests also for `upper()` `lower()`?

kfriedberger:
The methods `upper/lower` are similar to `getModel` and return values that are based on the last sat-check/optimization query. I would assume that these values become invalid if constraints are changed in any way.

baierd:
Thank you for the quick answer!
I added the model checks also to those 2 methods, and added tests for them. As a consequence we now require the model generation option for them. I guess that is fair.

PR #607, opened 2026-03-02
Cvc5: Fix parsing bug

daniel-raffler:
Hello,

our CVC5 parsing code has an issue that allows duplicate versions of the same variable to be created. The problem is that quotes are not removed from symbols when looking them up in the variable cache. Because of a new variable may be created when looking up a quoted symbol, even though the variable already exists without the quotes

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
@@ -581,13 +581,4 @@ public abstract class FormulaCreator<TFormulaInfo, TType, TEnv, TFuncDecl> {
       @SuppressWarnings("unused") TFormulaInfo pAdditionalF, TFormulaInfo pF) {
     return convertValue(pF);
   }
-
-  /** Variable names (symbols) can be wrapped with "|". This function removes those chars. */
-  public String dequote(String s) {
-    int l = s.length();
-    if (s.charAt(0) == '|' && s.charAt(l - 1) == '|') {
-      return s.substring(1, l - 1);
-    }
-    return s;
-  }
 }
--- a/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4Model.java
+++ b/src/org/sosy_lab/java_smt/solvers/cvc4/CVC4Model.java
@@ -120,7 +120,7 @@ public class CVC4Model extends AbstractModel<Expr, Type, ExprManager> {
         creator.encapsulateWithTypeOf(pKeyExpr),
         creator.encapsulateWithTypeOf(valueExpr),
         creator.encapsulateBoolean(creator.getEnv().mkExpr(Kind.EQUAL, pKeyExpr, valueExpr)),
-        ((CVC4FormulaCreator) creator).getName(pKeyExpr),
+        CVC4FormulaCreator.getName(pKeyExpr),
         creator.convertValue(pKeyExpr, valueExpr),
         argumentInterpretationBuilder.build());
   }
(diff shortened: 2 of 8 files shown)


src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
```diff
@@ -583,7 +583,7 @@ public Object convertValue(
   }
 
   /** Variable names (symbols) can be wrapped with "|". This function removes those chars. */
-  protected static String dequote(String s) {
+  public String dequote(String s) {
```
baierd: We usually don't want anything from the creator publicly accessible, as the creator itself is hidden. Maybe the better move would be to make this available in the `FormulaManager` as a common method in a protected way for now? (Possibly a parsing/dumping util class would be helpful in the future)

src/org/sosy_lab/java_smt/solvers/cvc4/CVC4Model.java
```diff
@@ -120,7 +120,7 @@ private ValueAssignment getAssignmentForUfInstantiation(Expr pKeyExpr) {
         creator.encapsulateWithTypeOf(pKeyExpr),
         creator.encapsulateWithTypeOf(valueExpr),
         creator.encapsulateBoolean(creator.getEnv().mkExpr(Kind.EQUAL, pKeyExpr, valueExpr)),
-        CVC4FormulaCreator.getName(pKeyExpr),
+        ((CVC4FormulaCreator) creator).getName(pKeyExpr),
```
baierd: That is exactly the reason why we don't want the methods in the creator to be public ;D (This can be closed by you after you've seen it)

src/org/sosy_lab/java_smt/test/ParserTest.java
```diff
@@ -265,4 +266,23 @@ public void testParseAll_invalid_reservedKeyword() throws SolverException, Inter
       assertThrows(IllegalArgumentException.class, () -> mgr.parseAll(smt));
     }
   }
+
+  @Test
+  public void testParseall_quotedSymbol() {
```
baierd: Try to stick to the Java naming scheme ;D
`testParseall_quotedSymbol` -> `parseAllQuotedSymbolTest`

src/org/sosy_lab/java_smt/test/ParserTest.java
```diff
@@ -265,4 +266,23 @@ public void testParseAll_invalid_reservedKeyword() throws SolverException, Inter
...
+  }
+
+  @Test
+  public void testParseall_quotedSymbol_reparse() {
+    // Parse a variable that was already defined in the context
+    var f = mgr.makeVariable(BooleanType, "my variable");
+    var str = mgr.dumpFormula(f).toString();
+    var g = mgr.parse(str);
+
+    assertThat(g).isEqualTo(f);
```
baierd, line 286: You should also assert that it was really quoted in `str`.

src/org/sosy_lab/java_smt/basicimpl/FormulaCreator.java
daniel-raffler: I've moved `dequote` to `Tokenizer` as it is the closest thing we have to a parsing/dumping util class

src/org/sosy_lab/java_smt/test/ParserTest.java
daniel-raffler: All the names in the class should be updated now

src/org/sosy_lab/java_smt/test/ParserTest.java
daniel-raffler: The variable name contains a space character. So it needs to be quoted or parsing would fail. However, it's difficult to assert that as different solvers will generate different output in `mgr.dumpFormula`

I've now changed the test to  use a fixed String as input

kfriedberger (approved):
lgtm. thanks for the PR.

PR #610, opened 2026-03-06
update several dependencies

kfriedberger:


kfriedberger:
Update of libraries will be done in another way.

PR #611, opened 2026-03-07
Switch from Java 11 to Java 17

kfriedberger:
This pull request upgrades JavaSMT from Java 11 to Java 17.
Following the plan outlined in the [JavaSMT 6.0.0 release notes](https://github.com/sosy-lab/java-smt/releases/tag/6.0.0), this move enables us to adopt newer Java features and sets the stage for future upgrades to Java 21.

### Key Changes:

  * **Build Configuration & CI:**
       * Updated local configuration files (`.classpath`, `.settings`, `build/build-compile.xml`) to set the source and target release to Java 17.
       * Updated `.github/workflows/ci.yml`, `.travis.yml`, and `.gitlab-ci.yml` to use JDK 17 as the primary build environment.
       * Updated Dockerfiles to use JDK 17 (not yet deployed).
   * **Java 17 Records:**
       * Replaced several `@AutoValue` classes with native Java records to eliminate generated code dependencies.
       * Converted internal data structures like `Coordinates` (N-Queens example) and `Z3TreeInterpolant` to records.
   * **Pattern Matching & Switch Expressions:**
       * Implemented Pattern Matching for `instanceof` to eliminate redundant type casting.
       * Modernized switch statements using the new Switch Expressions (arrow syntax) for more concise and safe control flow, i.e., removing default-cases where not required.
   * **Text Blocks:**
       * Refactored complex, multi-line SMT-LIB2 queries and model strings in test suites (`FormulaClassifierTest`, `CVC5NativeAPITest`, etc.) using Text Blocks (`"""`), improving the readability of solver interactions.
   * **API Improvements:**
       * Migrated from `stream.collect(Collectors.toList())` to the more concise `stream.toList()`.
       * Began transitioning `String.format(...)` calls to the more idiomatic `String.formatted(...)` instance method.
   * **Stability & Fixes:**
       * Resolved compilation errors and Error Prone warnings triggered by stricter checks in the Java 17 environment, including `@Immutable` container analysis and format string validation.

### Verification:
   * Successfully built the project using ant on JDK 17, 21, and 24.
   * Verified that all compilation warnings and Error Prone errors have been resolved.
   * No public API changed.

### DIFF
--- a/src/org/sosy_lab/java_smt/delegate/trace/TraceBooleanFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/delegate/trace/TraceBooleanFormulaManager.java
@@ -150,13 +150,12 @@ class TraceBooleanFormulaManager implements BooleanFormulaManager {
   public <R> R visit(BooleanFormula pFormula, BooleanFormulaVisitor<R> visitor) {
     return logger.logDefDiscard(
         "mgr.getBooleanFormulaManager()",
-        """
-        visit(%s, new DefaultBooleanFormulaVisitor<>() {
-          protected Formula visitDefault(Formula f) {
-          return %s;
-        }})
-        """
-            .formatted(logger.toVariable(pFormula), logger.toVariable(pFormula)),
+        String.format(
+            "visit(%s, new DefaultBooleanFormulaVisitor<>() {"
+                + "  protected Formula visitDefault(Formula f) {"
+                + "  return %s;"
+                + "}})",
+            logger.toVariable(pFormula), logger.toVariable(pFormula)),
         () -> delegate.visit(pFormula, visitor));
   }
 
@@ -165,13 +164,12 @@ class TraceBooleanFormulaManager implements BooleanFormulaManager {
       BooleanFormula f, BooleanFormulaVisitor<TraversalProcess> rFormulaVisitor) {
     logger.logStmtDiscard(
         "mgr.getBooleanFormulaManager()",
-        """
-        visitRecursively(%s, new DefaultBooleanFormulaVisitor<>() {
-          protected TraversalProcess visitDefault(Formula f) {
-          return TraversalProcess.CONTINUE;
-        }}})
-        """
-            .formatted(logger.toVariable(f)),
+        String.format(
+            "visitRecursively(%s, new DefaultBooleanFormulaVisitor<>() {"
+                + "  protected TraversalProcess visitDefault(Formula f) {"
+                + "  return TraversalProcess.CONTINUE;"
+                + "}})",
+            logger.toVariable(f)),
         () -> delegate.visitRecursively(f, rFormulaVisitor));
   }
 
--- a/src/org/sosy_lab/java_smt/delegate/trace/TraceFormulaManager.java
+++ b/src/org/sosy_lab/java_smt/delegate/trace/TraceFormulaManager.java
@@ -1133,13 +1133,12 @@ class TraceFormulaManager implements FormulaManager {
   public <R> R visit(Formula f, FormulaVisitor<R> rFormulaVisitor) {
     return logger.logDefDiscard(
         "mgr",
-        """
-        visit(%s, new DefaultFormulaVisitor<>() {
-          protected Formula visitDefault(Formula f) {
-          return f;
-        }})
-        """
-            .formatted(logger.toVariable(f)),
+        String.format(
+            "visit(%s, new DefaultFormulaVisitor<>() {"
+                + "  protected Formula visitDefault(Formula f) {"
+                + "  return f;"
+                + "}})",
+            logger.toVariable(f)),
         () -> delegate.visit(f, rFormulaVisitor));
   }
 
@@ -1147,13 +1146,12 @@ class TraceFormulaManager implements FormulaManager {
   public void visitRecursively(Formula f, FormulaVisitor<TraversalProcess> rFormulaVisitor) {
     logger.logStmtDiscard(
         "mgr",
-        """
-        visitRecursively(%s, new DefaultFormulaVisitor<>() {
-          protected TraversalProcess visitDefault(Formula f) {
-          return TraversalProcess.CONTINUE;
-        }})
-        """
-            .formatted(logger.toVariable(f)),
+        String.format(
+            "visitRecursively(%s, new DefaultFormulaVisitor<>() {"
+                + "  protected TraversalProcess visitDefault(Formula f) {"
+                + "  return TraversalProcess.CONTINUE;"
+                + "}})",
+            logger.toVariable(f)),
         () -> delegate.visitRecursively(f, rFormulaVisitor));
   }
 


PhilippWendler:
What kind of review do you want and for which parts? I am glad to help, but obviously I am not going to review 137 changed files with thousands of changed lines that seems to be a mix of tool-generated changes and manual changes.

In order to make review feasible for such large changes, I recommend to clearly separate automatic and manual changes, make only one particular kind of change in one commit (important if you want to revert it because for example a bug in the transformation was detected), and clearly document which changes were done automatically and by which tool (to make them reproducible).

I also recommend to forbid classic switch case labels completely and enforce arrow rules, as we [did in CPAchecker](https://gitlab.com/sosy-lab/software/cpachecker/-/work_items/1339).

`sealed` has been useful for finding conditions that could never actually happen and I would recommend to adopt it where possible. I would say for example that everything in the `solvers` package should be either `final` or `sealed`.

Recommendations and thoughts on Java 17 features from me you can also find [here](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/100).

kfriedberger:
Thanks for the reponse, @PhilippWendler.

I've already split the commits according to smaller logical and compilable steps, for example with changes for `instanceof` and `switch-case` (both combined in 099d476eada5d7a8d33c2a34acd259e9fc2450c1 and 27b7c11c1f9cbb719d7c097ac628febe79164f7c), `string formatting` and `multiline strings` (053921bd83d013b5e50c3b2afaf95417dc0c92bf, 39effb8233548aa20d1d373e2df8ef14ca026540). The steps were not as small and detailed documented as those for CPAchecker. I do not expect you to check every single line of code, but rather get a broad overview and look at the most important files, such as compiler, styleguide, and pipeline confguration.

For the classic switch-case style, I see the point of forbidding it. Your feedback on exactly that kind of missed guidelines for developers is very good, and JavaSMT can align better with the standards of SoSy-Lab in general.

One point I'm still unsure about is the effect of using sealed classes on public classes. I'm concerned that users, such as CPAchecker, might be affected by these changes. I want to ensure we don't introduce unexpected breaking changes for them. We can also apply such changes in the future.

.travis.yml
```diff
@@ -9,7 +9,7 @@
 dist: focal
 language: java
 jdk:
-  - openjdk11 # Java 11 is sufficient, larger tests are executed in Gitlab-CI
+  - openjdk17 # Java 17 is sufficient, larger tests are executed in Gitlab-CI
```
PhilippWendler: Is JavaSMT still using Travis? I don't see such CI results.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaDeclaration.java
```diff
@@ -8,32 +8,26 @@
...
 
 // Declarations sometimes need the info of a Term, but mostly those of Kinds.
 // We can not discern between the two however, hence this wrapper
-@AutoValue
-public abstract class BitwuzlaDeclaration {
-  public abstract @Nullable Term getTerm();
-
-  public abstract @Nullable Kind getKind();
-
+public record BitwuzlaDeclaration(@Nullable Term getTerm, @Nullable Kind getKind) {
```
PhilippWendler: Most of the new records (like this one) miss nullability checks.

lib/ivy.xml
```diff
@@ -111,30 +111,30 @@ SPDX-License-Identifier: Apache-2.0
 
         <!-- Google Core Libraries for Java
              Contains a lot of helpful data structures. -->
-        <dependency org="com.google.guava" name="guava" rev="33.4.8-jre" conf="core->default; contrib->sources"/>
+        <dependency org="com.google.guava" name="guava" rev="33.5.0-jre" conf="core->default; contrib->sources"/>
```
PhilippWendler, line 114: Most changes here in this class do not seem to be related to the migration to Java 17.

lib/ivy.xml
```diff
@@ -111,30 +111,30 @@ SPDX-License-Identifier: Apache-2.0
...
 
         <!-- Dependency on Ivy itself so that we can ugprade it easily.
              Change version number in build/build-ivy.xml for upgrading. -->
         <dependency org="org.apache.ivy" name="ivy" rev="${ivy.target_version}" conf="build->default"/>
 
         <!-- Google Auto-Value
              Library for auto-generating value types. -->
-        <dependency org="com.google.auto.value" name="auto-value" rev="1.11.0" conf="build->default"/>
-        <dependency org="com.google.auto.value" name="auto-value-annotations" rev="1.11.0" conf="build->default; contrib->sources"/>
+        <dependency org="com.google.auto.value" name="auto-value" rev="1.11.1" conf="build->default"/>
```
PhilippWendler: I wonder why you kept AutoValue and did not remove it completely?

PhilippWendler:
> One point I'm still unsure about is the effect of using sealed classes on public classes. I'm concerned that users, such as CPAchecker, might be affected by these changes. I want to ensure we don't introduce unexpected breaking changes for them.

Of course for the public API one has to be careful. I would not make `Formula` and related classes sealed. For the solver-specific classes, I wonder why they are even public and which guarantees JavaSMT makes about being able to access and extend them directly? This also seems to be inconsistent.

Anyway, many of these classes already are not public or have constructors that are private or package-private (which prevents inheritance by others). It should be safe to make these `sealed`. Users of JavaSMT could still get compiler errors from such a change, but only if they did casts that can never succeed (which I would see as a good thing) or in rare cases like [here](https://github.com/sosy-lab/java-common-lib/commit/78f0073c40879fc3c3e7895c87a53cd3fd586808#diff-aabe0289d781f77078d8afa0ec91eae5f92103fac98fd013e71a0088cbc34ac8) (but even this only for public classes).

.travis.yml
kfriedberger: I could not find anything either. I will remove this file. Thanks for noticing.

lib/ivy.xml
kfriedberger: Done with bf3b658, where the last dependency on AutoValue was removed.

lib/ivy.xml
kfriedberger: Multiple dependencies no longer support Java 11  in newer versions. The update depends on Java 17. We could have done parts of this dependency update later.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaDeclaration.java
kfriedberger: I added more null-checks in 7b8596050b.

ThomasHaas:
> Of course for the public API one has to be careful. I would not make `Formula` and related classes sealed. 

I agree with this from a user's side: being able to inherit from the API types is useful, for example, to implement a cheap [TupleFormula](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/smt/TupleFormula.java) on top of JavaSMT (having this directly in the JavaSMT API would be even better though :P).
That being said, the `FormulaType` class has a private constructor (a bummer IMO) and so is effectively sealed already. AFAIK, its the only class in the API with that property and I'm not sure why? If it is intended, you can make the class sealed. If not, you should remove the private constructor. In either case, the class could also be changed to an interface.

kfriedberger:
@ThomasHaas 
Thanks for your feedback. We try to keep the (main) API stable, but will remove solver-specific public parts with the next release (see #614).
Your comment contains two feature requests, for which it would be good to create [issues](https://github.com/sosy-lab/java-smt/issues):
- convert abstract class `FormulaType` to interface, This would allow users to inherit and override it with self-defined formula types. This looks reasonable on first reading, but formulas of arbitrary type might not be usable in JavaSMT. THe change might be quite simple, see #615.
- add a new `TupleFormula` representing a ordered list of formulas, including recursive cases as well. This sounds like a type-unsafe pattern and will not give us compile-time safety for formula usage. The idea is interesting. Z3 might even have support for such cases (see [Z3 docu](https://z3prover.github.io/api/html/class_microsoft_1_1_z3_1_1_tuple_sort.html).

kfriedberger:
@PhilippWendler 
I marked some classes as `sealed`. There are some abstract classes remaining, which are no longer visible from outside and do not get any benefit from sealing. Do you have any more feedback on this PR?

.idea/JavaSMT.iml
```diff
@@ -367,6 +367,15 @@ SPDX-License-Identifier: Apache-2.0
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="module-library">
+      <library>
+        <CLASSES>
+          <root url="jar://$MODULE_DIR$/lib/java/core/jspecify.jar!/" />
```
baierd, line 382: Do we use jspecify? I thought we use error-prone annotations per default.

baierd:
Thank you very much for this "small" change :D 
I did not manage to look at everything (roughly half), but what I've seen looks good! I'll try to look at the rest once i can.

docker/ubuntu2204.Dockerfile
daniel-raffler: Could you update the JDK version here too? Same goes for the other docker files

.idea/JavaSMT.iml
kfriedberger: jspecify is a transitive dependency of google-guava, so we keep it

docker/ubuntu2204.Dockerfile
kfriedberger: done. thanks.

PhilippWendler:
> I marked some classes as `sealed`. There are some abstract classes remaining, which are no longer visible from outside and do not get any benefit from sealing.

Note that the benefit of sealing abstract classes is not really to prevent external inheritance (this can be done with (package-)private constructors), but to to tell the compiler about this and let it use this knowledge when reasoning about code like `instanceof` checks and pattern-matching `switch` expressions. This can find accidentally wrong conditions. So while I am not saying that it must be added immediately when upgrading to Java 17, I would personally just use `sealed` everywhere where it is possible.

> Do you have any more feedback on this PR?

If you want I can have another look in April.

PR #614, opened 2026-03-13
reduce visibility of several library-internal classes.

kfriedberger:
This PR reduces the visibility of several internal classes within JavaSMT to better define our public API and prevent accidental dependencies on internal logic.

**Note:**
This is a breaking change. If a user application was previously accessing these internal classes directly, it may need to update its implementation to use the supported public interfaces. This change ensures a more stable and maintainable API surface moving forward.


daniel-raffler (approved):
Seems like a good idea! I think you missed a few classes:
- `BitwuzlaFloatingPointManager`
- `CVC4BitvectorFormulaManager`
- `CVC4FloatingPointFormulaManager`
- `CVC5FloatingPointFormulaManager`

Other than that, lgtm

kfriedberger:
@daniel-raffler 
Thanks for the quick review. I missed a view other locations, too.
I searched again and reduced visibility for several classes even more.

PR #615, opened 2026-03-13
API change: convert the abstract class "FormulaType" to interface.

kfriedberger:
The idea for this change was brought up in a [comment about potential API changes](https://github.com/sosy-lab/java-smt/pull/611#issuecomment-4058097915).

This change makes JavaSMT more open for user applications, as it allows them to define their own formula types and use them in JavaSMT. This also blocks us from marking this class as "sealed" at some time in the future, because users can inherit from it.

We can discuss benefit and/or negative effects from this change.

### DIFF
--- a/src/org/sosy_lab/java_smt/api/FormulaType.java
+++ b/src/org/sosy_lab/java_smt/api/FormulaType.java
@@ -412,29 +412,25 @@ public interface FormulaType<T extends Formula> {
     }
   }
 
-  @SuppressWarnings("ClassInitializationDeadlock")
   FormulaType<FloatingPointRoundingModeFormula> FloatingPointRoundingModeType =
-      new FloatingPointRoundingModeType();
+      new FormulaType<>() {
 
-  final class FloatingPointRoundingModeType
-      implements FormulaType<FloatingPointRoundingModeFormula> {
+        @Override
+        public boolean isFloatingPointRoundingModeType() {
+          return true;
+        }
 
-    @Override
-    public boolean isFloatingPointRoundingModeType() {
-      return true;
-    }
+        @Override
+        public String toString() {
+          return "FloatingPointRoundingMode";
+        }
 
-    @Override
-    public String toString() {
-      return "FloatingPointRoundingMode";
-    }
-
-    @Override
-    public String toSMTLIBString() {
-      throw new UnsupportedOperationException(
-          "rounding mode is not expected in symbol declarations");
-    }
-  }
+        @Override
+        public String toSMTLIBString() {
+          throw new UnsupportedOperationException(
+              "rounding mode is not expected in symbol declarations");
+        }
+      };
 
   @SuppressWarnings("MethodTypeParameterName")
   static <TD extends Formula, TR extends Formula> ArrayFormulaType<TD, TR> getArrayType(


ThomasHaas:
Just to extend the discussion from #611 a bit.

 Inheriting from the JavaSMT types is generally dangerous in that the user-specified types will not interact well (if at all) with the existing methods in JavaSMT. Essentially, a custom type cannot be used as a parameter to any method of JavaSMT.
However, the user can put simple wrappers around JavaSMTs interface to handle the custom types in several cases. To give a concrete example, I refer to Dartagnan's custom [TupleFormula](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/smt/TupleFormula.java)/[TupleFormulaManager](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/smt/TupleFormulaManager.java) and the necessary wrappers [FormulaManagerExt](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/smt/FormulaManagerExt.java) for extending `ITE` to custom tuples, and [ModelExt](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/smt/ModelExt.java) for extending model evaluation to tuples.

In a similar way, record types could be implemented on the user-side, or more specialized types like `PointerFormula` (consisting of, e.g., `EnumerationFormula object` and an `IntegerFormula address`).

I have not tested how far the user could go with custom wrappers. IIRC, the fact that `FormulaType` was effectively sealed was a problem though.

baierd:
CPAchecker also uses wrappers around many classes, but not as deep as Dartagnan. However, maybe we can actually improve our interfaces in the future to ease usage as described above.

PhilippWendler:
Uhm, have you really thought this through? I don't really see any discussion about the benefits and disadvantages of this change.

I think this is quite a large API change, and needs to be carefully considered!

AFAIK it is a quite hard breaking change for binary compatibility. And I do not see any reason for this, because if the goal would simply be to make custom subclasses possible, then it would be completely easy to do so in a backwards compatible manner by simply making the constructor of the abstract class `protected`.

But of course one also needs to discuss the API implications. Is it really a good idea to do this? In general, in software design the best approach is to make an API as restricted as possible and only open it up where necessary. For example, this makes it harder to JavaSMT in the future to evolve these types. One needs to consider how JavaSMT would react if someone implements their own subclass and uses it instead of JavaSMT's classes, one needs to document what implementors of subclasses need to consider, what the stability guarantees are, etc.

So, frankly, I would recommend to revert this and then (if you want to) start an actual discussion about this, with arguments for both sides and considerations what is important for JavaSMT.

Furthermore, I am a little bit baffled, that you ask for my review but then just merge it without even waiting. Do you want my reviews or not?

kfriedberger:
Hi @PhilippWendler , thanks for the feedback.

I understand the concern regarding the loss of strict control that a closed class provides. However, opening this hierarchy, whether as an interface or a protected class, improves the ability to mock types in unit tests, implement decorators, or create specialized subclasses without compromising JavaSMT's internal logic.

The choice between an interface and a protected class has a minor impact on JavaSMT internally, but a major impact on the consumer:
- With an interface, the user can decide on his own hierarchy and use any class of his project to inherit from it.
- With an protected class, the user is limited to only ever implement our hierachy, because Java does not allow multi-inheritance inheritance.

I had chosen "interface" also to be symmetric to `Formula` itself, which is an open, non-sealed class that can be overridden by users. Maybe I did not yet fully think of its affects on creating such classes in user applications and giving them to JavaSMT. Feel free to give a better solution.

Regarding binary compatibility, we use default methods in the interface. A "protected class" might even be more compatible. I thought default methods from an interface are already compatible enough. I have not tested it.

Regarding the timing: since the PR was open for 3 days without blockers (including weekend, not only plain work days) and the change is small enough to be easily reverted or refined if we hit edge cases, I decided to merge it to keep the workflow moving and showing progress on user-requested topics.

ThomasHaas:
In terms of the current public API, as far as I can see, all classes/interfaces are public for interitance (not that you could have prevented this on interfaces without the new sealed keyword) with the exception of `FormulaType`. 
So strictly speaking, `FormulaType` is the only class that JavaSMT has full control over in its current state. Allowing it to be extended makes the API more consistent and might be reasonable as an intermediate step even if you eventually end up deciding to seal most/all classes in the API.
Of course, this is only reasonable if you do not make any guarantees that it will stay public and don't give any guarantees about the interaction of JavaSMT and custom user types, which are precisely the guarantees current users get anyways, i.e., none at all.

However, that doesn't make custom subclasses inherently useless. The reason is that a tool that uses JavaSMT not only imports its functionality but also its interfaces. 
If everything is sealed, then a user that wants to add custom extensions on top has to wrap both the functionality and the interfaces behind custom classes/interfaces, i.e., come up with his own solver-agnostic SMT interface.
If at least the interface is reusable (and JavaSMT has a very nice interface :)), then one can "just" wrap the functionality which is IMO quite a bit easier to do.

That being said, I think providing an extensible API that works well with JavaSMT's existing functionality (i.e., gives guarantees) is a very difficult task and probably not worth it overall. This certainly requires a lot of thinking and discussion.

PhilippWendler:
> Regarding binary compatibility, we use default methods in the interface. A "protected class" might even be more compatible. I thought default methods from an interface are already compatible enough. I have not tested it.

AFAIK the change from a class to an interface changes the ABI and requires recompilation of all users of the type, no matter whether there are default methods.

> Regarding the timing: since the PR was open for 3 days without blockers (including weekend, not only plain work days) and the change is small enough to be easily reverted or refined if we hit edge cases, I decided to merge it to keep the workflow moving and showing progress on user-requested topics.

Uhm, 3 days is really not a lot of time, *especially* on the weekend where not everyone works. I am not a maintainer of JavaSMT and not requesting to be asked for every change, but if you do ask me and want my feedback, you cannot expect that I can always answer within less than one business day each time. Apart from weekends, people are also not available sometimes for other reasons, and for example I will actually not be available for the next two weeks.

For discussion the actual change, maybe an issue is a better and more discoverable place than a closed and reverted PR. If I were a maintainer, I would for example like to see a *concrete* case where the existing situation actually is problematic for a user of JavaSMT and opening up the hierarchy to a inheritable class or an interface would make things easier.

PR #618, opened 2026-03-18
Enable missing test for CVC5

daniel-raffler:
Hello,

this PR enables `TranslateFormulaTest` for CVC5 as parsing is now supported. It also removes the exception for CVC5 in `SolverThreadLocalityTest` as the concurrency issues have been resolved


daniel-raffler:
Thanks for the quick review!

PR #619, opened 2026-03-20
Bitwuzla: Simplify updating the SWIG patch

daniel-raffler:
Hello,

this PR is in response to a comment when we last updated the Bitwuzla bindings:
> We should document the steps for updating the swig-patchfile. :-) Quite some manual work to get right paths.
[link](https://github.com/sosy-lab/java-smt/pull/598#issuecomment-3962927859)

To simplify updating the patch the build process for Bitwuzla has now been split into two separate steps:
* We first generate the new SWIG bindings with `ant bitwuzla-generate-jni`
* We then have a chance to fix any issues with the generated code
* Finally, we run `ant publish-bitwuzla` to automatically generate the updated patch file before then compiling the solver binaries

I've documented the process in `Developers-How-to-Release-into-Ivy.md`, including some hints on how to fix common issues with the generated code

Note that I also switched to using the preinstalled `jdks` from the container: We were already using the preinstalled versions of gmp/mpfr, and this removes two more arguments from the command line


daniel-raffler:
Thanks for the merge!

Issue #613, opened 2026-03-11
[Feature request] Setting Solver Logic via JavaSMT


ThomasHaas:
Is it possible to add a way to set the solver logic from within JavaSMT similar to Smtlib's `(set-logic <LOGIC>)` command?

This is particularly interesting in the context of incremental solving, because no matter how smart the SMT solver is, with the default `ALL` logic, he has to assume that subsequent queries introduce new logics.
This can easily be demonstrated with Z3 and `verbose=5` on smtlib2 QF_BV queries (outside JavaSMT):
- on non-incremental queries, Z3 bit-blasts and uses its SAT solver even with `(set-logic ALL)`
- on incremental queries, Z3 runs its general-purpose SMT core with `(set-logic ALL)`, even if the whole file only uses bit vectors. When changing the logic to `(set-logic QF_BV)`, Z3 bit-blasts again.

In the context for program verification, we often do incremental queries (via JavaSMT) that only make use of `QF_BV`. We have observed that if we dump the generated queries to smtlib2 and simply change the logic from `ALL` to `QF_BV`, the solver performance can increase by 50-100%. Granted, there are also cases where changing the logic makes the solver slower, but it is hard to evaluate the exact impact if we have to dump and modify the encoding and then invoke the solver externally.

baierd:
Hi Thomas,

we discussed this feature a while back, but were not sure how helpful it really is. OpenSMT2 is currently the only solver that can choose the logic, as it needs it and does not support `all`.

I'll try to take a look at how difficult a general solution for this would be and add it once i can.

ThomasHaas:
Z3 also has `smt.logic` as option to set the logic. But I agree that setting the logic has only limited application, even if possible. For example, in the context of concurrent program verification, it is quite natural to capture data flow via `QF_BV` and ordering constraints via `QF_IDL`, but the only smtlib theory that combines both is `ALL`.
To exploit dedicated logic settings, the user either has to bit-blast to reduce the formula to `QF_IDL` or replace the ordering constraints by a SAT or BV-based encoding (e.g. vertex elimination).

So what would be even more useful in practice is the ability to pass more options to the SMT solver as was requested [a while ago](https://github.com/sosy-lab/java-smt/issues/276).
[...]

PR #621, opened 2026-03-20
620 smtlib2 parser tokenizer wrongly rejects set logic commands and has a non helpful error message for parse for 1 items

### LINKED ISSUE #620: SMTLIB2 parser (tokenizer) wrongly rejects set-logic commands and has a non-helpful error message for parse() for > 1 items
Our SMTLIB2 parser, to be exact its tokenizer, rejects `set-logic` commands that are not at the very beginning of the SMT-LIB2 file. This is wrong. ~~these commands are allowed to be everywhere in the file and may be repeated with certain settings.~~ `set-logic` has 2 restrictions; it may not be used again as long as the `reset` command has not been used, and it must be used before any `define-*`, `declare-*`, or `assert` commands are used. But many solvers don't stick to this, leading to us usually implicitly assuming that the `ALL` logic is set.
For benchmarks the `set-info` command required to be in the very first line (or to be the first command?), and it needs to set the `:smt-lib-version` attribute. But this is nothing we can or should require, as not everything is a benchmark.

Also, the `parse()` method returns a not so helpful exception if there is more than 1 returned formula.

baierd:
Fixes issues from #620 by adding a state-machine for processing SMT-LIB2 programs.

Closes #620.


src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -399,9 +399,17 @@ private String sanitize(String formulaStr) {
...
-      if (Tokenizer.isSetLogicToken(token)) {
-        // Skip the (set-logic ...) command at the beginning of the input
-        Preconditions.checkArgument(pos == 0);
+      if (pos == 0 && Tokenizer.isSetInfoToken(token)) {
+        // set-info call is allowed to be the very first command in the benchmark, but can also
+        // appear repeatedly throughout a SMT2 program
+        // Technically it is even required to be the very first command, and it must set the
+        //  :smt-lib-version attribute. See SMT-LIB 2.7 §3.11.3
+        // TODO: check that version >= 2?
+        Preconditions.checkArgument(token.contains(":smt-lib-version"));
```
daniel-raffler: I believe this only applies to *their* benchmark files, not Smtlib in general. Requiring `:smt-lib-version` to be set will cause crashes for most inputs. Including formulas that we printed and are now trying to read back

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -399,9 +399,17 @@ private String sanitize(String formulaStr) {
...
+
+      } else if (Tokenizer.isSetLogicToken(token)) {
+        // Skip (set-logic ...) commands for now. They may appear throughout an SMT-LIB2 program
+        // to set the 'current logic' to a new logic. Default (no command needed): ALL.
```
daniel-raffler: As mentioned, `set-logic` is only allowed as long as nothing has been pushed yet. Technically, it has to be set before doing anything else, but we can just assume logic `ALL` if the command is missing. There is also no way to change the logic later, unless `reset` was used, which effectively restarts the solver

daniel-raffler (changes requested):
> Our SMTLIB2 parser, to be exact its tokenizer, rejects set-logic commands that are not at the very beginning of the SMT-LIB2 file. This is wrong, these commands are allowed to be everywhere in the file and may be repeated with distinct settings!

I think this is backwards. Here is the diagram from the standard:

Image

(Source of diagram: https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf , 4.1 General Requirements, page 53)

So `set-logic` must be used before anything else can be done, and `set-info` can appear anywhere in the file

Did you run into any parsing issues with their benchmarks?

ThomasHaas:
> Did you run into any parsing issues with their benchmarks?

The issues he faces come from the smtlib files I gave in #613. The output we generate conforms to the shape required by SMT-COMP because we once submitted our benchmarks there.

daniel-raffler:
> > Did you run into any parsing issues with their benchmarks?
> 
> The issues he faces come from the smtlib files I gave in #613. The output we generate conforms to the shape required by SMT-COMP because we once submitted our benchmarks there.

Thanks! I've had a look at it, and there seem to be two issues:
- We expect `set-logic` to be the first command in the file
- We currently don't support `push`/`pop` while parsing

In this case the problem is easily fixed by just commenting out some lines:
[manyRegReads.smt2.txt](https://github.com/user-attachments/files/26157107/manyRegReads.smt2.txt)

When I now try the tests, I get a runtime of 8s with logic `ALL` and 3s for the BV logic. However, this is only true after adding back in the push/pop around the assertions. Otherwise, the runtime is <1s for both cases, and the BV logic appears to be slightly slower

@baierd: We could easily fix the first problem by skipping `set-option`/`set-info` before the `set-logic` command. However, I'm not sure how `push`/`pop` could be handled by our parser

ThomasHaas:
> In this case the problem is easily fixed by just commenting out some lines: [manyRegReads.smt2.txt](https://github.com/user-attachments/files/26157107/manyRegReads.smt2.txt)
> 
> When I now try the tests, I get a runtime of 8s with logic `ALL` and 3s for the BV logic. However, this is only true after adding back in the push/pop around the assertions. Otherwise, the runtime is <1s for both cases, and the BV logic appears to be slightly slower

If the query is not incremental, Z3 will use the SAT Solver independently of the logic set (easy to see by executing Z3 with higher verbosity levels). I think most solvers ignore `(set-logic)` for non-incremental queries: they might only check syntactic conformance to the specified logic but otherwise exploit the fact that they know the whole query and what settings are best for solving it.
But for incremental queries, the solver has do expect the worst-case, i.e., that every further query introduces new logics.
This is particularly true over the programmatic API, where the solver simply cannot know about future queries; all he knows is the upper bound on logics given by `(set-logic)`.

> However, I'm not sure how push/pop could be handled by our parser

I think you would need a more general concept of an `SMTQuery` which is a list of `SMTCommands: assert(formula), push, pop, check-sat`. I don't think the existing API can do it.

src/org/sosy_lab/java_smt/api/FormulaManager.java
```diff
@@ -193,6 +193,9 @@ default BooleanFormula makeDistinct(Formula... pArgs) {
   default BooleanFormula parse(String s) throws IllegalArgumentException {
     List<BooleanFormula> formulas = parseAll(s);
     checkArgument(!formulas.isEmpty(), "No assertion found in the SMTLIB string.");
+    checkArgument(
+        formulas.size() == 1,
+        "More than one assertion found in the SMT-LIB string, " + "please use parseAll(String).");
```
kfriedberger: String concatenation not required. Remove the PLUS.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
kfriedberger: According to the SMTLIB standard ([v2.6 from 2021](https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf)), `set-logic` belongs to the section `4.2.1 (Re)starting and terminating`. SMTLIB has a `reset` comand and allows multipe appearences of `set-logic`, where only the most recent logic is actually used.  I could not find direct evidence that forbits its appearence after a `push` or `assert` comand, except the header of section 4.2.1 in the standard. Nevertheless, most solvers might not expect it at random locations. JavaSMT does not expect and not handle the `reset` command when parsing SMT queries and skips `set-logic` anyway.

**Update**: I found the given diagram from above at `4.1 General Requirements` page 53 in the SMTLIB standard ([v2.6 from 2021](https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf)). `set-logic` may only appear between `start mode` and `assert mode`. We might have some missing comands and states in our parser here. 😄

kfriedberger (dismissed):
lgtm.

Why does the formatter or Checkstyle not warn about the String concatenation?
Please fix, then merge. Thanks.

baierd:
> So `set-logic` must be used before anything else can be done, and `set-info` can appear anywhere in the file

If you want to solve the query, it is needed before something is asserted. Which is less strict. 
But we don't want to solve the query, we parse it to a formula. Hence it does not matter for us currently, as Karlheinz already stated.

Btw. i left the token detection in for if we actually decide that we want to do something with it + documentation.

The standard however states that `For benchmarks in the official repository a call to set-info with this attribute
can occur only as the first command of a script.`, hence i wanted to allow it explicitly.

@ThomasHaas do you know whether comments are allowed to be placed before this `set-info` command?
Also, thank you for the clarification!

baierd:
> However, I'm not sure how push/pop could be handled by our parser.

If there is only `push` and `pop`, everything is easy, you just need a list of lists of formulas. The problem occurs when `check-sat` is used, because that can not be easily reflected with generic data-structures without being easily confusing or wrongly used.

We could just add a class `ParsingResult` that returns the information in a proper form. This could easily also reflect the other commands like `set-logic` as well. With such a structure it would also be easy to add Prover support (i.e. solve the formulas as dictated in the SMTLIB2), and therefore a parsing command that also solves could easily be added.

ThomasHaas:
> @ThomasHaas do you know whether comments are allowed to be placed before this `set-info` command? Also, thank you for the clarification!

I'm not sure if this is permitted in SMT-COMP, but does it really matter? You want to parse general smtlib2 which is far less restrictive: you can have comments everywhere and `(set-info)` doesn't need to be called at all.
Only when dumping whole queries to smtlib you might want to adhere to a more strict form.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
baierd: Correct. But that's not what we are checking, we are only checking that if there is a `set-info` in the very first line, then it has to be this kind of `set-info`. But i agree that since this is only meant for competition benchmarks we should actually not do it! I'll change it.

daniel-raffler:
> If you want to solve the query, it is needed before something is asserted. Which is less strict.
> But we don't want to solve the query, we parse it to a formula. Hence it does not matter for us currently, as Karlheinz already stated.

But shouldn't we at least try to keep to the standard? If we just want to allow `set-info` before `set-logic`, it's easy enough to just add a flag:

Screenshot From 2026-03-21 19-26-28

There is really no need to remove the check for that

> The standard however states that For benchmarks in the official repository a call to set-info with this attribute can occur only as the first command of a script., hence i wanted to allow it explicitly.

The problem here is that MathSAT adds `(set-info :source |printed by MathSAT|)` at the top of its output when dumping formulas. This will then cause the new precondition to fail:

```Java
private String sanitize(String formulaStr) {
    List<String> tokens = Tokenizer.tokenize(formulaStr);

    StringBuilder builder = new StringBuilder();
    int pos = 0; // index of the current token

    for (String token : tokens) {
      if (pos == 0 && Tokenizer.isSetInfoToken(token)) {
        // set-info call is allowed to be the very first command in the benchmark, but can also
        // appear repeatedly throughout a SMT2 program
        // Technically it is even required to be the very first command, and it must set the
        //  :smt-lib-version attribute. See SMT-LIB 2.7 §3.11.3
        // TODO: check that version >= 2?
        Preconditions.checkArgument(token.contains(":smt-lib-version"));
```
The token is `(set-info :source |printed by MathSAT|)` here, and does not contain `:smt-lib-version`

We could probably fix this, but I'm not sure the check is even needed. According to the standard these are additional requirements for their benchmark files:

>Benchmarks in the official SMT-LIB repository at www.smt-lib.org must satisfy additional
requirements on the meta-level information they contain and the order in which it appears.
>- `:smt-lib-version`, `:source`, `:license`, and `:category` must be set exactly once,
>- `:status` must be set as many times as needed so that each occurrence of the command
check-sat-assuming or check-sat in the benchmark is preceded (not necessarily immediately) by a corresponding `:status` info
>
>Moreover, the set set-info call for attribute :smt-lib-version must be the very first com-
mand in the benchmark.

We're trying to parse "regular" SMTLIB files, so this shouldn't apply to us

> If there is only push and pop, everything is easy, you just need a list of lists of formulas. The problem occurs when check-sat is used, because that can not be easily reflected with generic data-structures without being easily confusing or wrongly used.

Yes, that might be one way to solve it. In this particular example, however, the interesting assertion isn't even on the stack anymore at the end of the SMTLIB file. Instead, it was replaced by `false`, which makes the problem trivially unsat. We would need some way to get the assertions at different points in time. However, this might just become a little to confusing for the user

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
baierd: The question is whether we want to track all of them.

The following commands are allowed to be used before the very first `set-logic` command:
echo, reset, reset-assertions, get-info, get-option, set-info, set-option

Whats not allowed is:
assert, declare*, define*, and many more, depending on the state according to the diagram.

The best solution is most likely just tracking the state switching commands and tracking the current state. I'll try to work something out.

baierd:
> But shouldn't we at least try to keep to the standard?

Yes and no. We try to follow the standard as far as it makes sense. Many solvers do non-standard things and we adapt.

> If we just want to allow set-info before set-logic, it's easy enough to just add a flag:

No. This is a clear example for a state-machine. Starting with flags etc. would lead to confusing/complex code.

And yes, as stated above, the rule with `set-info`  having to specify `:smt-lib-version` first is only relevant for the competition, i will remove the precondition.
I find it curious that the standard specifies things about the competition though.

baierd:
So the problem is the following: if we assume the diagram above to be accurate, most solvers don't produce valid SMT-LIB2. Since we know this, we allow some leniency, most notably; we assume implicit logic selection for logic `ALL` if there is none. This merges the 2 states essentially. Since we are only parsing, and not solving, the `SAT_MODE` and `UNSAT_MODE` are essentially merged into `RESULT_MODE` for us. And since we are not allowing `reset` currently, what is left is a very limited automaton with 2 states.

Btw. i also found that the diagram above is incomplete, as `exit` is for example not handled. But it is quite clear how it should be handled (additional sink state).

I now added a implementation that can easily switch between all of this, depending on a single boolean flag to control whether we allow implicit `set-logic 'ALL'`. 


@daniel-raffler i found that Boolector seems to return an empty SMTLIB2 String in a test case (that i now disabled). Could you open a issue and take a look?

Also, i found that we throws exceptions every time we load Bitwuzlas options due to our "further options" implementation (in our code, not Bitwuzlas). While not part of this PR, I still added a small simplification to check for an empty options string to skip the evaluation (as it made my debugging harder ;D). @daniel-raffler would you be so kind and open another issue for this and take a look? Exception based control flow is Java is a very bad idea and we should not do it if possible.

baierd:
~~TODO:  is a SMTLIB2 program with only `(check-sat)` a valid program if we assume that the logic is set implicitly? (Because we don't allow this currently)~~

Edit: nvm we actually have tests for this. I need to change the current behavior.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
```diff
@@ -396,21 +402,38 @@ private String sanitize(String formulaStr) {
     List<String> tokens = Tokenizer.tokenize(formulaStr);
 
     StringBuilder builder = new StringBuilder();
-    int pos = 0; // index of the current token
+
+    // SMTLIB2ProgramState tracks that the SMTLIB2 query conforms to the rules outlined in the
+    // standard. SMTLIB2ProgramStateMachine models the
```
daniel-raffler: This comment seems to have been cut off

daniel-raffler (dismissed):
Looks fine to me! There are a few `TODOs` left open, and a lot of the methods in `SMTLIB2ProgramStateMachine` are currently unused. Do you plan to extend this later? Otherwise, we could probably simplify the class a bit

baierd:
> ... a lot of the methods in `SMTLIB2ProgramStateMachine` are currently unused. Do you plan to extend this later? Otherwise, we could probably simplify the class a bit

I wanted to have this as complete as possible so that if we decide that we need this (e.g. once we integrate the solver independent parser branch(es)) we are already prepared.

src/org/sosy_lab/java_smt/basicimpl/AbstractFormulaManager.java
baierd: Thanks! I updated it.

baierd:
@kfriedberger this should be ready for a second review/merge.

daniel-raffler (approved):
LGTM

PR #622, opened 2026-03-20
Update OpenSMT to version 2.9.2

daniel-raffler:
Hello,

this is a minor update of OpenSMT that doesn't include any major API changes. It does bring some bug fixes and internal changes compared to our current version:

>### 2.9.2 (2025-06-16)
>
>#### New features:
>
>    Supporting unsat-core options in the command line executable.
>
>### 2.9.1 (2025-06-06)
>
>#### Bug fixes:
>
>    Fixed assertion violation for LIA.
>    Fixed handling of split clauses.
>
>##### Other:
>
>    Optimized and sanitized LA substitutions.
>    Unified functions for polyToPTRef and back.

### DIFF
--- a/doc/Developers-How-to-Release-into-Ivy.md
+++ b/doc/Developers-How-to-Release-into-Ivy.md
@@ -201,9 +201,9 @@ Example:
 ```bash
 ant publish-opensmt \
     -Dopensmt.path=/workspace/solvers/opensmt/opensmt \
-    -Dopensmt.customRev=2.9.0 \
-    -Dgmp-linux-x64.path=/dependencies/gmp-6.2.1/install/x64-linux \
-    -Dgmp-linux-arm64.path=/dependencies/gmp-6.2.1/install/arm64-linux \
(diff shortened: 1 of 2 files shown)


kfriedberger:
@daniel-raffler : Thanks for preparing this well-done PR. 👍 

For me, it looks like the additional option `-j8` in the build-script has no visible effect when building the solver. This is no issue for us, as building is done only once in a while. Let's keep the option, maybe the solver's makefile gets improved for parallel build steps in the future.

daniel-raffler:
Thanks for handling the update!

> For me, it looks like the additional option `-j8` in the build-script has no visible effect when building the solver. This is no issue for us, as building is done only once in a while. Let's keep the option, maybe the solver's makefile gets improved for parallel build steps in the future.

Yes, I think you're right

The Makefile in OpenSMT actually just seems to be a wrapper for calling CMake. We'll have to set a flag there if we want a parallel build

PR #630, opened 2026-03-27
Add an option to reject/log Smtlib inputs with illegal commands

daniel-raffler:
Hello,

this PR extends the Tokenizer to recognize all commands from the SMTLIB standard. We also add a configuration option `parser.illegalInput` to let the user decide how to handle illegal commands. Currently, we simply skip over unrecognized/uninteresting commands while parsing. This can hide potential bugs or lead to confusing error messages. With the new option, skipping is still the default behavior. However, it's also possible to log illegal commands or abort the parse entirely by setting the option to a different value

The idea is to eventually make the Tokenizer strict and always throw an exception. However, we may still have to add more non-standard commands first


daniel-raffler:
@baierd Should I still split up the option for parsing/dumping? Or is this fine with you?

baierd:
I generally agree with the idea of this PR (adding an option that configures the parser), but i think we should combine this with #621, as the other PR provides an approach that can handle all situations.

Note: i think dumping also should have an option in theory, but in practice this is currently not really needed. At least as far as i know. Should be have cases in which we need output that strictly conforms to SMTLIB2s definition, then we should of course do it.

PR #634, opened 2026-03-29
Update Intellij project files for Java17

daniel-raffler:
Hello,

this PR tries to update the Intellij project files for Java17. It's a draft for now, as I'm not sure if more changes that will be needed

@baierd 
i believe you're also using Intellij? Could you please give it a try to see if it's working?


.idea/misc.xml
```diff
@@ -17,36 +17,43 @@ SPDX-License-Identifier: Apache-2.0
...
-          <item index="3" class="java.lang.String" itemvalue="android.support.annotation.Nullable" />
-          <item index="4" class="java.lang.String" itemvalue="javax.annotation.CheckForNull" />
+        <list size="14">
+          <item index="0" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.qual.Nullable" />
+          <item index="1" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NullableDecl" />
+          <item index="2" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NullableType" />
+          <item index="3" class="java.lang.String" itemvalue="javax.annotation.Nullable" />
+          <item index="4" class="java.lang.String" itemvalue="android.support.annotation.Nullable" />
           <item index="5" class="java.lang.String" itemvalue="androidx.annotation.Nullable" />
           <item index="6" class="java.lang.String" itemvalue="androidx.annotation.RecentlyNullable" />
```
kfriedberger: JavaSMT does not depend on the package `androidx`. We might want to remove these line.

kfriedberger (dismissed):
The entry for JDK is a good catch. CPAchecker does not use the JDK-entry, see https://github.com/sosy-lab/cpachecker/blob/main/.idea/ , and seems to work well, too.

Overall, if that helps someone, ok for me.

baierd:
I reworked the IntelliJ configs of CPAchecker last year and can take a look or even do it myself quickly on the bases of CPAchecker. But we can't just re-use them from other projects, as the setup needs to be specific. (It took me quite a while to understand how this works, as the documentation of IntelliJ IDEA was very incomplete on several issues. The behavior of a IntelliJ configuration that is only partially correct, as every IntelliJ user knowns, is that things (e.g. dependencies) are not automatically updated or pulled, licenses are removed/added in config files all the time randomly, the configs are changed constantly by IntelliJ despite settings that it should not etc.)

baierd:
@daniel-raffler Review here instead of the code:
(I think it is a good idea to have someone in the team that understands how this works besides me, so I'll provide a review for you to work on.)

In CPAchecker i used Ant to get all information about classpaths etc. by adding the flag `-verbose`.
Then i transfered the information to the sections relevant in the IntelliJ IDEA configs. This ensures that both do (roughly) the same. 

You can get the classpath info by searching for the string `classpath` in the output for example. 
Other relevant parts for dependencies are sources and native libraries. This is specified in IntelliJ IDEAs `Project Settings` -> `Modules`. 
Important: we may only load dependencies that are loaded when compiling using Ant! Many projects are present multiple times and with distinct source-locations. This can cause conflicts in IntelliJ IDEA!

If we want to do this properly, we need to define modules for all solvers and their dependencies. Since i don't know how this works if we can't load a solver, we need to test this thoroughly!

Running Ant in verbose mode seems to indicate a problem with `OptionCollector` by the way.

Further, a default copyright header can be defined in `Settings` -> `Copyright` -> `Formatting` -> `Java`. This should be done as in CPAchecker.

Also, we can get rid of many edits of IntelliJ IDEA on its own configuration files by extracting the licenses in them into a file, for example like [this](https://gitlab.com/sosy-lab/software/cpachecker/-/blob/main/.idea/CPAchecker.iml.license?ref_type=heads).

Also, we should define which nullness annotations we actually want to use. This is defined in our Ant configs and IntelliJ should not diverge. Example from CPAchecker [here](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/341/diffs#2e3fa18a7568df36e24e300a656f185b38c3633c_5_5). This can be set in the settings (no need to touch the config code directly). I also don't see a reason to load all annotation frameworks. It is sufficient to load only the ones we need ([see lines following this](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/341/diffs#2e3fa18a7568df36e24e300a656f185b38c3633c_6_7)).

We should set all flags that stop IntelliJ from editing itself all the time. That's not perfect, but helps a lot (or i did not find all flags? But it is a common problem reported online, so i don't know....). I don't remember where you can set this in the settings, but the config change in CPAchecker can be seen [here for example](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/341/diffs#6ecd6000a7b6f4a2884412ff19f74193ed089648_12_13).

I don't see a reason to change the compiler or compiler settings in JavaSMT. That was only needed in CPAchecker due to problems that are not present in JavaSMT!

You can find the MR from CPAchecker [here](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/341). Some things are not optimal there, and some are not applicable for JavaSMT. E.g. i used a single module dependency for CPAchecker, but JavaSMT should define multiple.

.idea/misc.xml
daniel-raffler: Hello Karlheinz,
thanks for the hint! I've now removed some of the nullable annotations we're not using. For the rest the "remove" button is greyed out in the GUI, maybe because they're used by one of our dependencies?

From what I understand, this is just a list of annotations that Intellij will recognize for its own nullable analysis. So having some additional ones in there shouldn't cause too many problems

daniel-raffler:
@baierd Thanks for the summary!

I've already split off the license headers and cleaned up the `@Nullable` annotations. The project now uses a separate `library` dependency for each solver backend that has a `*.jar` file. Solvers that only bundle native code don't need a dependency as we use `NativeLibrary` for loading anyway. (It would be possible to still add one, but we would have to pick a CPU architecture)

Whenever possible I also included the `*-sources.jar` in the library dependencies, so that it can be found when running the debugger. However, we're currently missing sources for some solvers, and I wasn't able to step into the OpenSMT jar (maybe we compiled without debugging info?)

> Important: we may only load dependencies that are loaded when compiling using Ant! Many projects are present multiple times and with distinct source-locations. This can cause conflicts in IntelliJ IDEA!

Hmm.. I'm not sure, but shouldn't Ivy already have taken care of these conflicts?

.idea/ant.xml.license
```diff
@@ -0,0 +1,9 @@
+<!--
+This file is part of JavaSMT,
+an API wrapper for a collection of SMT solvers:
+https://github.com/sosy-lab/java-smt
+
+SPDX-FileCopyrightText: 2020 Dirk Beyer <https://www.sosy-lab.org>
```
baierd: Would you be so kind and also update the years in all licenses touched by this PR?

baierd (changes requested):

.idea/misc.xml
baierd: [In CPAchecker](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/341/diffs#2e3fa18a7568df36e24e300a656f185b38c3633c_5_4) i kicked out all of the unused annotation preprocessors. We rely on the (to my knowledge) strongest nullable preprocessor in `checkerframework` anyway. IntelliJ had lots of false-positives/negatives before i changes this in CPAchecker for me.

daniel-raffler: Ok, I've removed the other Nullable/NotNull annotations now

.idea/ant.xml.license
daniel-raffler: Should be updated now :+1:

.idea/libraries/princess.xml
```diff
@@ -0,0 +1,16 @@
+<component name="libraryTable">
```
baierd: I tested the branch and found the same problem that i found in CPAchecker a while back, in that IntelliJ fails to load the dependencies if they were not present when the initial build (and subsequent partial re-builds) are triggered.

IntelliJ refuses to load all declared dependencies, like the ones marked here, if one fails to load. My guess (from observation and reading lots of sources/posts online, including the official documentation) is: 
IntelliJ will try to load all dependencies, and if some are missing it will just stop and NOT reload them automatically once they are present. (You can force it to load them, but this is dependent on a new build being triggered with all files being present, which can not be automated according to the official documentation.

The solution is to pack all dependencies (sources and JavaDoc) in a single dependency (either on module or library level). This forces IntelliJ to re-load everything for a change! 
Of course this solution is not documented 🙄 but since i had to re-discover this just now because i tested this branch and noticed that this is happening again, i will now document this for our projects at least.

I'll explain module based dependencies, as that is the default.
Usually we have one module that defines multiple distinct dependencies, nicely split by their folders or other identifiers. 

Image

Problem: it seems like IntelliJ does not load all of these if it finds a problem with one. So especially if you set up your project, you will be stuck with only partially available dependencies. This can be fixed manually by triggering a full re-build of the project with all dependencies being present, but this can't be automated (according to IntelliJs documentation).

Defining only a single dependency that includes classes (Jars), sources, native libraries, as well as JavaDoc,  solves this:
Image
Image

You might notice that there is only a single JavaDoc source defined here. That is due to 2 problems:
- if you give it the folder that includes sources and JavaDoc, it just recognizes it as a source folder, but you can't also mark it as a JavaDoc folder
- if you mark multiple JavaDoc Jars and add them, it also fails to do so, adding only one
Solutions: either add them one by one using the GUI. Or even better, edit the correct file (`MODULE_NAME.iml` for module based dependency management), as this allows you to define JavaDoc, sources, native lib. folders etc. directly.

baierd: I already fixed it.

.idea/ant.xml.license
baierd: Thank you!

.idea/misc.xml
baierd: This should solve this. Thanks!

baierd (approved):

baierd:
@daniel-raffler nice work!

PR #636, opened 2026-03-30
Yices: Add an option to set the logic

daniel-raffler:
Hello,

this PR adds two options that allow the user to set the logic and enable/disable iterative solving. Yices will then pick a solver based on the value of these two options:
- If the logic requires non-linear arithmetics, MCSat will be chosen. Otherwise DPLLT is used
- If the logic is set to `QF_IDL`/`QF_RDL` and `one-shot` mode is used, Yices may use the Floyd-Warshal solver. Otherwise, simplex is used

The PR also removes the earlier `solver-type` option as solver selection is now done automatically. Note, that this effectively makes DPLLT the default solver again

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/yices2/Yices2InterpolatingProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/yices2/Yices2InterpolatingProver.java
@@ -17,6 +17,8 @@ import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
+import com.sri.yices.Config;
+import com.sri.yices.Context;
 import com.sri.yices.InterpolationContext;
 import com.sri.yices.Status;
 import com.sri.yices.Terms;
@@ -31,6 +33,7 @@ import org.sosy_lab.java_smt.api.BooleanFormulaManager;
 import org.sosy_lab.java_smt.api.InterpolatingProverEnvironment;
 import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 import org.sosy_lab.java_smt.api.SolverException;
+import org.sosy_lab.java_smt.solvers.yices2.Yices2SolverContext.Yices2Parameters;
 
 class Yices2InterpolatingProver extends Yices2AbstractProver<Integer>
     implements InterpolatingProverEnvironment<Integer> {
@@ -39,8 +42,8 @@ class Yices2InterpolatingProver extends Yices2AbstractProver<Integer>
       Set<ProverOptions> pOptions,
       BooleanFormulaManager pBmgr,
       ShutdownNotifier pShutdownNotifier,
-      String pSolverType) {
-    super(creator, pOptions, pBmgr, pShutdownNotifier, pSolverType);
+      Yices2Parameters pSolverParameters) {
+    super(creator, pOptions, pBmgr, pShutdownNotifier, pSolverParameters);
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


src/org/sosy_lab/java_smt/test/ModelTest.java
```diff
@@ -2747,10 +2743,10 @@ public void testArrayStore2BvBvBv() {
 
     assume().that(solver).isNotEqualTo(Solvers.BOOLECTOR); // Doesn't support multiple indices
     assume()
-        .withMessage("Yices does not support constant arrays when mcsat is used")
+        .withMessage("Yices returns an empty array here")
         .that(solver)
-        .isNotEqualTo(Solvers.YICES2);
-    // FIXME CVC4 array model is sometimes broken in JavaSMT. Unfixable in CVC4, fixed in CVC5.
+        .isNotEqualTo(Solvers.YICES2); // FIXME
```
baierd, line 2748: This looks like it misses a line-break now.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2InterpolatingProver.java
```diff
@@ -66,10 +69,18 @@ public BooleanFormula getInterpolant(Collection<Integer> formulasOfA)
             transformedImmutableSetCopy(setB, stack.peekLast()::get)));
   }
 
+  private Context newSolverStack() {
+    try (var cfg = new Config()) {
+      cfg.set("solver-type", "mcsat");
```
baierd, line 74: Could you add a little description why we need this? Maybe with a link to the documentation.

src/org/sosy_lab/java_smt/solvers/yices2/Yices2SolverContext.java
```diff
@@ -28,12 +28,61 @@
 public final class Yices2SolverContext extends AbstractSolverContext {
 
   @Options(prefix = "solver.yices2")
-  private static class Yices2Parameters {
+  static class Yices2Parameters {
     @Option(
         secure = true,
-        description = "Backend to use for the solver",
-        values = {"dpllt", "mcsat"})
-    String solverType = "mcsat";
```
baierd, line 36: I guess `solverType` is now deprecated? Usually we want to keep it, mark it as deprecated and then throw if someone uses it, but i am unsure whether this is possible with our options system (maybe @PhilippWendler can tell us :D)

src/org/sosy_lab/java_smt/solvers/yices2/Yices2SolverContext.java
```diff
@@ -28,12 +28,61 @@
...
+        description = "Sets the SMTLIB logic for solver selection",
```
baierd, line 34: As far as i understood it, and please correct me if i am wrong, `MCSAT` is now selected via the correct logic? Or is it just the only option? What about `dpllt`?

src/org/sosy_lab/java_smt/solvers/yices2/Yices2InterpolatingProver.java
```diff
@@ -66,10 +69,18 @@ public BooleanFormula getInterpolant(Collection<Integer> formulasOfA)
...
+      cfg.set("model-interpolation", "true");
+      return new Context(cfg);
+    }
+  }
+
   private int interpolate(Collection<Integer> setA, Collection<Integer> setB)
       throws InterruptedException {
-    try (var ctxA = newContext("mcsat");
-        var ctxB = newContext("mcsat")) {
+    try (var ctxA = newSolverStack();
```
baierd, line 82: Aren't we missing the logic (and other options) in those 2 new solver stacks?

baierd (changes requested):
Nice work! Some small questions remain, but we can merge soon.

PR #647, opened 2026-04-12
Bitwuzla: Optimize memory usage

### LINKED ISSUE #646: Bitwuzla: Optimize AllSat implementation
Hello,

our current `AllSat` implementation consumes too much memory on Bitwuzla (and to a lesser degree CVC5), see [this](https://gitlab.com/sosy-lab/software/cpachecker/-/work_items/1603#note_3234625372) discussion. The problem is, that the AllSat wrapper generates a full model for each iteration of the AllSat loop. This is a very memory intensive operation for Bitwuzla as we have to generate Java proxies for all values in the model. These proxies then can't be collected by the garbage collector until the context is closed and will quickly pile up when there are many models to cycle through

A better solution would be to only return an `Evaluator` in the AllSat loop. This evaluator can then be used to check the `importantPredicates` for this solution without having to evaluate the full model

Alternatively, we could consider switching to the fallback in `AbstractProverWithAllSat.allSat`:
```java
  @Override
  public <R> R allSat(AllSatCallback<R> callback, List<BooleanFormula> importantPredicates)
      throws InterruptedException, SolverException {
    checkGenerateAllSat();

    push();
    try {
      // try model-based computation of ALLSAT
      iterateOverAllModels(callback, importantPredicates);
    } catch (SolverException e) {
      // fallback to direct SAT/UNSAT-based computation of ALLSAT
      iterateOverAllPredicateCombinations(callback, importantPredicates, new ArrayDeque<>());
      // TODO should we completely switch to the second method?
    }

    pop();
    return callback.getResult();
  }
```

daniel-raffler:
Hello,

this PR adds two optimizations for Bitwuzla to bring down memory usage:
- it adds an `Evaluator` to skip model generation in `allSat`
- it further optimizes `extractVariablesAndUFs` to avoid creating unnecessary Java proxies in the visitor

Here are some results from SVCOMP with a 60s time limit: [SV-COMP26_no-overflow](https://github.com/user-attachments/files/26657072/results.2026-04-12_11-51-12.table.html)

The left column is before this PR, the middle is with the new Evaluator, and the right is with the `extractVaraiblesAndUFs` optimization (and the Evaluator)

And here's the scatter plot for a before-after comparison:
Screenshot From 2026-04-12 12-16-21

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
+++ b/src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
@@ -169,7 +169,12 @@ abstract class BitwuzlaAbstractProver<T> extends AbstractProverWithAllSat<T> {
     checkGenerateModels();
     // special case for Bitwuzla: Models are not permanent and need to be closed
     // before any change is applied to the prover stack. So, we register the Model as Evaluator.
-    return getEvaluatorWithoutChecks();
+    return registerEvaluator(
+        new BitwuzlaModel(
+            env,
+            this,
+            creator,
+            Collections2.transform(getAssertedFormulas(), creator::extractInfo)));
   }
 
   /**
@@ -225,15 +230,9 @@ abstract class BitwuzlaAbstractProver<T> extends AbstractProverWithAllSat<T> {
     super.close();
   }
 
-  @SuppressWarnings("resource")
   @Override
-  protected BitwuzlaModel getEvaluatorWithoutChecks() {
-    return registerEvaluator(
-        new BitwuzlaModel(
-            env,
-            this,
-            creator,
-            Collections2.transform(getAssertedFormulas(), creator::extractInfo)));
+  protected BitwuzlaEvaluator getEvaluatorWithoutChecks() {
+    return registerEvaluator(new BitwuzlaEvaluator(this, creator));
   }
 
   public boolean isClosed() {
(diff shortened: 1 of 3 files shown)


baierd:
LGTM! Nice work! 

I'll start a 900s run in CPAchecker that fits the initial problem that uncovered the problem. That way we can compare it properly (i already have results for current JavaSMT `master`).

kfriedberger (approved):
lgtm.

baierd:
Results (w comparison to before) for the overflow category: 
[bitwuzla.overflow.cmp.results.2026-04-13_00-23-37.table.html](https://github.com/user-attachments/files/26662692/bitwuzla.overflow.cmp.results.2026-04-13_00-23-37.table.html)


No regressions and only 123 out-of-memories now (546 before).
We can still try to find out whether we can do something for the other tasks, but this PR can definitely be merged!
Good job!

daniel-raffler:
> Results (w comparison to before) for the overflow category: [bitwuzla.overflow.cmp.results.2026-04-13_00-23-37.table.html](https://github.com/user-attachments/files/26662692/bitwuzla.overflow.cmp.results.2026-04-13_00-23-37.table.html)
> 
> No regressions and only 123 out-of-memories now (546 before). We can still try to find out whether we can do something for the other tasks, but this PR can definitely be merged! Good job!

Thanks!

baierd:
Results for all `unreach-call` tasks in SV-Benchmarks for the old (JavaSMT 6.0.0) and new Bitwuzla: 
[bitwuzla.cmp.memory-fix.results.2026-04-14_16-58-36.table.html](https://github.com/user-attachments/files/26716509/bitwuzla.cmp.memory-fix.results.2026-04-14_16-58-36.table.html)

Roughly 1000 less out-of-memory, but still 600 left. @daniel-raffler would you be so kind and open a issue for further investigation, reference this improvement and take a look?

There is a regression of 23 tasks, but the used config is Value-Analysis centric. So we can ignore that.

PR #648, opened 2026-04-13
Update SMTInterpol and add new theories

### LINKED ISSUE #468: Update to SMTInterpol Version 2.5-1388
SMTInterpol is available in version 2.5-1388 while we still use the 3 year old version 2.5-1242.

I don't know the details of changes, but the new version does include changes to the proof format and potentially bitvector support.

daniel-raffler:
Hello,

this PR updates SMTInterpol to the latest Maven release (= version `2.5-1388`) and adds support for bitvectors and quantifiers

Updating SMTInterpol has been blocked for several years by conflicts with LassoRanker, see issue #468. However, there is now a PR ([link](https://github.com/ultimate-pa/ultimate/pull/712)) for Ultimate that makes SMTInterpol a regular dependency of LassoRanker, which fixes most of the conflicts. The remaining clashes are mostly in JavaCUP and some of the utility classes of SMTInterpol/Ultimate. Neither are likely to change in the future, and I couldn't find any issues when trying the new version with Benchcloud (See [here](https://github.com/user-attachments/files/26668587/results.2026-04-12_18-38-12.table.html): both columns use SMTInterpol `2.5-1388`, the left uses LassoRanker from the PR, on the right we're using the current version, which causes crashes due to conflicting classes in the jars)

### DIFF
--- a/src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
+++ b/src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
@@ -70,8 +70,12 @@ class FormulaCollectionScript implements Script {
 
   @Override
   public void declareFun(String fun, Sort[] paramSorts, Sort resultSort) throws SMTLIBException {
-    FunctionSymbol fsym = theory.getFunction(fun, paramSorts);
-
+    FunctionSymbol fsym = null;
+    try {
+      fsym = theory.getFunction(fun, paramSorts);
+    } catch (SMTLIBException e) {
+      // fsym = null
+    }
     if (fsym == null) {
       script.declareFun(fun, paramSorts, resultSort);
     } else {
@@ -89,8 +93,12 @@ class FormulaCollectionScript implements Script {
     for (int i = 0; i < paramSorts.length; i++) {
       paramSorts[i] = params[i].getSort();
     }
-    FunctionSymbol fsym = theory.getFunction(fun, paramSorts);
-
+    FunctionSymbol fsym = null;
+    try {
+      fsym = theory.getFunction(fun, paramSorts);
+    } catch (SMTLIBException e) {
+      // fsym = null
+    }
     if (fsym == null) {
       script.defineFun(fun, params, resultSort, definition);
     } else {
@@ -166,6 +174,11 @@ class FormulaCollectionScript implements Script {
   public Term quantifier(int quantor, TermVariable[] vars, Term body, Term[]... patterns)
       throws SMTLIBException {
     return script.quantifier(quantor, vars, body, patterns);
+  }
+
+  @Override
+  public Term lambda(TermVariable[] pTermVariables, Term pTerm) throws SMTLIBException {
+    return script.lambda(pTermVariables, pTerm);
   }
 
   @Override
--- a/src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
+++ b/src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
@@ -10,6 +10,7 @@ package org.sosy_lab.java_smt.solvers.smtinterpol;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.collect.FluentIterable;
 import de.uni_freiburg.informatik.ultimate.logic.Annotation;
 import de.uni_freiburg.informatik.ultimate.logic.ApplicationTerm;
 import de.uni_freiburg.informatik.ultimate.logic.Assignments;
@@ -28,6 +29,7 @@ import de.uni_freiburg.informatik.ultimate.logic.Theory;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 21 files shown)


daniel-raffler:
@kfriedberger 
SMTInterpol `2.5-1388` is already in Ivy, but we'll need a new version of LassoRanker for CPAchecker. Would you be willing to handle this? 

You can build the binaries from my fork [here](https://github.com/daniel-raffler/ultimate). It includes a script to patch the version numbers for us, and the PR for SMTInterpol has been updated for the latest Ultimate release. The process should look something like this:
```console
git clone git@github.com:daniel-raffler/ultimate.git
cd ultimate
git checkout smtinterpol-dependency
cd releaseScripts
./patchVersion 0.3.1.20260413
cd ../trunk/source/BA_MavenParentUltimate
mvn install source:jar
```
(The version string for `patchVersion` may only contain numbers, so I've used a date to tag the release. We'll have to update Ultimate every time there is an API breaking change to SMTInterpol, so just the version (= `0.3.1`) is probably not enough)

The last line will build Ultimate and then install all packages to your local Maven repository. You can then pull it from there into your Ivy repo. I simply changed `ivysettings.xml` in the repo for that:
```patch
daniel@notebook:~/workspace/ivy$ svn diff ivysettings.xml 
Index: ivysettings.xml
===================================================================
--- ivysettings.xml	(revision 3248)
+++ ivysettings.xml	(working copy)
@@ -11,8 +11,7 @@
         </filesystem>
         <chain name="extern">
           <ibiblio name="maven2" m2compatible="true"/>
-          <ibiblio name="jcenter" m2compatible="true" root="https://jcenter.bintray.com/" />
-            <!--          <ibiblio name="restlet" m2compatible="true" root="https://maven.restlet.org/" /> -->
+          <ibiblio name="local" m2compatible="true" root="file://${user.home}/.m2/repository"/>
         </chain>
     </resolvers>
 </ivysettings>
```
Now the Maven packages can be copied to Ivy like this:
```console
ant install -Dorganisation=de.uni_freiburg.informatik.ultimate -Dmodule=de.uni_freiburg.informatik.ultimate.lib.lassoranker -Drevision=0.3.1.20260413
```
Note that this will also pull some dependencies into the Ivy repository, which should probably be added separately:
```console
daniel@notebook:~/workspace/ivy$ svn status repository/
?       repository/ch.qos.reload4j
?       repository/com.sun.xml.bind/jaxb-osgi
M       repository/commons-io/commons-io/ivy-2.16.1.xml
M       repository/commons-io/commons-io/ivy-2.16.1.xml.md5
M       repository/commons-io/commons-io/ivy-2.16.1.xml.sha1
?       repository/de.uni_freiburg.informatik.ultimate
?       repository/net.sf.trove4j
M       repository/org.glassfish.hk2/osgi-resource-locator/ivy-1.0.3.xml
M       repository/org.glassfish.hk2/osgi-resource-locator/ivy-1.0.3.xml.md5
M       repository/org.glassfish.hk2/osgi-resource-locator/ivy-1.0.3.xml.sha1
?       repository/org.osgi/org.osgi.service.prefs
?       repository/org.osgi/osgi.annotation
```
Also note that the Ultimate packages are now under `de.uni_freiburg.informatik.ultimate` (with an `_` after `uni`, instead of a `-`). This is probably for the best as the structure of the modules has also changed. However, if we want to keep the old scheme, it should be possible to patch the names before building Ultimate

src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
```diff
@@ -70,8 +70,12 @@ public LBool assertTerm(Term pTerm) throws SMTLIBException {
...
   @Override
   public void declareFun(String fun, Sort[] paramSorts, Sort resultSort) throws SMTLIBException {
-    FunctionSymbol fsym = theory.getFunction(fun, paramSorts);
-
+    FunctionSymbol fsym = null;
+    try {
+      fsym = theory.getFunction(fun, paramSorts);
+    } catch (SMTLIBException e) {
+      // fsym = null
+    }
```
kfriedberger: Is ignoring the exception really the best way to go?

src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
```diff
@@ -89,8 +93,12 @@ public void defineFun(String fun, TermVariable[] params, Sort resultSort, Term d
...
       paramSorts[i] = params[i].getSort();
     }
-    FunctionSymbol fsym = theory.getFunction(fun, paramSorts);
-
+    FunctionSymbol fsym = null;
+    try {
+      fsym = theory.getFunction(fun, paramSorts);
+    } catch (SMTLIBException e) {
+      // fsym = null
+    }
```
kfriedberger: Is ignoring the exception really the best way to go?

src/org/sosy_lab/java_smt/solvers/smtinterpol/SmtInterpolFormulaCreator.java
```diff
@@ -99,8 +112,12 @@ public Term makeVariable(final Sort type, final String varName) {
...
   private FunctionSymbol declareFun(String fun, Sort[] paramSorts, Sort resultSort) {
     checkSymbol(fun);
-    FunctionSymbol fsym = environment.getTheory().getFunction(fun, paramSorts);
-
+    FunctionSymbol fsym = null;
+    try {
+      fsym = environment.getTheory().getFunction(fun, paramSorts);
+    } catch (SMTLIBException e) {
+      // fsym = null
+    }
```
kfriedberger: Is ignoring the exception really the best way to go?

kfriedberger (dismissed):
Nice work.
Lets update JavaSMT now, and handle conflicts CPAchecker with LassoRanker later, as soon as JavaSMT is released.

src/org/sosy_lab/java_smt/solvers/smtinterpol/FormulaCollectionScript.java
daniel-raffler: Good point! I've rewritten the function, and we're now checking for collisions before trying to declare the new symbol

baierd:
Completely unchecked and potentially untrue statement: wasn't the problem with updating SMTInterpol that it requires Java 21?

kfriedberger:
Seems to work well on Java 17: https://github.com/sosy-lab/java-smt/actions/runs/24358593939/job/71132099693

daniel-raffler:
> Completely unchecked and potentially untrue statement: wasn't the problem with updating SMTInterpol that it requires Java 21?

No, it's just Ultimate (i.e LassoRanker, Eliminator, etc) that requires Java 21. This should be fine, though, as I believe CPAchecker is also using Java 21

The new SMTInterpol version only requires Java 8

baierd:
Awesome! Thank you both for the clarification and also for the update!

PR #655, opened 2026-04-27
boolean formular manager: horn api

Bixilon:
This makes it easy to create horn clauses in the form: `(constraint \/ body) => head`.

This is the "base" for adding eldarica as an additional parser.

I am not sure what to put into the debugging delegate manager, as all called functions are from the debugging interface. Does it make sense to add all the assertions again?


baierd:
I am not sure whether we need this explicitly.

Spacer (Z3) simply does not work well in cases that are distinct from HORN clauses. But it does work well if you resolve the default form of horn-clauses into conjunctions for example.

Eldarica seems to also allow this according to [its tests](https://github.com/uuverifiers/eldarica/blob/master/src/test/scala/lazabs/horn/HornAPI.scala).

Bixilon:
Superseeded by #683

PR #660, opened 2026-05-10
Yices: Use official repository for building the Java bindings

daniel-raffler:
Hello,

this PR changes the build script for Yices so that the Java bindings are built directly from the official repository. We used to have a fork with some changes, but since everything has now been merged into `main` this will no longer be necessary

(see https://github.com/SRI-CSL/yices2_java_bindings/pull/8 for our pull request)

### DIFF
--- a/build/build-publish-solvers/solver-yices.xml
+++ b/build/build-publish-solvers/solver-yices.xml
@@ -194,10 +194,10 @@ SPDX-License-Identifier: Apache-2.0
 
     <target name="download-jdk21-windows" depends="build-initialize" unless="jdk.downloaded">
         <!-- Download a more recent Jdk to avoid compile errors -->
-        <get src="https://download.oracle.com/java/21/latest/jdk-21_windows-x64_bin.zip" dest="/dependencies" verbose="true"/>
-        <unzip src="/dependencies/jdk-21_windows-x64_bin.zip" dest="/dependencies"/>
+        <get src="https://download.oracle.com/java/21/archive/jdk-21.0.10_windows-x64_bin.zip" dest="/dependencies" verbose="true"/>
+        <unzip src="/dependencies/jdk-21.0.10_windows-x64_bin.zip" dest="/dependencies"/>
         <delete file="/dependencies/jdk-21_windows-x64_bin.zip" failonerror="true"/>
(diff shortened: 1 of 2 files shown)


build/build-publish-solvers/solver-yices.xml
```diff
@@ -194,16 +194,16 @@ SPDX-License-Identifier: Apache-2.0
 
     <target name="download-jdk21-windows" depends="build-initialize" unless="jdk.downloaded">
         <!-- Download a more recent Jdk to avoid compile errors -->
-        <get src="https://download.oracle.com/java/21/latest/jdk-21_windows-x64_bin.zip" dest="/dependencies" verbose="true"/>
-        <unzip src="/dependencies/jdk-21_windows-x64_bin.zip" dest="/dependencies"/>
+        <get src="https://download.oracle.com/java/21/archive/jdk-21.0.10_windows-x64_bin.zip" dest="/dependencies" verbose="true"/>
```
baierd, line 197: @PhilippWendler do we need to add a license of we use Oracle JDKs?

lib/native/source/yices2/jni.patch
```diff
@@ -75,48 +75,53 @@ diff --git a/src/main/java/com/sri/yices/Makefile b/src/main/java/com/sri/yices/
   $(error "Unkown OS: $(OS)")
 @@ -48,7 +50,11 @@
  endif
- 
+
  # name of the library
 -libyices2java := libyices2java.$(EXTENSION)
 +ifeq ($(OS),win32)
 + libyices2java := yices2java.$(EXTENSION)
```
baierd, line 82: `libyices2java` is hardcoded as a name several times below. Do we need to update those as well?

baierd (changes requested):

build/build-publish-solvers/solver-yices.xml
PhilippWendler: One always needs a license for every piece of software that one uses.

Whether the current license of the Oracle JDK allows us to use it for free I don't know, you have to check.

baierd: Thank you! We will take a look and add the appropriate license if needed.

PhilippWendler: What do you mean with "add"? Why and where would you want to "add" a license? The license is specified by Oracle, not us. We need to *be given* a license that allows us to use the software.

PR #662, opened 2026-05-22
Update solvers in our example projects

daniel-raffler:
Hello,

this PR updates the version numbers in our Maven example to the JavaSMT `6.0.0-148-gba08f432a` release. There is a newer version of JavaSMT in the Maven repository, however, Yices2 is broken on that one, so this is the last version that works

The PR now also updates the `web` version of the Maven example, and our two Gradle project templates, which have been unmaintained for several years


doc/Example-Gradle-Project-Kotlin/build.gradle.kts
```diff
@@ -12,27 +12,30 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
     // Apply the kotlin.jvm plugin to add support for Kotlin.
-    kotlin("jvm") version "1.6.21"
+    kotlin("jvm") version "2.3.21"
 
     // Apply the application plugin to add support for building a CLI application in Java.
     application
 }
 
-// Set the used Java version
-java.sourceCompatibility = JavaVersion.VERSION_11
```
baierd, line 22: Don't we want to set this to 17?

doc/Example-Gradle-Project/src/test/java/org/sosy_lab/java_smt_example/SudokuTest.java
```diff
@@ -69,35 +73,56 @@
...
+@RunWith(Parameterized.class)
 public class SudokuTest {
 
+  @Parameterized.Parameters(name = "{0}")
+  public static List<Solvers> getSolvers() {
+    return Arrays.asList(Solvers.values());
+  }
+
+  @Parameterized.Parameter(0)
+  public Solvers solver;
```
baierd, line 90: Does this need to be public? (Can be closed by you)

baierd (changes requested):
Nice work! Thank you!

Did you run/test the examples in a fresh checkout?

doc/Example-Gradle-Project/src/test/java/org/sosy_lab/java_smt_example/SudokuTest.java
daniel-raffler: Yes, `solver` is injected by JUnit, so the field needs to be public
```
SudokuTest > checkSudoku[YICES2] FAILED
    java.lang.IllegalAccessException: Cannot set parameter 'solver'. Ensure that the field 'solver' is public.
        at org.junit.runners.parameterized.BlockJUnit4ClassRunnerWithParameters.createTestUsingFieldInjection(BlockJUnit4ClassRunnerWithParameters.java:77)
        at org.junit.runners.parameterized.BlockJUnit4ClassRunnerWithParameters.createTest(BlockJUnit4ClassRunnerWithParameters.java:48)
        at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:260)
        at org.junit.runners.BlockJUnit4ClassRunner$2.runReflectiveCall(BlockJUnit4ClassRunner.java:309)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
        at org.junit.runners.Suite.runChild(Suite.java:128)
        at org.junit.runners.Suite.runChild(Suite.java:27)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
```

doc/Example-Gradle-Project-Kotlin/build.gradle.kts
daniel-raffler: Ideally yes, along with the option for Kotlin that was further down:
```
// compile bytecode to java 11
tasks.withType<KotlinCompile> {
    kotlinOptions.jvmTarget = "11"
}
```

Unfortunately the Kotlin option has been deprecated, and only setting the Java version causes the build to fail, which is why I removed both

However, I've done some digging now, and the latest version should now set the JVM target correctly

daniel-raffler:
> Nice work! Thank you!

Thanks!
 
> Did you run/test the examples in a fresh checkout?


Yes, I worked on all of these as separate projects and then simply copied the files back into the JavaSMT folder. So they are tested and should be working. However, it might still be better to wait until we have the CI set up for these example projects before merging anything

doc/Example-Gradle-Project-Kotlin/build.gradle.kts
baierd: Thank you!

baierd:
@daniel-raffler are there any changes to the behavior of the projects or their tests due to the changes in this PR?

daniel-raffler:
> @daniel-raffler are there any changes to the behavior of the projects or their tests due to the changes in this PR?

No, nothing major. The Gradle and Kotlin were quite outdated and were missing tests for several of the new solvers. I tried to make them more aligned with what we're doing in the Maven project, which was much more up to date, and they're now also using parametrized tests for all the solvers. As a side-effect this means that the test is now always for `Sudoku.BooleanBasedSudokuSolver` when it used to be `Sudoku.IntegerBasedSudokuSolver` before for solvers that support integers

PR #666, opened 2026-05-24
Make (SMTLIB) Tokenizer iterable via a new iterator to make SMTLIB operators more efficient

baierd:
This PR adds a `Iterator` implementation for tokenizing SMTLIB Strings based on the existing `Tokenizer` code. `Tokenizer` itself now implements `Iterable`, and gives access to the `Iterator` on demand.

This allows processing of SMTLIB in for-each loops efficiently and without creating multiple `Collection`s in the process.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
@@ -35,6 +35,16 @@ public final class SMTLibTokenizer implements Iterable<String> {
     input = checkNotNull(inputToTokenize);
   }
 
+  /**
+   * Creates a new {@link SMTLibTokenizer} from the input {@link String}. The resulting {@link
+   * SMTLibTokenizer} is an {@link Iterable} for the SMTLIB tokens. A new {@link Iterator} on the
+   * input can be created with {@link #iterator()}. A {@link ImmutableList} of all tokens can be
+   * created via {@link #toImmutableList()}. Tokenized input is returned sanitized from unnecessary
+   * characters, and checked for basic (but not full) syntactic validity (e.g. that every opened
+   * bracket is also closed in a valid way).
+   *
+   * @param inputToTokenize SMTLIB input {@link String} that may be empty, but not null.
+   */
   public static SMTLibTokenizer of(final String inputToTokenize) {
     return new SMTLibTokenizer(inputToTokenize);
   }
@@ -50,7 +60,8 @@ public final class SMTLibTokenizer implements Iterable<String> {
 
   /**
    * Split up a sequence of lisp expressions into a list of nen-empty tokens. This method simply
-   * uses the {@link TokenizerIterator} and creates a list out of all tokens.
+   * uses the {@link TokenizerIterator} and creates a list out of all tokens. Repeated calls to this
+   * method are inefficient, as a new {@link java.util.List} is created every time it is called.
    *
    * <p>This is used by {@link AbstractFormulaManager#parse(String)} as part of the preprocessing
    * before the input is passed on to the solver. SMT-LIB2 scripts are sequences of commands that
--- a/src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
@@ -308,20 +308,6 @@ public final class SMTLibTokenizer implements Iterable<String> {
      * Optional#empty()} if there is none.
      */
     private Optional<String> getNextToken() {
-      while (!endOfInputReached()) {
-        String maybeNextToken = getNextPossiblyEmptyToken();
-        if (!maybeNextToken.isEmpty()) {
-          return Optional.of(maybeNextToken);
-        }
-      }
-      return Optional.empty();
-    }
-
-    /**
-     * Progresses the iterator and returns the next token. These tokens might be empty due to our
-     * lenient way of interpreting SMTLIB.
-     */
-    private String getNextPossiblyEmptyToken() {
       StringBuilder tokenBuilder = new StringBuilder();
       String token = null;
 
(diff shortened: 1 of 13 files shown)


src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
```diff
@@ -0,0 +1,437 @@
...
+     * on to the solver.
+     *
+     * <p>As an example <code>tokenize("(define-const a Int)(assert (= a 0)")</code> will return the
+     * following tokens: <code>["(define-const a Int)", "(assert (= a 0))"]</code>
+     *
+     * @param inputToTokenize SMTLIB2 {@link String}.
+     */
+    private TokenizerIterator(final String inputToTokenize) {
+      input = inputToTokenize;
+      inputLength = input.length();
```
daniel-raffler: I think `String.length()` is already `O(1)` in Java ;)

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
```diff
@@ -0,0 +1,437 @@
...
+
+      // Generate first token if possible. There might be none, in which case isNext() needs to
+      // return 'false' immediately.
+      while (nextToken.isEmpty() && !endOfInputReached()) {
+        String maybeFirstToken = getNextPossiblyEmptyToken();
+        // Can be optimized by returning null for empty tokens in getNextPossiblyEmptyToken()
+        if (!maybeFirstToken.isEmpty()) {
+          nextToken = Optional.of(maybeFirstToken);
+        }
+      }
```
daniel-raffler: It should be fine to just do something like `nextToken = getNextPossiblyEmptyToken()`

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
```diff
@@ -0,0 +1,437 @@
...
+
+      // Get lookahead if possible
+      while (!endOfInputReached()) {
+        String maybeNextToken = getNextPossiblyEmptyToken();
+        // Can be optimized by returning null for empty tokens in getNextPossiblyEmptyToken()
+        if (!maybeNextToken.isEmpty()) {
+          nextToken = Optional.of(maybeNextToken);
+          break;
+        }
+      }
```
daniel-raffler: Same here

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
```diff
@@ -0,0 +1,437 @@
...
+      // nextToken may be present or empty!
+      return checkNotNull(currentToken);
```
daniel-raffler: I'm not sure this can actually be `null` here

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
```diff
@@ -0,0 +1,437 @@
...
+      if (level != 0) {
+        // Throw an exception if the brackets don't match
+        throw new IllegalArgumentException("parentheses do not match, too many open parentheses");
+      }
+
+      if (token == null) {
+        // This is always the empty token. Can only happen at tWe need to fix this first.he end of
+        // the input.
+        token = tokenBuilder.toString();
+      }
```
daniel-raffler, line 421: IIRC this can only happen at the very end of the input if the last token hasn't been "closed". One example would be `(assert (= a 0)) "(` where we forgot to close the `"` at the end. In that case the partial token still needs to be added to the output so that the solver can trip over the syntax error

I think you could probably just return an `Optional` here:
```java
if (token == null) {
  // Can only happen at the end of the input.
  var str = tokenBuilder.toString();
  return str.isEmpty() ? Optional.empty() : Optional.of(str);
}
return Optional.of(token);
```

This should then make the code for getting the next `lookahead` token a bit simpler (see the other comments)

daniel-raffler (dismissed):
LGTM!

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
baierd: I added the check just to be safe. Especially since i thought about returning `null` instead of empty tokens at some point (which might still happen at some point in the future).

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
baierd: Can you guarantee that the `StringBuilder` is never empty in all cases in which it builds a String?

Because i was not sure whether there is some edge-case in this implementation in which this may happen!

baierd: I removed the while-loops and made this method end-recursive instead (i.e. it does not longer return empty tokens, but `Optional`s as proposed by you)

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
baierd: Good point ;D

src/org/sosy_lab/java_smt/basicimpl/SMTLibTokenizer.java
daniel-raffler: > Can you guarantee that the StringBuilder is never empty in all cases in which it builds a String?

Well, at least it wasn't supposed to 😅. But I agree that the code is a bit messy, so it might be better to keep the option just to be on the safe side

> I removed the while-loops and made this method end-recursive instead (i.e. it does not longer return empty tokens, but Optionals as proposed by you)

That sounds like a good solution 👍

daniel-raffler (approved):

baierd:
Note: i would like to test this at least a little (so that no unexpected regressions are introduced). So we will have to wait until our test-cluster is online again.

PR #667, opened 2026-05-30
Add CI for example projects

### LINKED ISSUE #663: Add CI jobs for Gradle and Maven
Since Gradle and Maven is used by a non-trivial number of our users, it would be good to add a CI job for each, checking that all solvers (including Yices2) are available and work.

daniel-raffler:
Hello everyone,

this PR adds CI checks for all project examples in the `doc` directory to verify that they're working, and make sure that there are no issues with the packages that we've published to Maven. It is based on two earlier PRs that update the example projects and add another project template for using JavaSMT in Ivy. As a result, this PR now contains CI for using JavaSMT with 5 different build systems:
- Maven
- Maven Web
- Gradle
- Kotlin/Gradle
- Ivy

Except for `Maven Web`, all CI checks are run multiple times to cover all supported platforms. For `Maven Web` the project is a web application and we only test `x64-linux` as this is most likely what the server will use. Also note, that for the web application we don't actually test if the `*.war` is working as this would be too difficult to do inside of a Github workflow

Results from the latest CI run can be found [here](https://github.com/daniel-raffler/java-smt/actions/runs/26681661028) for my fork of the project. To make it work here we would still have to whilelist an `action` that is used by the workflow, see [here](https://github.com/sosy-lab/java-smt/actions/runs/26681433235)

Closes #663


doc/Example-Gradle-Project-Kotlin/src/test/kotlin/org/sosy_lab/java_smt_example/AppTest.kt
```diff
@@ -2,48 +2,220 @@
...
+                        || (IS_WINDOWS && !IS_ARCH_ARM64)
+
+                Solvers.MATHSAT5 -> (IS_WINDOWS && !IS_ARCH_ARM64)
+                Solvers.Z3 -> (IS_LINUX && isSufficientVersionOfLibcxx("z3"))
+                        || IS_WINDOWS
+                        || IS_MAC
+
+                Solvers.Z3_WITH_INTERPOLATION -> IS_LINUX && !IS_ARCH_ARM64
+            }
+        }
```
daniel-raffler: This code is taken from `SolverContextFactory` and clutters up the example. Maybe there's a way to make `isSupportedOperatingSystemAndArchitecture` part of the JavaSMT API to avoid all the duplicate code?

.github/workflows/example-project-ci.yml
```diff
@@ -0,0 +1,298 @@
...
+
+      - name: Upload test report
+        uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          path: doc/Example-Gradle-Project/build/reports/tests/test/index.html
+          archive: false
+
+  kotlin:
+    name: Kotlin Example Project
```
baierd: Adding the used build-system would be good.

.github/workflows/example-project-ci.yml
```diff
@@ -0,0 +1,298 @@
...
+  push
+
+jobs:
+  ivy:
+    name: Ivy Example Project
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [ 21 ]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2025, windows-11-arm, macos-26-intel, macos-26]
```
baierd: Selecting the OS this way would require us to manually update this file every time we have a newer OS version. This might lead to this being overseen. You can just use the latest tag with wildcards, as can be seen [here](https://github.com/sosy-lab/java-smt/blob/70644689e62182cadaa025498b4bb3a5620b536f/.github/workflows/ci.yml#L62) for example, so that we automatically use the newest OS.

.github/workflows/example-project-ci.yml
```diff
@@ -0,0 +1,298 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK ${{ matrix.java-version }}
```
baierd, line 26: Why not use a `dependencies` job as in `ci.yml` to set up our environment correctly, i.e. so that Java and other dependencies are present. And why don't we just extend `ci.yml`?

doc/Example-Gradle-Project/src/test/java/org/sosy_lab/java_smt_example/SudokuTest.java
```diff
@@ -109,77 +140,66 @@ public void init() throws InvalidConfigurationException {
...
+      case Z3 -> (IS_LINUX && isSufficientVersionOfLibcxx("z3")) || IS_WINDOWS || IS_MAC;
+      case Z3_WITH_INTERPOLATION -> IS_LINUX && !IS_ARCH_ARM64;
+    };
   }
 
-  private void checkSudoku(Solvers solver)
+  @Test
+  public void checkSudoku()
       throws InvalidConfigurationException, InterruptedException, SolverException {
+    assumeTrue(isSupportedOperatingSystemAndArchitecture(solver));
```
baierd: Wouldn't a `@BeforeEach` that checks this make more sense?

Note: and please name and document the expected behavior such that it becomes clear what went wrong fast.

The same thing applies to all test classes in this PR.

doc/Example-Ivy-Project/build.xml
```diff
@@ -0,0 +1,140 @@
+<!--
+This file is part of JavaSMT,
+an API wrapper for a collection of SMT solvers:
+https://github.com/sosy-lab/java-smt
+
+SPDX-FileCopyrightText: 2025 Dirk Beyer <https://www.sosy-lab.org>
```
baierd, line 6: 2026 😉

doc/Example-Ivy-Project/build.xml
```diff
@@ -0,0 +1,140 @@
...
+    <property name="project.organization" value="org.sosy_lab.java_smt_example"/>
+    <property name="project.module" value="${ant.project.name}"/>
+    <property name="project.jar" value="${project.module}.jar"/>
+
+    <!-- Path to the "main" function -->
+    <property name="project.main" value="org.sosy_lab.java_smt_example.SolverOverviewTable"/>
+
+    <target name="download-ivy" unless="skip.download">
+        <mkdir dir="lib"/>
+        <get src="https://repo1.maven.org/maven2/org/apache/ivy/ivy/2.5.0/ivy-2.5.0.jar"
```
baierd: We install Ivy from Maven? ;D

I think it would be better to fail instead of installing something here. You can take a look at how CPAchecker fails if Ant or Ivy are missing [here](https://gitlab.com/sosy-lab/software/cpachecker/-/blob/main/build.xml?ref_type=heads#L15).

doc/Example-Ivy-Project/build.xml
```diff
@@ -0,0 +1,140 @@
...
+        <taskdef resource="org/apache/ivy/ant/antlib.xml" uri="antlib:org.apache.ivy.ant" classpathref="ivy.path"/>
+        <ivy:settings file="lib/ivy-settings.xml"/>
+    </target>
+
+    <target name="resolve" depends="init-ivy" description="--> retrieve dependencies with Ivy">
+        <ivy:resolve log="download-only"/>
+        <ivy:retrieve pattern="lib/java/[conf]/([arch]/)[artifact](-[classifier]).[ext]"/>
+    </target>
+
+    <target name="copy" depends="resolve" description="--> copy solver libraries">
```
baierd, line 43: Since we usually don't want users to do this, we should document why we do it here!

doc/Example-Ivy-Project/build.xml
baierd: Note: applies to several files.

baierd (changes requested):
Nice work, thank you!

Something to discuss (input from @kfriedberger and @PhilippWendler would be appreciated):

I noticed that we copy around files in the example projects. This might be confusing to people that use them as templates. A possible solution for this would be to extract the example projects into their own repositories and link the repositories to ours (so that they are folders that just link to the other repo).  
Edit: [this](https://github.com/sosy-lab/java-smt/pull/662#issuecomment-4554203360) also seems to indicate that this would be a good idea, as it would make developing/updating in the projects easier.

.github/workflows/example-project-ci.yml
daniel-raffler: I've renamed it to "Kotlin Gradle Project" now. Is that fine with you?

.github/workflows/example-project-ci.yml
daniel-raffler: I've now updated them to `latest` as far as possible. However, for the arm runners there is no `latest` release and the exact version needs to be provided (see [here](https://github.com/orgs/community/discussions/148648#discussioncomment-11858187))

.github/workflows/example-project-ci.yml
daniel-raffler: Java is fetched by `actions/setup-java` and we're using caches to avoid downloading the dependencies multiple times. We could try integrating it further with `ci.yml`, but since the examples are using various different build systems, there may not be that much overlap

In total the `example-project-ci` workflow takes 7-8 minutes to run, which is probably still fine, especially as we may not have to run it as often as our regular CI jobs

doc/Example-Ivy-Project/build.xml
daniel-raffler: Ivy does need to be downloaded before building as it is not included in the `ant` installation. However, I did change the download location to `sosy-labs` now

.github/workflows/example-project-ci.yml
```diff
@@ -0,0 +1,298 @@
...
+# https://github.com/sosy-lab/java-smt
+#
+# SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: example-project-ci
+
+on:
+  push
```
daniel-raffler, line 12: We should remember to change this before merging. Right now it will trigger whenever anything is pushed to github. We could restrict it to only pushes to `main`, or maybe even have it manually triggered (whenever we release a new version of JavaSMT)

PhilippWendler:
> I noticed that we copy around files in the example projects. This might be confusing to people that use them as templates. A possible solution for this would be to extract the example projects into their own repositories and link the repositories to ours (so that they are folders that just link to the other repo).

I do not know what files you are copying around nor when, why, or how a separate repository is supposed to help with that. If you want input from me please explain the current solution and its problems.

baierd:
> > I noticed that we copy around files in the example projects. This might be confusing to people that use them as templates. A possible solution for this would be to extract the example projects into their own repositories and link the repositories to ours (so that they are folders that just link to the other repo).
> 
> I do not know what files you are copying around nor when, why, or how a separate repository is supposed to help with that. If you want input from me please explain the current solution and its problems.

JavaSMT requires solver binaries to be in a particular location (if you don't use custom loading mechanisms). Since the example projects are located in some nested folders, we need to move the solver binaries around after we retrieved them, so that the example projects work properly. You can see this for example [here](https://github.com/sosy-lab/java-smt/pull/667/changes#diff-01cde70dd855b6d6c55cbc18fe9146da0ca987988932f101c4817ab0cc4dc539L142-L148). This might be confusing for users that just copy the example projects.
If we were to use dedicated repositories per example project, this problem would not exist.

PhilippWendler:
> If we were to use dedicated repositories per example project, this problem would not exist.

Why? How would the solver binaries end up in the example project then? And why couldn't the same mechanism be used for example projects in a subdirectory?

doc/Example-Gradle-Project-Kotlin/src/test/kotlin/org/sosy_lab/java_smt_example/AppTest.kt
daniel-raffler: See issue https://github.com/sosy-lab/java-smt/issues/692

PR #670, opened 2026-06-08
Add MathSAT5 logic selection via option

baierd:
This PR adds context wide (i.e. all provers created from a certain context) logic selection via a new option for MathSAT5 with tests. This adds a new function to the wrapper and MathSAT5 needs to be recompiled and re-published for this PR.


baierd:
MathSAT5 needs to be updated in our Ivy/Maven repos before this can be merged. But everything is tested locally and seems to work fine.

PR #671, opened 2026-06-09
Add common optimization prover delegate

baierd:
This PR extracts common checks of `OptimizationProverEnvironment`s into a single new delegate `OptimizationProverDelegate`.

Changes:
- fixes a bug in Z3. In certain cases it was possible to request a model even if the preconditions were not fulfilled.

PR #673, opened 2026-06-09
Handle model generation api through implementation delegates

baierd:
Model generating API (e.g. getModel()) was implemented for every solver. Meaning that also all checks that are required before executing solver specific code were added on a per solver basis. Since this is error prone, this PR collects these common checks now in `AbstractProver` and delegates the solver specific implementations to new, internal API that does no longer need to think about these common checks.

Note: i did not remove the default implementations in `BasicProverEnvironment`, as this would potentially break client code that inherits from it.

To be merged after #671 as it is based on its branch.

### DIFF
--- a/src/org/sosy_lab/java_smt/basicimpl/OptimizationProverDelegate.java
+++ b/src/org/sosy_lab/java_smt/basicimpl/OptimizationProverDelegate.java
@@ -1,0 +1,141 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.sosy_lab.java_smt.basicimpl;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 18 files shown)


src/org/sosy_lab/java_smt/basicimpl/OptimizationProverDelegate.java
```diff
@@ -0,0 +1,139 @@
...
+ * This delegate enables common implementations for methods in {@link OptimizationProverEnvironment}
+ * based on the implementations in the abstract/theorem prover that can not be done using abstract
+ * implementations.
+ */
+public class OptimizationProverDelegate implements OptimizationProverEnvironment {
+
+  private final OptimizationProverEnvironment optimizationProver;
+
+  OptimizationProverDelegate(OptimizationProverEnvironment pBaseProver) {
+    optimizationProver = checkNotNull(pBaseProver);
```
daniel-raffler, line 38: Maybe we could add a check here to make sure that the argument is a subclass of `AbstractProver`:
```
checkArgument(AbstractProver.class.isAssignableFrom(pBaseProver.getClass()));
```
It will throw later either way, but I think it might be helpful to get the exception as early as possible

daniel-raffler (dismissed):
LGTM

src/org/sosy_lab/java_smt/basicimpl/OptimizationProverDelegate.java
baierd: Failing early is always better. Good thought. I added it.

Issue #672, opened 2026-06-09
Methods with default implementations missing from delegates


baierd:
Several methods defined in `BasicProverEnvironment` with default implementations are missing in the delegates (Tracing/Debugging etc.). Since most of these default implementations are overridden for some specialized solvers/behavior, we miss these interactions in our delegates.

PR #674, opened 2026-06-09
Add common precondition checks for unsat core and allSat

baierd:
UnsatCore and AllSat implementations performed common checks per solver. This PR adds a common implementation that handles these checks before delegating to solver specific or default implementations.

Should only be merged after #673 as it is based on its branch.


src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -297,6 +332,63 @@ protected void closeAllEvaluators() {
...
+
+  @Override
+  public final Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
+      Collection<BooleanFormula> assumptions) throws SolverException, InterruptedException {
+    checkGenerateUnsatCoresOverAssumptions(assumptions);
+    return unsatCoreOverAssumptionsImpl(assumptions);
+  }
+
+  /**
+   * TODO: once javac does not complain if we use a default impl here, add one.
```
daniel-raffler: Adding `@SuppressWarnings("unused")` worked for me:
```java
@SuppressWarnings("unused")
protected Optional<List<BooleanFormula>> unsatCoreOverAssumptionsImpl(
    Collection<BooleanFormula> assumptions) throws SolverException, InterruptedException {
  throw new UnsupportedOperationException(UNSAT_CORE_WITH_ASSUMPTIONS_NOT_SUPPORTED);
}
```

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
```diff
@@ -161,8 +160,7 @@ public boolean isUnsatWithAssumptions(Collection<BooleanFormula> assumptions) {
    */
   @SuppressWarnings("resource")
   @Override
-  public Model getModel() throws SolverException {
-    checkGenerateModels();
+  protected Model getModelImpl() throws SolverException {
```
daniel-raffler, line 167: This method could be `protected` now. The same goes for this method and `getStatisticsImpl` in several of the other solver implementations

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
```diff
@@ -80,26 +82,41 @@ protected final void checkGenerateModels() {
...
+  protected final void checkGenerateUnsatCoresOverAssumptions(
+      Collection<BooleanFormula> assumptions) {
+    // TODO: should this close all evaluators as well?
     Preconditions.checkState(!closed);
+    Preconditions.checkNotNull(assumptions);
     Preconditions.checkState(
         generateUnsatCoresOverAssumptions,
         TEMPLATE,
         ProverOptions.GENERATE_UNSAT_CORE_OVER_ASSUMPTIONS);
+    // TODO: why is there no check for changedSinceLastSatQuery or wasLastSatCheckSatisfiable?
```
daniel-raffler: When `unsatCoreOverAssumptions` is called, JavaSMT will always run a SAT check first, so there's no need to check the solver status. It's mentioned in the documentation for `BasicProverEnvironment.unsatCoreOverAssumptions`, but maybe we could add some more documentation here?

daniel-raffler (dismissed):
LGTM

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: I am thinking about not having default implementations at all to be honest. Mainly due to https://github.com/sosy-lab/java-smt/issues/672. I would just leave it as it is for now.

src/org/sosy_lab/java_smt/solvers/bitwuzla/BitwuzlaAbstractProver.java
baierd: Thanks! I missed quite a few actually :sweat_smile:. They should be fixed now.

src/org/sosy_lab/java_smt/basicimpl/AbstractProver.java
baierd: Ah! Thanks for the hint. So three things:
- I forgot `isUnsatWithAssumptions` for the common checks impl. delegates!
- We should document the requirements for `unsatCoreOverAssumptions`. Bitwuzla for example does not use `isUnsatWithAssumptions`, but SAT checks on its own.
- Why doesn't Bitwuzla have an impl. for `isUnsatWithAssumptions` but for `unsatCoreOverAssumptions`?

baierd:
There is a potential problem with the approach used until now (adding a default impl in `AbstractProver` with a new abstract method that implements the solver behavior).
The assumptions wrapper is a delegate that allows us to use assumption solving in solvers that do not support it (or not like we expect it). But this wrapper directly implements the prover interfaces (`BasicProverEnvironment` etc.) and are layered above the `AbstractProver`. Thus we actually can't check these methods before at least some operations happen (for example adding the assumptions to our tracking layer). Fortunately, there are no solver interactions that are not checked at some point (e.g. in `AbstractProver`).

We can simply accept this, as it is not a big deal.

The alternative is that we add a "checking" delegate that is always guaranteed to be used first. But such a solution would require us to provide a generalized interface for delegates to access the `AbstractProver` through multiple stacked delegates. Such a solution would add more complexity, and i don't think that it is worth it.

PR #677, opened 2026-06-11
Remove experimental FP-to-BV impl from Bitwuzla

baierd:
Remove experimental FP-to-BV impl from Bitwuzla as it adds new formulas to the solver stack silently and in unexpected circumstances. This PR pulls these changes from #512, to reduce the amount of changes in #512.


daniel-raffler (approved):
I think you did get all the parts that were added to handle `FP-to-BV` in Bitwuzla

LGTM!

PR #683, opened 2026-06-29
Add support for eldarica horn solver

Bixilon:
- [x] Conversion from Princess to "Eldarica"
 - [x] Tests, sample smt files from [chc-comp-2026](https://github.com/chc-comp/chc-comp-2026) 
 - [x] Run [chc-comp-2026](https://github.com/chc-comp/chc-comp-2026) competition on benchcloud and compare results (looking good so far,  about 380 not running, some are broken; Evaluation looking very similar to CHC-COMP-2026)
 - [x] Get model/counter example
 - [ ] (Get eldarica from ivy @baierd)
 - [x] Get, integrate and fixup  `provide_javasmt_main_to_execute_as_solver_for_smt-comp` @gcarpio21 (it works and is not bad, but we can also revert it)
 - [x] Bump princess  (`2025-11-17` to `2026-05-20`); required for eldarica; there are still issues #697, ...
 - [ ] Adding support for intervals in princess requires (safe) reflection, not sure how the policy about that is 
 - [x] convert eldarica horn tests to parameterized tests
 - [x] push some of the remaining changes (from self code review, minimal cleanup of old things)
 - [x] support ite terms
 - [ ] support (ex) quantifiers inside formula
 - [x] support epsilon terms properly (requires #684)

The evaluation (chc-comp 2026) is looking good, (7820 vs. 7892 correct results, but 3 vs 0 incorrect results)


src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
```diff
@@ -0,0 +1,73 @@
...
+import org.sosy_lab.java_smt.basicimpl.FormulaCreator;
+
+public class EldaricaModel extends AbstractModel<IExpression, Sort, PrincessEnvironment> {
+  private final Map<Predicate, IFormula> model;
+
+  EldaricaModel(
+      Map<Predicate, IFormula> model, EldaricaHornProver prover,
+      FormulaCreator<IExpression, Sort, PrincessEnvironment, ?> creator) {
+    super(prover, creator);
+    this.model = model;
```
baierd: @Bixilon you don't need `this` if you use non-clashing names. It is even preferred to use distinct names for variables and parameters, so that they don't hide each other.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
```diff
@@ -0,0 +1,73 @@
...
+    for (var entry : this.model.entrySet()) {
+      assignments.addAll(getAssignments(entry.getKey(), entry.getValue()));
+    }
+
+    return assignments.build().asList();
+  }
+
+  @Override
+  protected @Nullable IExpression evalImpl(IExpression formula) {
+    throw new UnsupportedOperationException(); // TODO?
```
baierd: @Bixilon 
Hint: implementing this helps you finishing `getAssignments()` above more easily.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
```diff
@@ -124,8 +127,58 @@ protected boolean isUnsatImpl() throws SolverException {
...
+  }
+
+  @Override
+  public Model getModelImpl() throws SolverException {
+    var result = solve();
+    if (result.isRight()) {
+      throw new IllegalStateException("Unsat, no model available!");
+    }
+
+    var solution = result.left().get().apply();
```
baierd: @Bixilon please only use `var` if the signature is obvious (or in tests). Here it is not obvious at all, and i would advice to change it to the actual type. (This is just an example. Try to adhere to the rule stated in the beginning.)

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
```diff
@@ -124,8 +127,58 @@ protected boolean isUnsatImpl() throws SolverException {
...
+  @Override
+  protected void setChanged() {
+    result = null;
+  }
+
+  @Override
+  public Model getModelImpl() throws SolverException {
+    var result = solve();
+    if (result.isRight()) {
+      throw new IllegalStateException("Unsat, no model available!");
```
baierd: @Bixilon good thought. But you don't need to check this. The `AbstractProver` checks whether the prover state allows certain operations; here whether the last result was SAT before using the `getModelImpl()` method.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
```diff
@@ -56,48 +64,42 @@ public class EldaricaHornProver extends PrincessAbstractProver<Void> implements
...
-  }
 
-  @Override
-  @Nullable
-  protected Void addConstraintImpl(BooleanFormula constraint) throws InterruptedException {
-    addConstraint1(constraint);
     return null;
   }
 
   private Either<Function0<Map<Predicate, IFormula>>, Function0<Dag<IAtom>>> solve() {
```
baierd, line 103: @Bixilon this method could benefit from some documentation. It is not clear what is happening from reading the code. It seems like this method is used to check queries for satisfiability, but also to resolve models from naively reading the code and its usage. But both should not be true at the same time.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
Bixilon: They are not shadowing each other (because of the `this`). It's my preferred way of writing it (then it is clear that I am accessing a member variable). Should I change it everywhere?

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
Bixilon: The type is (in that case) very long `Either<Function0<Map<Predicate, IFormula>>, Function0<Dag<IAtom>>>` and hard to read, but yes, changed.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
Bixilon: @baierd This is how Eldarica works, its more like a wrapper for setup eldarica + run the solver. The type also comes from Eldarica. And its private.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
Bixilon: For other operations (`pop`, ...) I am explicitly overriding it, as they are supported (and implemented) in princess, but not in eldarica.

baierd: I see.
If you merge current `main`, this won't be an issue. (We recognized this as a problem a while ago and `main` now has a solution for this that is independent of any solver implementation)

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
baierd: Of course you should just stick to what a solver does. But this boosts my claim that this needs documentation ;D

src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
baierd: The modern way of writing Java avoids `this` as it adds visual clutter. And not using the same name (even with `this` etc.) is generally better maintainable, as changes have less of a chance of accidentally adding problems.

We care about is maintainability and easy to read code, especially since many students work on our projects. 
This does not break either of the concepts (or not in a way that hurts), my comment was just a suggestion.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
baierd: The point of that `var` should only be used if the type is obvious. Inferring the type from the existing code/information is impossible for such a complex type.

Thank you!

src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
Bixilon: I completely disagree. Having the same name in the constructor and as member variables makes it very clear that it is the same variable, like:
```java
var record = new Something(/*model=*/ "test");
record.model // -> "test"
``` 

But well, you are the maintainer, I will change it :hand_over_mouth:

src/org/sosy_lab/java_smt/solvers/princess/EldaricaModel.java
Bixilon: I implemented this without the evalImpl, I am not really sure how to implement eval properly. Is this really needed? I am getting the value assignments.

src/org/sosy_lab/java_smt/solvers/princess/EldaricaHornProver.java
Bixilon: I already have current `main` merged into my branch. Should I leave it as it is?

PR #685, opened 2026-07-07
Override `default` methods in `BasicProverEnvironment` for all delegates

### LINKED ISSUE #672: Methods with default implementations missing from delegates

daniel-raffler:
Hello,

this PR adds overrides for all `default` methods in `BasicProverEnvironment` for all of our delegates. This makes sure that the solver implementation is used when available, and that we don't fall back to the default implementation when using a delegate


src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
```diff
@@ -35,6 +39,13 @@ class StatisticsBasicProverEnvironment<T> implements BasicProverEnvironment<T> {
     stats.provers.getAndIncrement();
   }
 
+  @Override
+  public @Nullable T push(BooleanFormula f) throws InterruptedException {
+    stats.push.getAndIncrement();
+    stats.pop.getAndIncrement();
```
baierd: I think this line should not be here.

src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
```diff
@@ -99,6 +123,11 @@ public Optional<List<BooleanFormula>> unsatCoreOverAssumptions(
     return delegate.unsatCoreOverAssumptions(pAssumptions);
   }
 
+  @Override
+  public ImmutableMap<String, String> getStatistics() {
+    return ImmutableMap.of();
```
baierd: Missing delegate call?

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluator.java
```diff
@@ -0,0 +1,108 @@
...
+import org.sosy_lab.java_smt.api.FloatingPointNumber;
+import org.sosy_lab.java_smt.api.Formula;
+import org.sosy_lab.java_smt.api.NumeralFormula;
+import org.sosy_lab.java_smt.api.SolverContext;
+import org.sosy_lab.java_smt.api.StringFormula;
+
+class SynchronizedEvaluator implements Evaluator {
+
+  private final Evaluator delegate;
+  final SolverContext sync;
```
baierd: Do we want this to be package-private?

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluatorWithContext.java
```diff
@@ -0,0 +1,105 @@
...
+import org.sosy_lab.java_smt.api.StringFormula;
+
+class SynchronizedEvaluatorWithContext implements Evaluator {
+
+  static final String UNSUPPORTED_OPERATION = "translating non-boolean formulae is not supported";
+
+  private final Evaluator delegate;
+  final SolverContext sync;
+  final FormulaManager manager;
+  final FormulaManager otherManager;
```
baierd: Do we want all of these to be package-private?

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluatorWithContext.java
```diff
@@ -0,0 +1,105 @@
...
+      FormulaManager pOtherManager) {
+    delegate = checkNotNull(pDelegate);
+    sync = checkNotNull(pSync);
+    manager = checkNotNull(pManager);
+    otherManager = checkNotNull(pOtherManager);
+  }
+
+  @Override
+  public <T extends Formula> @Nullable T eval(T pFormula) {
+    throw new UnsupportedOperationException(UNSUPPORTED_OPERATION);
```
baierd, line 63: Why can't we synchronize evaluators with contexts?

I know that the old code also did not allow this, but why?

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedModelWithContext.java
```diff
@@ -11,87 +11,19 @@
...
-import org.sosy_lab.java_smt.api.StringFormula;
 
-class SynchronizedModelWithContext implements Model {
-
-  private static final String UNSUPPORTED_OPERATION =
-      "translating non-boolean formulae is not supported";
+class SynchronizedModelWithContext extends SynchronizedEvaluatorWithContext implements Model {
 
+  @SuppressWarnings("unused")
   private final Model delegate;
```
baierd: If its unused, why not get rid of it?

src/org/sosy_lab/java_smt/delegate/trace/TraceBasicProverEnvironment.java
```diff
@@ -80,6 +93,39 @@ public Model getModel() throws SolverException {
...
+        () -> new TraceEvaluator(delegate.getEvaluator(), mgr, logger));
+  }
+
+  @Override
+  public ImmutableList<Model.ValueAssignment> getModelAssignments() throws SolverException {
+    ImmutableList<Model.ValueAssignment> result =
+        logger.logDefDiscard(
+            logger.toVariable(this), "getModelAssignments()", delegate::getModelAssignments);
+    return FluentIterable.from(result)
+        // TODO Fix this in the Z3 model
```
baierd: Would you be so kind and open a issue for this?

baierd (changes requested):
Looks good besides the open points! Nice work!

Please open multiple smaller PRs next time instead of one giant one ;D

Did you test whether we handle `push(BooleanFormula)` correctly in all cases? (I saw that you added it to the trace test. Good!) The used delegate calls `addConstraint(BooleanFormula)` due to the default implementation, but is that `addConstraint()` call then tracked by the delegate layers?
Do we even need to add `push(BooleanFormula)` to the delegates? The default impl. uses two methods that are already part of the layers, and it is never overridden.

We should test most of our additional layers (logging/debugging etc.) from time to time! Would you be so kind and open issues (per delegate) to do this? Running CPAchecker with distinct analyses for each should be fine for now to get a sense of whether they work or collect problems.

src/org/sosy_lab/java_smt/delegate/trace/TraceBasicProverEnvironment.java
daniel-raffler: I think the issue has already been fixed by 48b66d485cee5cae0b2593cbdfe413588a23257c and 3e665e58fa7980f04bf6b12b61777f85d9a859b2

src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
daniel-raffler: Nice catch! I've replaced it with
```java
@Override
  public @Nullable T push(BooleanFormula f) throws InterruptedException {
    stats.push.getAndIncrement();
    stats.constraint.getAndIncrement();
    return delegate.push(f);
  }
```
which is what the default method would have produced

Should I get rid of the `stats.constraint.getAndIncrement()` entirely?

src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
daniel-raffler: Hmm.. I wasn't quite sure what to put there. In `StatisticsSolverContext` the `getStatistics` method includes the data that we're collecting. However, this won't work for `StatisticsBasicProverEnvironment` as we're not tracking statistics for each prover individually

I've now simply added a call to the delegate

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluator.java
daniel-raffler: They should be private now

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedModelWithContext.java
daniel-raffler: It's removed now

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluatorWithContext.java
daniel-raffler: I think it's just because `FormulaManager.translate` expects a boolean formula. We could allow `eval` for boolean formulas, but throwing an `UnsupportedOperationException` only for some inputs may be a bit confusing to the user

daniel-raffler:
> Looks good besides the open points! Nice work!

Thanks!

> Please open multiple smaller PRs next time instead of one giant one ;D

Sure, I'll try to keep the PRs smaller 😅. Most of the large changes are just copy and paste, but I can see how it's hard to review

> Did you test whether we handle `push(BooleanFormula)` correctly in all cases? (I saw that you added it to the trace test. Good!) The used delegate calls `addConstraint(BooleanFormula)` due to the default implementation, but is that `addConstraint()` call then tracked by the delegate layers?

The tracing delegate will just write `prover.push(f)` to the log. The formula for `f` has already been unrolled and is just a variable name. When translating to SMTLIB the call then needs to be split up and we generate `(push 1) (assert f)`

There is an example of how the trace changed in one of the tests. Before adding the override we had:
```
var var1 = mgr.getBooleanFormulaManager().makeTrue();
var0.push();
var var2 = var0.addConstraint(var1);
```
Now it's just:
```
var var1 = mgr.getBooleanFormulaManager().makeTrue();
var var2 = var0.push(var1);
```
> We should test most of our additional layers (logging/debugging etc.) from time to time! Would you be so kind and open issues (per delegate) to do this? Running CPAchecker with distinct analyses for each should be fine for now to get a sense of whether they work or collect problems.

I've open some issues [here](https://github.com/sosy-lab/java-smt/issues/686)

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedModelWithContext.java
baierd: Thanks!

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluatorWithContext.java
baierd: I think the current message is confusing, as it actually says that only boolean formulas are allowed.

Would you be so kind and test boolean formulas only + add a small test for this? Essentially just try to eval a boolean and a non-boolean formula.

src/org/sosy_lab/java_smt/delegate/trace/TraceBasicProverEnvironment.java
baierd: Good point. 
If that's the case, we should remove the TODO. Could you double check with the commit that added the TODO whether we can be sure that it is related to the 2 fixes and remove the TODO if they are?

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluator.java
baierd: Thank you!

src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
baierd: We can always just return the solvers statistics if it provides some.

But i agree that this is actually a interesting question. I'll take a closer look.

src/org/sosy_lab/java_smt/delegate/synchronize/SynchronizedEvaluatorWithContext.java
daniel-raffler: I've implemented it for booleans now, see [here](https://github.com/sosy-lab/java-smt/commit/71ca20cc5020552febdb623cb2bba062812a7a42)

src/org/sosy_lab/java_smt/delegate/statistics/StatisticsBasicProverEnvironment.java
baierd: No i think the version proposed by you here (i.e. the version that default would have produced) is accurate.

PR #710, opened 2026-09-17
Provide JavaSMT main to execute as solver for SMT-COMP

gcarpio21:
# Command-line interface for running JavaSMT

Adds a command-line interface so that JavaSMT can be run like a regular SMT solver on a single SMT-LIB 2 file, e.g., for SMT-COMP or for benchmarking the solvers behind JavaSMT with BenchExec. 

```
$ javasmt --solver z3 benchmark.smt2
sat
```

`JavaSMTMain` parses the assertions with `FormulaManager.parseAll`, checks them in one `ProverEnvironment`, and prints exactly one of `sat`, `unsat`, or `unknown` on stdout; diagnostics got to stderr.  `--solver` selects the solver (default SMTInterpol), `--logic` is forwarded to OpenSMT. The launcher `javasmt` in the project root builds the classpath from `bin/` or the JAR and all solver runtimes, passes `-X*` arguments to the JVM, and shares its Java version check and platform handling with `runExamples.sh`.

- Input without `(check-sat)` is rejected.
- Unparsable input and solvers without a parser (Yices2) are reported with a message on stderr. Unexpected exceptions terminate the program with a stack trace, so that a benchmarking framework can report the exception class.
- Ctrl+C and SIGTERM request a shutdown from the solver via a JVM shutdown hook; `unknown` is reported and the solver is closed. Solvers that do not respond get a grace period of 10 s.
- `parseAll` runs before the `ProverEnvironment` is created, because Princess does not register symbols declared after its prover exists.

`JavaSMTMainTest` covers the argument parsing and the whole interface in-process with SMTInterpol and Princess, including all error cases and shutdown requests before and during solving; one test runs `main` in a separate JVM and sends SIGTERM.

### DIFF
--- a/src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
+++ b/src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
@@ -1,0 +1,128 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.sosy_lab.java_smt.cmdline;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSortedSet;
+import java.io.PrintStream;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import org.sosy_lab.java_smt.cmdline.CmdLineArgument.CmdLineArgument1;
+import org.sosy_lab.java_smt.cmdline.CmdLineArgument.PropertyAddingCmdLineArgument;
+
+/** Processes command-line arguments for JavaSMT. */
+public final class CmdLineArguments {
+
+  private CmdLineArguments() {}
+
+  private static final ImmutableSortedSet<CmdLineArgument> CMD_LINE_ARGS =
+      ImmutableSortedSet.of(
+          new CmdLineArgument1("--solver", "-solver")
+              .settingOption("solver.solver")
+              .withDescription("Set SMT solver to use"),
+          new CmdLineArgument1("--logic", "-logic")
+              .settingOption("solver.opensmt.logic")
+              .withDescription("Set SMT logic (only for OpenSMT)"),
+          new PropertyAddingCmdLineArgument("--help", "-h", "-help")
+              .settingProperty("help", "true")
+              .withDescription("Print this help message"));
+
+  /**
+   * Processes command-line arguments and returns a map of option names to values.
+   *
+   * @param pArgs Raw command-line arguments
+   * @return Map of option names to their values
+   * @throws InvalidCmdlineArgumentException if arguments are invalid
+   */
+  public static Map<String, String> processArguments(String[] pArgs)
+      throws InvalidCmdlineArgumentException {
+    Preconditions.checkNotNull(pArgs);
+
+    Map<String, String> properties = new HashMap<>();
+    Iterator<String> argsIt = Arrays.asList(pArgs).iterator();
+
+    while (argsIt.hasNext()) {
+      String arg = argsIt.next();
+
+      boolean found = false;
+      for (CmdLineArgument cmd : CMD_LINE_ARGS) {
+        if (cmd.apply(properties, arg, argsIt)) {
+          found = true;
+          break;
+        }
+      }
+
+      if (!found) {
+        if (arg.startsWith("-")) {
+          throw new InvalidCmdlineArgumentException("Unknown command-line argument: " + arg);
+        } else {
+          if (properties.containsKey("smt2.file")) {
+            throw new InvalidCmdlineArgumentException(
+                "Multiple input files are not supported: "
+                    + properties.get("smt2.file")
+                    + " and "
+                    + arg);
+          }
+          Path file = Paths.get(arg);
+          properties.put("smt2.file", file.toString());
+        }
+      }
+    }
+
+    return properties;
+  }
+
+  static boolean isOldStyleArgument(String arg) {
+    return arg.length() > 2 && arg.startsWith("-") && !arg.startsWith("--");
+  }
+
+  private static void printVersion(PrintStream out) {
+    out.println();
+    Package pkg = CmdLineArguments.class.getPackage();
+    String version = pkg != null ? pkg.getImplementationVersion() : "unknown";
+    out.println("JavaSMT " + version);
+  }
+
+  /**
+   * Prints the help message to the given output stream.
+   *
+   * @param out The output stream to print to
+   */
+  public static void printHelp(PrintStream out) {
+    printVersion(out);
+    out.println();
+    out.println("Usage: javasmt [options] <file.smt2>");
+    out.println("Options:");
+    for (CmdLineArgument cmdLineArg : CMD_LINE_ARGS) {
+      if (!isOldStyleArgument(cmdLineArg.getMainName())) {
+        out.println(" " + cmdLineArg);
+      }
+    }
+    out.println();
+    out.println("JavaSMT executes SMT2 files using the selected solver.");
+    out.println("javasmt --solver <SOLVER> <file.smt2>");
+  }
+
+  static void putIfNotExistent(Map<String, String> properties, String key, String value)
+      throws InvalidCmdlineArgumentException {
+    if (properties.containsKey(key) && !properties.get(key).equals(value)) {
+      throw new InvalidCmdlineArgumentException(
+          String.format(
+              "Option %s specified twice on command-line with values '%s' and '%s'.",
+              key, properties.get(key), value));
+    }
+    properties.put(key, value);
+  }
+}
--- a/src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
+++ b/src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
@@ -14,7 +14,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSortedSet;
 import java.io.PrintStream;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -75,7 +74,7 @@ public final class CmdLineArguments {
                     + " and "
                     + arg);
           }
-          Path file = Paths.get(arg);
+          Path file = Path.of(arg);
           properties.put("smt2.file", file.toString());
         }
       }
--- a/src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
+++ b/src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
@@ -14,7 +14,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -91,7 +91,7 @@ public class JavaSMTMain {
       return;
     }
 
-    if (!Files.isReadable(Paths.get(options.smt2File))) {
+    if (!Files.isReadable(Path.of(options.smt2File))) {
       throw Output.fatalError(
           "Please provide a valid, readable SMT2 file: %s", options.smt2File);
     }
@@ -141,7 +141,7 @@ public class JavaSMTMain {
 
     String input;
     try {
-      input = Files.readString(Paths.get(options.smt2File));
+      input = Files.readString(Path.of(options.smt2File));
     } catch (IOException e) {
       throw Output.fatalError("Could not read SMT2 file: %s", e.getMessage());
     }
--- a/src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
+++ b/src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
@@ -1,0 +1,180 @@
+/*
+ * This file is part of JavaSMT,
+ * an API wrapper for a collection of SMT solvers:
+ * https://github.com/sosy-lab/java-smt
+ *
+ * SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.sosy_lab.java_smt.cmdline;
+
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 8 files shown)


src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+      return ERROR_EXIT_CODE;
+    }
+
+    // The parser silently ignores everything that is not a declaration, definition, or assertion,
+    // so an empty or unparsable file would be reported as sat. Every benchmark that asks a question
+    // contains (check-sat), so we require it.
+    final boolean hasCheckSat;
+    try {
+      hasCheckSat = containsCheckSat(input);
+    } catch (IllegalArgumentException e) {
```
baierd, line 171: We should not silently ignore all input that we can't and don't want to handle.
This includes 2 categories:
- malformed/illegal input. For example `(check-sat-garbage)` and `(check-sat 1)` are currently accepted.
- unsupported input. `check-sat-assuming` is not supported by `parseAll`, and is silently ignored.

Most of this is a problem that should be tackled by `parseAll` in my opinion. @daniel-raffler would you be so kind and open a issue to track this problem, and add unit tests for all allowed SMTLIB2 commands, including malformed ones, e.g. with more arguments than allowed, arguments before the command etc. Then improve the state-machine in `parseAll` to throw for unsupported or malformed input as long as it can change results. Example:
```
(declare-fun p () Bool)
(assert p)
(check-sat-assuming ((not p)))
```
Is UNSAT, but we return SAT currently.

@gcarpio21 what must be blocked are all `pop` and all resetting commands for now, as we do not properly track assertion stacks through parseAll and also ignore resets. Please add a TODO that explains the problem and that we add it later. `(exit)` should also throw, except for if its the very last command.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+            e.getMessage() == null ? "" : " " + e.getMessage());
+        return ERROR_EXIT_CODE;
+      }
+      // Any other exception is unexpected, e.g., a bug in a solver binding, and is intentionally
+      // not caught, such that it terminates the program with a stack trace on stderr.
+
+      boolean isUnsat;
+      try (ProverEnvironment prover = context.newProverEnvironment()) {
+        for (BooleanFormula formula : formulas) {
+          prover.addConstraint(formula);
```
baierd, line 230: This is currently incorrect.
Sat checks should be performed for each and every `(check-sat)` command. We currently perform one giant SAT check, but this violates SMTLIB2. 

I propose to only allow ONE `(check-sat)` (i.e. use `parse` instead of `parseAll`) for now. And extend it later on.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+      return 0;
+
+    } catch (InvalidConfigurationException e) {
+      Output.error(err, "Invalid configuration: %s", describe(e));
+      return ERROR_EXIT_CODE;
+    } catch (InterruptedException e) {
+      // Thrown by the solver after a shutdown request, see ShutdownHook.
+      String reason = shutdownNotifier.shouldShutdown() ? shutdownNotifier.getReason() : "";
+      logManager.log(Level.WARNING, "SMT execution was interrupted.", reason);
+      Output.println(out, "unknown");
```
baierd: Please transform all repeated strings into constants. When it comes to Solver results like UNKNOWN, SAT, and UNSAT, a Enum might be better, as we can re-use it internally.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+      }
+      Output.println(out, isUnsat ? "unsat" : "sat");
+      return 0;
+
+    } catch (InvalidConfigurationException e) {
+      Output.error(err, "Invalid configuration: %s", describe(e));
+      return ERROR_EXIT_CODE;
+    } catch (InterruptedException e) {
+      // Thrown by the solver after a shutdown request, see ShutdownHook.
+      String reason = shutdownNotifier.shouldShutdown() ? shutdownNotifier.getReason() : "";
```
baierd: Please do not use shorthand if-then-else. They are hard to read and provide no benefit at all.

```java
String reason = "";
if (shutdownNotifier.shouldShutdown()) {
  reason = shutdownNotifier.getReason();
}
```

Is much easier to read.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+      // not caught, such that it terminates the program with a stack trace on stderr.
+
+      boolean isUnsat;
+      try (ProverEnvironment prover = context.newProverEnvironment()) {
+        for (BooleanFormula formula : formulas) {
+          prover.addConstraint(formula);
+        }
+        isUnsat = prover.isUnsat();
+      }
+      Output.println(out, isUnsat ? "unsat" : "sat");
```
baierd: While `PrintStream` does not swallow `IOException`s from `Appendable`s, `System.out` and `System.err` record these failures internally, and flushing does not report them. Thus result delivery can fail while the process reports success. Better: after the existing flushes, check both streams with `checkError()` and change the exit status to `ERROR_EXIT_CODE` on failure.

Also, redirect the launchers JVM-probe failure messages to `stderr` using `>&2;` that branch currently uses plain echo on `stdout`.

@PhilippWendler why are we using `PrintStream` in CPAchecker?

src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
```diff
@@ -0,0 +1,134 @@
...
+          throw new InvalidCmdlineArgumentException("Unknown command-line argument: " + arg);
+        } else {
+          if (properties.containsKey(FILE_OPTION)) {
+            throw new InvalidCmdlineArgumentException(
+                "Multiple input files are not supported: "
+                    + properties.get(FILE_OPTION)
+                    + " and "
+                    + arg);
+          }
+          Path file = Path.of(arg);
```
baierd: This may throw a `InvalidPathException`, which is currently unhandled. You can just wrap it in a `InvalidCmdlineArgumentException`.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
```diff
@@ -0,0 +1,286 @@
...
+        && options.solver != Solvers.OPENSMT) {
+      logManager.log(
+          Level.WARNING,
+          "Option --logic is only effective with OpenSMT solver, but solver is set to",
+          options.solver + ". The logic setting will be ignored.");
+    }
+
+    final String input;
+    try {
+      input = Files.readString(Path.of(options.smt2File));
```
baierd, line 162: This may throw a `InvalidPathException`, which is currently unhandled. You can just wrap it in a `InvalidCmdlineArgumentException`.

src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param pArgs Raw command-line arguments
+   * @return Map of option names to their values
+   * @throws InvalidCmdlineArgumentException if arguments are invalid
+   */
+  public static Map<String, String> processArguments(String[] pArgs)
+      throws InvalidCmdlineArgumentException {
+    Preconditions.checkNotNull(pArgs);
+
+    Map<String, String> properties = new HashMap<>();
+    Iterator<String> argsIt = Arrays.asList(pArgs).iterator();
```
baierd: Please avoid `Arrays.asList()` (and most other default `toList`, `asList` etc. methods). Take a look at [CPAcheckers StyleGuide](https://gitlab.com/sosy-lab/software/cpachecker/-/blob/main/doc/StyleGuide.md?ref_type=heads) for why and for alternatives. In this case, just use `ImmutableList.of(pArgs)`.

src/org/sosy_lab/java_smt/test/JavaSMTMainTest.java
```diff
@@ -0,0 +1,436 @@
...
+                    "-cp",
+                    System.getProperty("java.class.path"),
+                    JavaSMTMain.class.getName(),
+                    "--solver",
+                    "SMTINTERPOL",
+                    smt2File(pigeonholeInput(12))))
+            .redirectOutput(outFile.toFile())
+            .redirectError(ProcessBuilder.Redirect.DISCARD)
+            .start();
+    Thread.sleep(3000); // let the JVM start and the solver run
```
baierd: The test may fail after this waiting period (unlikely, but possible). Please add cleanup that always works by adding a finally block.

src/org/sosy_lab/java_smt/test/JavaSMTMainTest.java
```diff
@@ -0,0 +1,436 @@
...
+    assertThat(result).containsKey("help");
+  }
+
+  @Test
+  public void testProcessArgumentsOnlyFile() throws Exception {
+    Map<String, String> result = CmdLineArguments.processArguments(new String[] {"input.smt2"});
+    assertThat(result.get("smt2.file")).isEqualTo("input.smt2");
+  }
+
+  @Test(expected = InvalidCmdlineArgumentException.class)
```
baierd: Please use `assertThrows` instead of `expected` annotation, as `assertThrows` provides better error messages. (This should be applied in all cases).

baierd (changes requested):
Nice job!
We only need to work out some minor points before merging this.

Could you please also add a `--help` command that explains restrictions and allowed arguments?

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
daniel-raffler: Hello Daniel,

> We should not silently ignore all input that we can't and don't want to handle. This includes 2 categories:
> 
>     * malformed/illegal input. For example `(check-sat-garbage)` and `(check-sat 1)` are currently accepted.
> 
>     * unsupported input. `check-sat-assuming` is not supported by `parseAll`, and is silently ignored.
> 
> 
> Most of this is a problem that should be tackled by `parseAll` in my opinion

I did some work on this a while ago, and there is an open PR [here](https://github.com/sosy-lab/java-smt/pull/630) that should at least take case of your first point about illegal inputs

If we want to support commands like `check-sat-assuming`, `pop` or `exit`, I think it would be easiest to write a proper parser with ANTLR. The tokenizer is more of a stop-gap solution that works well enough for parsing simple formulas, but it was never meant to execute entire SMTLIB scripts

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
PhilippWendler: I disagree. Having a single point of assignment is typically easier to follow (less mutable state) and allows variables to be final.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
PhilippWendler: For example because we do not want to stop proceeding through all the code for statistics and output files if there is an I/O error for the statistics output.

src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
PhilippWendler: CPAchecker's style guide definitively does not forbid `Arrays.asList()`, and I wouldn't know what "default `toList`, `asList` methods" are.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
baierd: Fair point.
But that is also quite easy.

```
final String reason;
if (shutdownNotifier.shouldShutdown()) {
  reason = shutdownNotifier.getReason();
} else {
  reason = "";
}
```

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
baierd: Makes sense. Thanks for the explanation.
The question now is whether we should use `PrintStream` or not here as well. Do you have a strong opinion on that @PhilippWendler ?

src/org/sosy_lab/java_smt/cmdline/CmdLineArguments.java
baierd: My bad, i was a little over-eager here.
`Collectors.toList()` should be avoided.

The code can be improved either way by using Guavas utilities from its `Iterators` class;
`Iterator<String> argsIt = Iterators.forArray(pArgs);`

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
baierd: ANTLR may be overkill for what we want.

I opened a issue for this [here](https://github.com/sosy-lab/java-smt/issues/712), lets discuss it there.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
PhilippWendler: This depends a lot on the use case. For I/O in most cases it is bad, but for particular situations it can be ok.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
gcarpio21: `JavaSMTMain.checkScript` now rejects `(push ...)`, `(pop ...)`, `(reset)` and `(reset-assertions)`, any `(exit)` that is not the last command, anything that looks like `(check-sat)` but is not exactly `(check-sat)`, and text outside a command, which the tokenizer would otherwise drop. `(push ...)` is included because the tokenizer's `isForbiddenToken` covers it and `parseAll` fails on it  anyway.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
gcarpio21: I kept `parseAll` for now. `parse` delegates to `parseAll` and then throws unless the query contains exactly one assertion. The restriction is instead enforced in `JavaSMTMain.checkScript` before parsing: exactly one `(check-sat)`, no `(assert ...)` after it.

src/org/sosy_lab/java_smt/cmdline/JavaSMTMain.java
gcarpio21: > Fair point. But that is also quite easy.
> 
> ```
> final String reason;
> if (shutdownNotifier.shouldShutdown()) {
>   reason = shutdownNotifier.getReason();
> } else {
>   reason = "";
> }
> ```

I decided to replace the shorthands with this suggestion.

gcarpio21:
> Nice job! We only need to work out some minor points before merging this.
> 
> Could you please also add a `--help` command that explains restrictions and allowed arguments?

the `--help` command is now expanded to do this.

